PR #7, opened 2015-02-21
Merged Guides content from https://github.com/emberjs/website/pull/1956

bsclifton:

PR #17, opened 2015-02-25
Add Ember Inspector guides

teddyzeenny:
Screenshots at https://github.com/emberjs/website/pull/2030


joostdevries:
@bsclifton this is related to your pr right?

bsclifton:
This is really similar! It has images, which make it easier to understand.  I can definitely merge this into my PR here: 
https://github.com/emberjs/guides/pull/7

source/ember-inspector/container.md
```diff
@@ -0,0 +1,29 @@
+Ember.js implements dependency injection by using a container.  Every Ember
+application has a container that creates instances for you. You can
+inspect these instances using the inspector's container tab.
+
+
+<img src="/images/guides/ember-inspector/container-screenshot.png" width="680"/>
```
trek: images can't be absolute, since the output of a build lives inside a subdirectory.

trek:
Looks good but we'd need to make all the absolute urls relative instead.

teddyzeenny:
Updated to use relative image urls. Thanks @trek for reviewing.

bsclifton:
Wow- lots of updates since I've seen this last, great job :smile: When I made the comment above, there was only one page which had Ember Inspector details, which was similar to what I had (but better).  I'm going to remove the content from my PR

knownasilya:
@teddyzeenny awesome job!

PR #27, opened 2015-03-03
Moves version selector out of sidebar nav borders

danmcclain:
Updates the styling to get rid of the double bordering on the select2, the result is below

ember js - models working with records 2015-03-03 15-24-03


rwjblue:
:heart:

danmcclain:
The fact that the horizontal rule on the right and the top of the nav on the left line up was a happy accident. Would have likely went back and lined them up after I noticed that they weren't :smile:

PR #32, opened 2015-03-07
Remove select2 input outline on focus

karelvuong:
This will make the select2 input more consistent with the site's inputs and their lack of `outline` when focused.

### DIFF
--- a/source/stylesheets/select2.scss
+++ b/source/stylesheets/select2.scss
@@ -2,6 +2,7 @@
   box-sizing: border-box;
   display: inline-block;
   margin: 0;
+  outline: none;
   position: relative;
   vertical-align: middle; }
   .select2-container .select2-selection--single {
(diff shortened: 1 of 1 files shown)


trek:
Can you post a screenshot?

karelvuong:
#### Pre-fix:

Focus on input:
screen shot 2015-03-07 at 8 38 34 am

Click on option; lose focus on input:
screen shot 2015-03-07 at 8 38 40 am

---

#### Post-fix:

Focus on input:
screen shot 2015-03-07 at 8 40 32 am

Click on option; lose focus on input:
screen shot 2015-03-07 at 8 49 33 am

trek:
Ach, that scss file is the one that comes with select2. Can you make these changes in a new/existing file we control?

karelvuong:
Sure! Any thoughts on what it should be named? 

I was also thinking of reverting the changes in `select2.scss` and moving it into vendor alongside `swiftype.autocomplete.scss` to keep the lib styles together.

trek:
we can name it select2 in both places. One in vendor, one not.

PR #34, opened 2015-03-07
update select2 style to be more like sidebar doc

mateuspv:
newone


rwjblue:
Looks nice, @trek would you prefer the select to be white or blend with the TOC colors?

rwjblue:
Also, this scss file is the stock one provided by select2, I think we should make our modifications in a separate file, and also use variables for the colors (so when we change it'll be easier)

trek:
I think we want it to be the same level of emphasis as the other control on the page (the search box), so it's white by design, not accident.

mateuspv:
no problem ^^

PR #35, opened 2015-03-07
Adds file names to code blocks when present

### LINKED ISSUE #30: Use fenced code metadata to annotate both file location and language
From the common mark spec:
> The line with the opening code fence may optionally contain some text following the code fence; this is trimmed of leading and trailing spaces and called the info string. The info string may not contain any backtick characters. (The reason for this restriction is that otherwise some inline code would be incorrectly interpreted as the beginning of a fenced code block.)

Where we annotate the code's language
~~~
```javascript
var someCode;
~~~

We can annotate filename and extension:
~~~
```app/templates/foo.hbs
{{rock-on}}
~~~

We have a [custom highlighter](https://github.com/emberjs/guides/blob/0c9a158855a3495ed4413c1512cb984c954adc1c/lib/highlighter.rb) that outputs particular markup/styles for specific languages.

A PR that addresses this Issue will:

- [ ] continue to detect bare named language fences
- [ ] begin to detect language name from file path fences
- [ ] output nice markup/style that decorates an HTML code block something that shows file location

@danmcclain has a [sample markup/design](https://github.com/emberjs/guides/issues/12#issuecomment-77424950)

danmcclain:
Closes #30

This only updates 1 code fence as an example, visible at http://localhost:4567/templates/the-application-template/


danmcclain:
Wanted to share progress with it so far for feedback. Will add tests later today/tomorrow

/cc @rwjblue @trek

trek:
Seems good. You can leave the `require 'pry'` in. It doesn't affect anything and helps those debugging.

danmcclain:
Sounds good. There are existing tests failing on master. Do you mind if I add tests for the fencing and don't touch the existing a tests in this PR, then address them in a second PR?

trek:
sounds good.

danmcclain:
@trek: Tests added. Ready to be merged

PR #39, opened 2015-03-10
Something like this?

joshross4:
My first attempt at a PR.


trek:
@joshross4 the urls are broken for the opposite reason: they need to be relative and not absolute. 

Everything in this project need to be built and referenced relative to the project root. We publish the output of this repo to https://github.com/emberjs/guides and it goes into a directory that represents the version.

PR #51, opened 2015-03-14
Migrates application section to ember-cli style

trek:


source/application/index.md
```diff
@@ -1,27 +1,14 @@
...
-
-```javascript
-window.App = Ember.Application.create();
-```
-
-Most people call their application `App`, but you can call it whatever
-makes the most sense to you. Just make sure it starts with a capital
-letter.
+Every Ember.js application has one instance of `Ember.Application`.
+This object is located at `./app/app.js` inside your application.
```
rwjblue: 'inside your project' perhaps?

Issue #10, opened 2015-02-22
Extract TOC and Pagination Helpers


trek:
Our table of contents and pagination generation is generally useful for anyone writing book-style content. We should extract these helpers so it will be easy to have a [Gitbook](https://github.com/GitbookIO/gitbook) like site.

A PR that successfully addresses this issue will:
- [ ] create a new repo for making a middleman plugin
- [ ] allow the author to provide a file in `data/<name>.yml` that is an array of chapter objects that contain section arrays. The data is in the following format:

``` yaml
- title: 'Introduction'
  url: 'index'
- title: 'Chapter 1: Call me Ishmael'
  url: 'chapter-1'
  sections:
    - title: 'A History of horror'
      url: 'index'
    - title: 'The Whale'
      url: 'the-whale'
- title: 'Chapter 2: The Imapler'
  url: 'chapter-2'
  sections:
    - title: 'Sex and Fear'
      url: 'index'
```

These will correspond to a file structure like this:

``` shell
.
├── data
│   └── chapters.yml
└── source
    ├── chapter-1
    │   ├── index.md
    │   └── the-whale.md
    ├── chapter-2
    │   └── index.md
    └── index.md
```
- [ ] provide a helper to generat the table of contents on each page
- [ ] provide a helper to link to the next section. If you're at the end of a chapter, this links to the first section in the next chapter if there is one
- [ ] provide a helper to link to the previous section. If you're at the beginning of a chapter, this links to the last section in the previous chapter if there is one
- [ ] provide a a helper to show the title of the next chapter/section pair
- [ ] provide a a helper to show the title of the previous chapter/section pair
- [ ] unit test all of this (not integation like we've had. but unit tests)

BONUS ROUND:
- [ ] like Gitbook, allow the author to express table of contents as markdown

``` yaml
- title: 'Introduction'
  url: 'index'
- title: 'Chapter 1: Call me Ishmael'
  url: 'chapter-1'
  sections:
    - title: 'A History of horror'
      url: 'index'
    - title: 'The Whale'
      url: 'the-whale'
- title: 'Chapter 2: The Imapler'
  url: 'chapter-2'
  sections:
    - title: 'Sex and Fear'
      url: 'index'
```

becomes

``` markdown
* [Introduction](index.md)
* [Chapter 1: Call me Ishmael]
    * [A History of horror](chapter-1/index.md)
    * [The Whale](chapter-1/the-whale.md)
* [Chapter 2: The Imapler]
[...]

PR #53, opened 2015-03-14
Handle nil language in highlighter

tundal45:
I noticed that `middleman build` was breaking so I thought it would be good to have that be executed as a part of the CI build. So I have added `travis.yml`. I believe there is an additional step in repository settings that needs to happen for this to work. I have also replaced the existing specs for `toc` with unit tests that do not require a local middleman server running. The hope is that this will help with https://github.com/emberjs/guides/issues/10.

### DIFF
--- a/lib/highlighter.rb
+++ b/lib/highlighter.rb
@@ -2,6 +2,7 @@ require "redcarpet"
 require "coderay"
 
 module Highlighter
+  class MissingLanguageError < StandardError; end
   class << self
     def registered(app)
       app.helpers Helpers
@@ -11,6 +12,11 @@ module Highlighter
 
   module Helpers
     def _highlight(string, language, class_name=nil)
+      error_message = "Code blocks must be fenced with proper language designated. If you don't know what language to use, specify ```text.\n\n"
+      error_message << "Offending Code Block:\n\n#{string}"
+
+      raise MissingLanguageError, error_message if language.nil?
+
       language, file_name = _detect_language_and_filename(language)
 
       result = %Q{<div class="highlight #{language} #{class_name}">}
@@ -23,6 +29,7 @@ module Highlighter
     end
 
     def _detect_language_and_filename(language)
+
       file_name = nil
       bare_language_regex = /\A\w+\z/
 
--- a/lib/toc.rb
+++ b/lib/toc.rb
@@ -57,19 +57,13 @@ module TOC
     end
 
     def page_title
-      buffer = ""
-
-      if current_guide
-        buffer << current_guide.title
-        if current_chapter
-          buffer << ": "
-          buffer << current_chapter.title
-        end
+      if current_guide && current_chapter
+        "#{current_guide.title}: #{current_chapter.title}"
+      elsif current_guide
+        current_guide.title
       else
-        buffer << "Guides"
+        "Guides"
       end
-
-      buffer
     end
 
     def guide_name
@@ -78,9 +72,9 @@ module TOC
 
     def chapter_name
       if current_chapter
-        return current_chapter.title
+        current_chapter.title
       else
-        return ""
+        ""
       end
     end
 
@@ -111,6 +105,10 @@ module TOC
       end
     end
 
+    def current_guide_index
+      data.guides.find_index(current_guide)
+    end
+
     def current_chapter
       return unless current_guide
 
@@ -122,6 +120,11 @@ module TOC
         chapter.url == chapter_path
       end
     end
+
+    def current_chapter_index
+      return unless current_guide
+      current_guide.chapters.find_index(current_chapter)
+   end
 
     def chapter_links
       %Q{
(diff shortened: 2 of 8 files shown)


trek:
Some feedback on the tests: I'm not a huge fan of this level of reliance on mocks. I feel like they've described the _exact_ implementation and not the expected behavior, which makes them particularly brittle to refactoring.

Ideally the way to test this is to create a middleman site instance, add this TOC mixin, and then call methods on the instance and verify their output

tundal45:
@trek As we discussed, the tests now primarily mock out `current_page` & `data.guides` to simulate all the edge cases for various methods in the helper. The only place `request` had to be mocked out was for `toc_for`. This should be less brittle than the earlier version. Let me know if there are still other improvements that can be made. Thanks!

trek:
Did you forget to push some commits? I'm still only seeing the 4 commits from a few days ago.

tundal45:
@trek I had accidentally amended the travis ci commit with the spec changes. Fixed now.

lib/toc.rb
```diff
@@ -77,11 +67,8 @@ def guide_name
     end
 
     def chapter_name
-      if current_chapter
-        return current_chapter.title
-      else
-        return ""
-      end
+      return "" unless  current_chapter
```
trek: We could probably nuke the returns in the original, but I don't think this adds readability

lib/toc.rb
```diff
@@ -57,19 +57,9 @@ def toc_for(guides)
...
-          buffer << current_chapter.title
-        end
-      else
-        buffer << "Guides"
-      end
-
-      buffer
+      return "Guides" unless current_guide
+      return "#{current_guide.title}" unless current_chapter
+      "#{current_guide.title}: #{current_chapter.title}"
```
trek: https://github.com/emberjs/guides/pull/53/files#r26567477

spec/toc_spec.rb
```diff
@@ -1,64 +1,396 @@
...
 
-unless `which phantomjs`.empty?
-  Capybara.register_driver :poltergeist do |app|
-    Capybara::Poltergeist::Driver.new(app, { :inspector => true, })
+  before(:each) do
+    class HelperTester
+      include TOC::Helpers
+    end
+
+    allow(helper).to receive(:current_page).and_return(testing_page)
```
trek: Your comment made it sound like you got rid of this level of mocking?

spec/toc_spec.rb
```diff
@@ -1,64 +1,396 @@
-#NOTE: start the middleman as in readme, then run bundle exec rspec
+require "spec_helper"
 
-require 'capybara/rspec'
-require 'capybara/poltergeist'
-require 'pry'
+describe Highlighter::Helpers do
```
trek: Your comment made it sound like you got rid of this level of mocking?

spec/toc_spec.rb
```diff
@@ -1,64 +1,396 @@
...
 
-describe "TOC", :type => :feature do
-  include Capybara::DSL
+  after do
+    # Remove our testing class
+    Object.send :remove_const, :HelperTester
+  end
+
+  describe "#toc_for" do
+    let(:skip_chapter)    { double(title: "Skip Me", url: "skip-me") }
```
trek: Your comment made it sound like you got rid of this level of mocking?

lib/toc.rb
tundal45: Makes sense. I'll switch back to the old version sans explicit `return`s.

spec/toc_spec.rb
tundal45: `current_page` is not something that we override in the helper and it felt was not in violation of mocking something we don't own. Is the issue that `current_page` is mocked out in a `before` block instead of each individual `example`?

trek: ah, yes, current page should stay. I guess I mostly confused by the giant `mock`/`double` block still being around?

tundal45:
@trek I need to squash the intermediary commits but this should have tests that use actual data rather than relying heavily on mocks & stubs. It also updates the highlighter to raise an error when there are nil languages rather than silently approving them.

lib/highlighter.rb
```diff
@@ -11,6 +12,11 @@ def registered(app)
 
   module Helpers
     def _highlight(string, language, class_name=nil)
+      error_message = "Code blocks must be fenced with proper language designated. If you don't know what language to use, specify text.\n\n"
```
trek: "If you don't know what language to use, specify text" -> If you don't know what language to use, specify ```text

tundal45:
@trek This addresses all the comments made here as well as those we discussed offline. I have [travis running builds](https://travis-ci.org/tundal45/guides/builds/55284701) on my fork as well so that should work on the main repo once this gets merged in & setup on travis.

trek:
@tundal45 I think the ordering of PRs is wrong. We'd need to wait for a the PR that adds languages to all language-less code blocks. If we merged this first, nobody could build the site!

trek:
Otherwise, looks good! Thanks for all the hard work on this. I know it's been a lot of back and forth.

tundal45:
@trek Happy to help. I will open the PR that fixes the guides with language-less code blocks shortly.

tundal45:
yayyayyay! <3

PR #56, opened 2015-03-16
Convert views (as best we can) to ember-cli

trek:


rwjblue:
Looks good, but needs a rebase...

trek:
rebased.

PR #58, opened 2015-03-16
Moved rails-assets-* to their own block.

tarebyte:
Hi There,

I was running the guides and I received the following warning.

```
Warning: this Gemfile contains multiple primary sources.
Using `source` more than once without a block is a security risk, and may result in installing unexpected gems.
To resolve this warning, use a block to indicate which gems should come from the secondary source.
To upgrade this warning to an error, run `bundle config disable_multisource true`.
```

According to the [rails-assets website](https://rails-assets.org/) the proper form seems to be to put the rails-assets specific gems into their own block.

There were also some updates to other dependencies, but I can revert them if you want.

Thanks for all the hard work, these guides are awesome! I'm really excited to seem them with the ember-cli changes.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,4 @@
 source 'https://rubygems.org'
-source 'https://rails-assets.org'
 
 gem "redcarpet"
 gem "activesupport"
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

PR #60, opened 2015-03-17
Moves 'concepts' section to cli style

trek:


source/concepts/naming-conventions.md
```diff
@@ -1,120 +1,117 @@
...
-you access points within an Ember object's lifecycle to intercept and 
-execute code to modify the default behavior at these points to meet 
+`application` template. Hooks are implemented as methods and provide
+you access points within an Ember object's lifecycle to intercept and
+execute code to modify the default behavior at these points to meet
 your needs. Ember provides several hooks for you to utilize for various
-purposes (e.g. `model`, `setupController`, etc). In the example below 
-`App.ApplicationRoute`, which is an `Ember.Route` object, implements 
+purposes (e.g. `model`, `setupController`, etc). In the example below
+`App.ApplicationRoute`, which is an `Ember.Route` object, implements
```
rwjblue: In some scenarios we talk in `type:name`, but here we reference the global location (`App.ApplicationRoute`).  Does this need to be changed?

source/concepts/naming-conventions.md
```diff
@@ -1,120 +1,117 @@
...
-Each of your routes will have a controller, and a template with the 
+Each of your routes will have a controller, and a template with the
 same name as the route.
 
 Let's start with a simple router:
 
-```javascript
-App.Router.map(function() {
+```app/router.js
+export default Ember.Router.extend().map(function(){
```
rwjblue: Ember CLI uses the following default `app/router.js`, should we use the default blueprint instead of what is here?

```javascript
var Router = Ember.Router.extend({
  location: config.locationType
});

Router.map(function() {
});

export default Router;
```

source/concepts/naming-conventions.md
```diff
@@ -176,20 +173,20 @@ handlers.
 
 * If your dynamic segment ends in `_id`, the default `model`
   hook will convert the first part into a model class on the
-  application's namespace (`post` becomes `App.Post`). It will
+  application's namespace (`post` looks for `app/models/post`). It will
```
rwjblue: `app/models/post.js`

source/concepts/what-is-ember-js.md
```diff
@@ -141,14 +141,10 @@ to a computed property.
 
 #### Auto-updating Templates
 
-Ember uses Handlebars, a semantic templating library. To take data from your JavaScript application
-and put it into the DOM, create a `<script>` tag and put it into your HTML, wherever you'd like the
-value to appear:
+Ember uses Handlebars, a semantic templating library
```
rwjblue: Need some sort of punctuation at the end of this sentence? `.` or `:` perhaps?

source/concepts/naming-conventions.md
trek: I wanted to simply the example to only show what is needed to get the specific effect. I was hoping people understand the difference between the example and the generated code.

rwjblue: Seems good to me, we can gauge any future issues and adjust if needed.  I just wanted to ensure that this divergence from the default blueprint was intentional (which it is).  Carry on :smile_cat:

trek:
@rwjblue review again and merge if you like.

locks:
seems :+1: to me

PR #62, opened 2015-03-18
CLI-style for 'understanding-ember' section

trek:


source/understanding-ember/debugging.md
```diff
@@ -1,20 +1,19 @@
-### Debugging Ember and Ember Data
+### Debugging Ember
+All Ember developers should install the
+[Ember Inspector](https://github.com/tildeio/ember-extension)
```
rwjblue: Should be https://github.com/emberjs/ember-inspector

source/understanding-ember/dependency-injection-and-service-lookup.md
```diff
@@ -41,14 +41,15 @@ A common use-case for service lookup is that of a singleton service. Often, thes
...
-var App = Ember.Application.create();
-App.SessionController = Ember.Controller.extend({
+```app/controllers/session.js
+export default Ember.Controller.extend({
   isAuthenticated: false
 });
-// The index controller may need access to that state:
-App.IndexController = Ember.Controller.extend({
+```
+```app/controllers/index.js
```
rwjblue: Need blank line between these two blocks

PR #65, opened 2015-03-19
[WIP] Updates getting ember section for ember-cli style

locks:
I was unsure if I should PR the fix for the links separately, but since I'm trying to get everything through for the next release, it probably doesn't matter.

I will add an `##Ember CLI` section akin to the current `##Bower` and `##Rubygems`. Despite the CLI focus of the guides, it seems useful to maintain the links to the builds page as well as bower/Rubygems. c/d?


joostdevries:
Personally I'd prefer the "Installing Ember" section from @bsclifton's #7 for this because we want people to install Ember CLI.

locks:
@joostdevries so scrap this section altogether?

PR #66, opened 2015-03-19
[New Guide] Composing Components

knownasilya:
So one of the biggest issues with components, is that users don't know how do use them, they are new and scary. This guide helps users dive into using components as they were meant to be used (beyond the single lonesome component).

This guide takes the *Data Down* and *Actions Up* approach that is being favored and will be prevalent once all of the Ember 2.0 component features are in (`mut`, `attrs`, etc).

[preview rendered](https://github.com/knownasilya/guides/blob/patch-1/source/components/composing-components.md)

### DIFF
--- a/source/components/composing-components.md
+++ b/source/components/composing-components.md
@@ -1,0 +1,261 @@
+Components really shine when you use them to their full potential, that is, by composing them with one another.
+Take for example the `<ul>` element, and the fact that only `<li>` elements are appropriate as children.
+If we want the same type of behavior, then we have to compose our components.
+
+Just like we compose regular HTML elements, we can do the same with components.
+
+```app/templates/application.hbs
+{{#user-list users=activeUsers sortBy='name' as |user|}}
+  {{user-card user=user}}
+{{/user-list}}
+```
+
+### Component Blocks
+
+Components can be used in two forms, just like regular HTML elements.
+
+**Inline Form**
+
+```app/templates/application.hbs
+{{user-list users=activeUsers}}
+```
+
+This is the basic blockless form of a component, which doesn't give the
+consumer the ability to insert their own content into the given component.
+It's even possible that a component can only be used in this form.
+Make sure to consult a component's documentation for its available uses
+and forms.
+
+**Block Form**
+
+```app/templates/application.hbs
+{{#user-list users=activeUsers}}
+  {{!-- custom template here --}}
+{{/user-list}}
+```
+
+To compose components, we must use the block form, but we must also
+be able to distinguish from within our component which form is being used.
+This can be done with the `hasBlock` property from within our component.
+
+```app/templates/components/user-list.hbs
+{{#if hasBlock}}
+  {{yield}}
+{{else}}
+  <p>No Block Specified</p>
+{{/if}}
+```
+
+Once we know that the component is being used in block form, we can yield
+whatever the user has placed inside the block wherever we want via the `{{yield}}` helper.
+
+This `{{yield}}` helper can be used in the component's template definition multiple times,
+allowing us to create a component that works like a list, where we output a headline and yielded body for each post.
+
+```app/templates/components/post-list.hbs
+{{#each posts as |post|}}
+  <h3>{{post.title}}</h3>
+  <p>{{yield}}</p>
+{{/each}}
+```
+
+Which can be used like so:
+
+```app/templates/posts.hbs
+{{#post-list posts=headlinePosts}}
+  Greatest post ever!
+{{/post-list}}
+```
+
+And will result in the following HTML:
+
+```html
+<div id="ember123" class="ember-view">
+  <h3>Tomster goes to town</h3>
+  <p>Greatest post ever!</p>
+  <h3>Tomster on vacation</h3>
+  <p>Greatest post ever!</p>
+</div>
+```
+
+But what use is it to just output the same thing over and over? Don't we want to customize our posts,
+and display the right content? Sure we do. Lets explore the `{{yield}}` helper a bit.
+
+### Data Down
+
+To accomplish composability beyond simple templates, we need to pass data to those templates. This can be done with the `{{yield}}` helper that was introduced above.
+
+The `{{yield}}` defines where the content we defined inside our component block will yield in the component's layout,
+as we saw in the previous section. Apart from that, the yield helper also allows us to send data down by yielding params back to the scope the component was invoked in.
+
+```hbs
+{{yield}}
+{{yield "post"}}
+{{yield item}}
+{{yield this "footer"}}
+```
+
+By default yield does not send any data, but you can provide an arbitrary number of parameters. Also, the yielded values are ordered sequentially by how they are defined, so you access them in the same order as the component exposes them.
+Sending data down allows the consumer of our component to utilize that data.
+For the consumer to get access to the data, she needs to know what data a component exposes,
+this is where documentation is so crucial, and from there the yielded data can be accessed
+with the `as` operator. Let's take the following template as an example:
+
+```app/components/user-list/template.hbs
+{{#each users as |user|}}
+  {{yield user}}
+{{/each}}
+```
+
+This `{{user-list}}` component will yield as many times as there are users.
+We can consume it in the following way:
+
+```app/users/template.hbs
+{{#user-list users=users as |user|}}
+  {{user-card user=user}}
+{{/user-list}}
+```
+
+Now `{{user-card}}` has access to the current user, which changes as the `{{user-list}}` iterates over each user
+since the `{{yield user}}` is inside the `{{each}}` block. Although this is nice, what if we used our component
+without a block, i.e. `{{user-list users=users}}`? This would make the component pretty useless since nothing is yielding,
+but the users are still being iterated. Let's mix in the `hasBlock` attribute and see if we can make it more useful.
+
+```app/components/user-list/template.hbs
+{{#if hasBlock}}
+  {{#each users as |user|}}
+    {{yield user}}
+  {{/each}}
+{{else}}
+  {{#each users as |user|}}
+    <section class="user-info">
+      <h3>{{user.name}}</h3>
+      <p>{{user.bio}}</p>
+    </section>
+  {{/each}}
+{{/if}}
+```
+
+This makes our component more useful due too it having sane defaults, but it also allows us to override those defaults. By using the `{{yield}}` helper we can pass
+down our params for the consumer to utilize. This makes the concept of passing
+data down very useful.
+
+#### `{{component}}` Helper
+
+To understand the `{{component}}` helper, we will add the following new features to the `{{user-list}}` component:
+
+* For each type of user, we want to use different layout.
+* The list can also be toggled between basic and detailed mode.
+
+Let's add a new type of user, a superuser, which will have different behavior in our application. How would the consumer use this new data to show a different
+UI for each type of user? This is where the `{{component}}` helper comes into play. This helper allows us to render a component chosen at runtime.
+
+Since we have two user types, 'public' and 'superuser', and we want to render
+a component for each type plus for the simple and detailed modes. We want to have the following component names:
+
+* `basic-card-public`
+* `basic-card-superuser`
+* `detailed-card-public`
+* `detailed-card-superuser`
+
+We can create these names by using the `{{concat}}` helper in nested form.
+
+```hbs
+{{component (concat 'basic-card-' user.type)}}
+```
+
+Now that our component names are valid due to the use of dashed, we can put together the full template with the two modes.
+
+```app/users/template.hbs
+{{#user-list users=users as |user basicMode|}}
+  {{#if basicMode}}
+    {{component (concat 'basic-card-' user.type) user=user}}
+  {{else}}
+    {{component (concat 'detailed-card-' user.type) user=user}}
+  {{/if}}
+{{/user-list}}
+```
+
+Now we have `{{basic-card-superuser}}` and `{{detailed-card-public}}` rendering in their respective places
+based on the `user.type` equaling 'superuser' and 'public' respectively. The use of the 'concat' helper in nested form allowed us to accomplish that. Nested helpers are evaluated before the parent helper
+and then that helper can use the value returned by the nested helper. In this case it's just a concatenation of two values. You can
+also experiment with other helpers in nested form, like the `if` helper.
+
+Apart from the use of `{{component}}`, we also have the second yielded value for our basic/detailed mode feature,
+which we've added to our yield, i.e. `{{yield user basic}}`.
+
+The consumer can decide how to name the yielded values. We could have named our yields like so:
+
+```app/users/template.hbs
+{{#user-list users=users as |userModel isBasic|}}
+  {{! something creative here }}
+{{/user-list}}
+```
+
+### Actions Up
+
+Now that we know how to send data down, we probably want to manipulate that data via some user interaction,
+like changing a user's avatar. We can accomplish this by using actions.
+
+We'll be looking at a `{{user-profile}}` component that implements a save action. By yielding the action as a parameter, we allow the consumer to use this component without having to know how to save, but still be able to trigger a save.
+
+Here's our component's definition:
+
+```app/components/user-profile/component.js
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  actions: {
+    saveUser() {
+      var user = this.get('user');
+
+      user.save()
+        .then(() => {
+          this.set('saved', true);
+        })
+        .catch(error => {
+          this.set('saveError', error);
+        });
+    }
+  }
+});
+```
+
+```app/components/user-profile/template.hbs
+{{! most likely we have some markup here }}
+{{yield profile (action "saveUser")}}
+```
+
+So now the consumer will have access to our data and the
+action that we have defined. Lets see how this component could be consumed
+in block form.
+
+```app/templates/user/profile.hbs
+{{#user-profile user=user as |profile saveUser|}}
+  {{user-avatar url=profile.imageUrl onchange=saveUser}}
+{{/user-profile}}
+```
+
+As you can see from this example, the consumer was able to hook into the
+`{{user-profile}}` component's `saveUser` action, which we passed down
+with the `(action "saveUser")` nested helper.  This helper reads a property off the actions hash of the current context, then creates a closure over that function and any arguments you pass.
+
+We could also leverage this format to place actions on native HTML elements
+like an input button:
+
+```hbs
+{{#user-profile user=user as |profile saveUser|}}
+  <button type="button" onclick={{action saveUser}}>Save</button>
+{{/user-profile}}
+```
+
+This is very useful since we are reusing the existing event hooks that
+are provided on these elements. This makes components much more reusable due to how we can compose
+small components that do one thing well.
+
+### Wrapping Up
+
+Composing components is all about isolating functionality into reusable
+chunks that are easy to reason about and easy to combine together so that they work well together.
+This also has the added benefit of easier testing since we try to stay away from
+monolithic components that do everything, and can isolate small parts of functionality.
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -112,6 +112,8 @@
       url: "handling-user-interaction-with-actions"
     - title: "Sending Actions from Components to Your Application"
       url: "sending-actions-from-components-to-your-application"
+    - title: "Composing Components"
+      url: "composing-components"
 
 - title: "Controllers"
   url: 'controllers'


source/components/composing-components.md
```diff
@@ -0,0 +1,108 @@
...
+{{#each item in data}}
+  <h3>{{item.title}}</h3>
+  <p>{{yield}}</p>
+{{/each}}
+```
+
+Which can be used like so:
+
+```hbs
+{{#post-list}}
```
jbrown: Need to bind in some data otherwise it's unclear where those titles are coming from below...

knownasilya: Keep em coming! Would love to make this a super nice guide!

source/components/composing-components.md
```diff
@@ -0,0 +1,143 @@
...
+### Data Down
+
+To accomplish composability beyond just simple templates, we need to pass context to those templates. This can be done with the `{{yield}}` helper.
+
+The `{{yield}}` defines where the template we defined inside our component block will yield in the component's layout,
+as we saw in the previous section. Apart from that, the yield helper also allows us to send data down, providing a context for the templates.
+
+```app/templates/components/my-component.hbs
+{{yield}}
+{{yield 'hello'}}
```
knownasilya: use double quotes

source/components/composing-components.md
```diff
@@ -0,0 +1,143 @@
...
+**Inline Form**
+
+```app/templates/application.hbs
+{{user-list}}
+```
+
+**Block Form**
+
+```app/templates/application.hbs
+{{#user-list}}
```
knownasilya: add data property

source/components/composing-components.md
```diff
@@ -0,0 +1,143 @@
+Components really shine when you use them to their full potential, which is when you compose them.
+Take for example the `<ul>` element, and the fact that only `<li>` elements are appropriate as children.
+If we want the same type of behavior, then we have to compose our components.
+
+Just like we compose regular HTML elements, we can do the same with components.
+
+```app/templates/application.hbs
+{{#user-list sort-by='name' as |user|}}
```
jbrown: Why not camelCased? `sortBy`

source/components/composing-components.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+We can check if `template` is truthy, and if it is that means that the user specified a custom template.
+Well, once we have a template, we probably want to use that in our component, and that's exactly what `{{yield}}` does.
+
+This helper can be used once, or many times. You can make your component into a type of `<ul>` element,
+that is a list that will repeat n times. Like the following example, where we can output a custom summary.
+
+```app/templates/components/post-list.hbs
+{{#each post in posts}}
```
jbrown: `{{each posts as |post|}}` ?

source/components/composing-components.md
knownasilya: Not really sure on the conventions. I think dasherized attrs should be eventually converted to camel-case automatically (hint-hint). The reasoning is that HTML uses dasherized properties.

jbrown: Ah okay if that conversion is going to happen automagically then this makes sense. :+1:

jbrown: @knownasilya This may be a bad example, but do you think it would look strange if you bound data down through multiple components?

```handlebars
{{#user-list sort-by='name' as |user|}}
  {{sort-controls sort-by=sortBy}}
{{/user-list}}
```

knownasilya: Looks good to me, but context should be clear if used as an example. The user should know that `sortBy` is coming from somewhere outside.

jbrown: I think my point was that you have to maintain a mental model.

dasherized => (converted to) camelCased => (bound to) dasherized => (converted to) camelCased => (repeat as nesting composing continues)

knownasilya: But in your example you didn't yield sortBy, so how are you accessing it?

knownasilya: I'll change this to camelCased for now, since we can change it if the feature is in Ember.

mgenev:
I'll review and make suggestions. Thanks for the link to this.

source/components/composing-components.md
```diff
@@ -0,0 +1,159 @@
+Components really shine when you use them to their full potential, which is when you compose them.
+Take for example the `<ul>` element, and the fact that only `<li>` elements are appropriate as children.
+If we want the same type of behavior, then we have to compose our components.
+
+Just like we compose regular HTML elements, we can do the same with components.
+
+```app/templates/application.hbs
+{{#user-list sortBy='name' as |user|}}
```
rwjblue: Need to pass `users` in here:

```
{{#user-list users=model sortBy='name' as |user|}}
```

knownasilya:
I think this is ready cc @rwjblue @trek

knownasilya:
Ping @trek

bf4:
Seems like momentum on this PR has been lost...

knownasilya:
@bf4 I think it's due to the fact that actions have changed in 1.13, so maybe a revision is in order.

knownasilya:
I'll try to tackle the changes this weekend. Would love for some people to be ready to check it once I have those in.

locks:
@knownasilya ping me if/when you do

knownasilya:
Already started, please review up until the CONTINUE HERE section.

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+
+Components can be used in two forms, just like regular HTML elements.
+
+**Inline Form**
+
+```app/templates/application.hbs
+{{user-list users=activeUsers}}
+```
+
+This is the one-off form, or you could call it the default form of a component.
```
locks: `one-off form`?

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+
+```app/templates/components/user-list.hbs
+{{#if hasBlock}}
+  {{yield}}
+{{else}}
+  <p>No Block Specified</p>
+{{/if}}
+```
+
+Once we know that the component is being used in block form, we can yeild
```
locks: typo in `yield`

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+whatever the user has placed inside the block wherever we want via the `{{yield}}` helper.
+
+This helper can be used once, or many times. You can make your component into a type of `<ul>` element,
```
locks: `can be used once, or many times` like the one-off comment, this confuses me

knownasilya:
Thanks for the comments! Will sort them out.

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+But what use is it to just output the same thing over and over? Don't we want to customize our posts,
+and display the right content? Sure we do. Lets explore the `{{yield}}` helper a bit.
+
+### Data Down
+
+To accomplish composability beyond just simple templates, we need to pass context to those templates.
+This can be done with the `{{yield}}` helper that was just introduced above.
+
+The `{{yield}}` defines where the content we defined inside our component block will yield in the component's layout,
+as we saw in the previous section. Apart from that, the yield helper also allows us to send data down, providing a context for
```
locks: Is it sending data down though, if the yielded template is in the same context as the yielding component?

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+the content inside of the component's block. This allows anyone using our component to get context about what's going on
+inside, and to customize the output of the component with that information.
+
+```app/templates/components/my-component.hbs
```
locks: Is this meant to show the different ways you can yield things? I would suggest not putting them under a path'ed code sample, and to preface the code sample with said disclaimer.

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+{{yield item}}
+{{yield this "footer"}}
+{{yield attrs.onchange}}
+```
+
+By default yield does not send any context, but you can provide an arbitrary number of arguments.
+Sending data down allows the consumer of our component to utilize that data.
+For the consumer to get access to the data, she needs to know what data a component exposes,
+this is where documentation is so crucial, and from there the yielded data can be accessed
+with the `as` operator. Let's take `{{yield user "Dummy Text"}}` as an example:
```
locks: You should put the `{{yield` snippet in a path'ed code fence, it might be confusing for beginners otherwise

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+<h3>Profile</h3>
+{{yield "user-avatar" user}}
+{{yield "user-details" user}}
+```
+
+```app/templates/user/profile.hbs
+{{#user-profile user=model as |componentName user|}}
+  {{component componentName user=user}}
+{{/user-profile}}
+```
```
locks: I got lost during this section. How do we go from `user-list`/`user-card` components to `user/profile` route and `user-profile` component?

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+
+With the `{{component}}` helper, we can bind our context to names of components dynamically, which in this case means that we can customize
+the user profile with custom components bound to the relevant data. This means we can have multiple extension points in our components,
+making them much more versatile. Note that the data that is yielded is ordered sequentially by how it's defined, so you
+access it in the same order as the component exposes it. Although, the naming of the variables is entirely up to the consumer, he
```
locks: replace "he" with "they"

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+
+```app/templates/user/profile.hbs
+{{#user-profile user=model as |componentName user|}}
+  {{component componentName user=user}}
+{{/user-profile}}
+```
+
+With the `{{component}}` helper, we can bind our context to names of components dynamically, which in this case means that we can customize
+the user profile with custom components bound to the relevant data. This means we can have multiple extension points in our components,
+making them much more versatile. Note that the data that is yielded is ordered sequentially by how it's defined, so you
```
locks: `Note that the data yielded`

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+access it in the same order as the component exposes it. Although, the naming of the variables is entirely up to the consumer, he
+can choose to be more verbose, or context sensitive with the names.
+
+### Actions Up
+
+Now that we can send data down, we probably want to manipulate that data via some user interaction,
+like changing a user's avatar, or whatever it is you're doing. We can accomplish this by using actions.
```
locks: Remove `or whatever it is you're doing`, the tone could be misconstrued.

source/components/composing-components.md
locks: I also don't know how futureproof `default form of a component` is, given CLI's default template for a component is `{{yield}}`

source/components/composing-components.md
knownasilya: Down to the block/child components/elements.

source/components/composing-components.md
knownasilya: Good point.

source/components/composing-components.md
knownasilya: Will revise and double check that everything is clear. Maybe simplify the example as well.

source/components/composing-components.md
knownasilya: I was referring to the `{{yield}}` helper, and that it can be used in the component's template definition multiple times. Will clarify.

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+Now `{{user-card}}` has access to the current user, which changes as the `{{user-list}}` iterates over each user
+since the `{{yield user "Dummy Text"}}` is inside the `{{each}}` block.
+This opens up the possibility to use the `{{component}}` helper, for example:
+
+```app/templates/components/user-profile.hbs
+<h3>Profile</h3>
+{{yield "user-avatar" user}}
+{{yield "user-details" user}}
```
locks: I got confused by the multiple `{{yield}}`s at first. Did I miss the explanation?

source/components/composing-components.md
```diff
@@ -0,0 +1,202 @@
...
+as we saw in the previous section. Apart from that, the yield helper also allows us to send data down, providing a context for
+the content inside of the component's block. This allows anyone using our component to get context about what's going on
+inside, and to customize the output of the component with that information.
+
+```app/templates/components/my-component.hbs
+{{yield}}
+{{yield "post"}}
+{{yield item}}
+{{yield this "footer"}}
+{{yield attrs.onchange}}
```
locks: `attrs` is being discouraged at the moment, not sure what the idiom is here. cc @mixonic

locks:
Don't forget to rebase

source/components/composing-components.md
knownasilya: Forgot to remove that.

knownasilya:
All set, although I feel like the actions section needs something more..

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+  <p>Greatest post ever!</p>
+  <h3>Tomster on vacation</h3>
+  <p>Greatest post ever!</p>
+</div>
+```
+
+But what use is it to just output the same thing over and over? Don't we want to customize our posts,
+and display the right content? Sure we do. Lets explore the `{{yield}}` helper a bit.
+
+### Data Down
```
pixelhandler: I'm still curious about the term "data down" I don't seen any problem with having a data structure inside a component if it needs one. But the issue is with passing bound data around, data with bindings should only be passed down. Should we consider the distinction between 'data' and 'data bindings' ?

locks:
lgtm, confirm @michaelrkn

michaelrkn:
looks pretty good to me.

source/components/composing-components.md
knownasilya: From experience this depends if the data is isolated to your component, and if it is then go ahead and fetch the data from within your component. Otherwise, pass the data in. 

This also has to do with what kind of API you want the component to have.

trek:
@michaelrkn does this fit into the new TOC? Does this topic really merit a separate guide or should the composition of components be part of any guide on components?

michaelrkn:
imho it feels complex enough to warrant its own page. there might be another way to organize the content, but given that this is written and ready to go and there aren't any definitive plans to do things otherwise, i'd suggest accepting it as-is.

michaelrkn:
@knownasilya What do you think about @trek's suggestion to try to integrate this content into the rest of the pages in the Components section?

knownasilya:
Sounds like it could be done, as long as we know what the TOC is for that section.

michaelrkn:
Yeah, that's a bit tough, because I imagine the Components section will have to be re-thought a bit when routable components lands. I'd lean towards just incorporating your ideas into the guides as they are now and not worry too much about the future, because the future is uncertain and your ideas are good :) But maybe I am not enough of a visionary. @trek, do you have any guidance here?

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+  <p>No Block Specified</p>
+{{/if}}
+```
+
+Once we know that the component is being used in block form, we can yield
+whatever the user has placed inside the block wherever we want via the `{{yield}}` helper.
+
+This `{{yield}}` helper can be used in the component's template definition multiple times,
+allowing us to create a component that works like a list, where the user insert/modify custom content
+at every yield. Take the following example, where we output a custom summary.
```
mixonic: That this first example uses `#each` and `yield` together is a bit intense.

At the least I suggest `where we output a headline and yielded body for each post.`

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+</div>
+```
+
+But what use is it to just output the same thing over and over? Don't we want to customize our posts,
+and display the right content? Sure we do. Lets explore the `{{yield}}` helper a bit.
+
+### Data Down
+
+To accomplish composability beyond just simple templates, we need to pass context to those templates.
+This can be done with the `{{yield}}` helper that was just introduced above.
```
mixonic: Zero uses of the work `just` would be better than two in this paragraph :-) I always write `just` into my content then end up editing it out. It just sounds dismissive of the reader. Right?

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+
+The `{{yield}}` defines where the content we defined inside our component block will yield in the component's layout,
+as we saw in the previous section. Apart from that, the yield helper also allows us to send data down, providing a context for
+the content inside of the component's block. This allows anyone using our component to get context about what's going on
+inside, and to customize the output of the component with that information.
```
mixonic: That phrase `context` might be better as `params` here. `context` of a template means something fairly specific, the `this` of a template. We definitely call it `context` elsewhere.

Starting to introduce "block params" by talking about `yielding params back to the scope the component was invoked in` seems good here.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+  {{#each users as |user|}}
+    <section class="user-info">
+      <h3>{{user.name}}</h3>
+      <p>{{user.bio}}</p>
+    </section>
+  {{/each}}
+{{/if}}
+```
+
+Now we our component is a little more useful, because it has sane defaults, but it also allows us to override
```
mixonic: `Now we our`

Could also use to have less commas in this sentence, maybe avoid `because` with rephrasing

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+those defaults. I'm sure there is more that we can do to our `{{user-list}}` component to make it more useful,
+like handling different types of users.
```
mixonic: > I'm sure there is **more** that we can do to our `{{user-list}}` component to make it **more** useful

I'm not sure this sentence adds anything, it can likely be removed or replaces with a summary of what was done in this section if the goal is to summarize.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### `{{component}}` Helper
+
+Now we need to add a new type of user, a superuser, which will have different behavior in our
```
mixonic: `Now we need` -> `Let's add`. There is nothing compelling the user forward, no need.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+application. How would the consumer use this new data to show a different UI for each type of user?
+This is where the `{{component}}` helper comes into play. This helper allows us to render a component
+based on an arbitrary value. Along with this new feature request, we were also asked to add a toggle
```
mixonic: > This helper allows us to render a component based on an arbitrary value

-> This helper allows us to render a component chosen at runtime

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+between a basic and a detailed view, which we need to expose in our block.
```
mixonic: If there are two things you want to add as requirements, I suggest introducing both of them before talking about the `{{component` helper or any code. And try a list:

> To introduce the `{{component` helper, we will add to new behaviors to our `user-list` component:
>
> * For each type of user, we want to use different layout.
> * Each row can be toggled between displaying basic user information (name and bio) and detailed user information.
>

etc ...

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+
+Since components require a dash in the component name, e.g. `basic-card-superuser`,
+the simplest way would be to use a helper to concatenate 'basic-card-' and `user.type`.
```
mixonic: `would be` -> `is`

I might suggest listing out what all the component names are first, then showing how you might use concat to build those. "the simplest way is to use a helper to concatenate 'basic-card-' and `user.type`" is slightly cruel- it introduces helpers, the `concat` helper (is that public?), and nested helpers in one go. Does not seem simple.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+    {{component (concat 'basic-card-' user.type) user=user}}
+  {{else}}
+    {{component (concat 'detailed-card-' user.type) user=user}}
+  {{/if}}
+{{/user-list}}
+```
+
+Now we have `{{basic-card-superuser}}` and `{{detailed-card-public}}` rendering in their respective places
+based on the `user.type` equaling 'superuser' and 'public' respectively. Here we used the 'concat' helper
+as a subexpression inside of our `{{component}}` helper. Subexpressions are evaluated before the helper
```
mixonic: Please use the phrase `nested helper` instead of `subexpression`. We believe `nested helper` may be more friendly to those not brainwashed by the htmlbars hype ;-)

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+and then the helper can use the value, in this case it's just a concatenation of two values. You can
+also experiment with other helpers as subexpressions, like the `if` helper.
+
+Apart from the use of `{{component}}`, we also have the second yielded value for our basic/detailed mode feature,
+which we've added to our yield, i.e. `{{yield user basic}}`.
+
+Something to remember is that the yielded values are ordered sequentially by how they are defined, so you
```
mixonic: I think you mention that they are positional parameters above when you introduce valid usages of `{{yield}}` maybe that introduction is a better place to make clear you can pass any number of params.

source/components/composing-components.md
knownasilya: Glad you said this, was thinking it when writing.

source/components/composing-components.md
knownasilya: Good point.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+
+The consumer can decide how to name the yielded values. We could have named our yields like so:
+
+```app/users/template.hbs
+{{#user-list users=users as |userModel isBasic|}}
+  {{! something creative here }}
+{{/user-list}}
+```
+
+Just remember to respect the order that they are yielded.
```
mixonic: Seems unnecessary.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Actions Up
+
+Now that we know how to send data down, we probably want to manipulate that data via some user interaction,
+like changing a user's avatar. We can accomplish this by using actions.
+
+We'll be looking at a `{{user-profile}}` component that implements a save action, which we
+can yield to the block. This will allow a consumer to use this component without having to know how to save,
```
mixonic: > which we can yield to the block

Seems a strange phrase? "the block" "we can yield"?

knownasilya:
@mixonic thanks for the comments.

source/components/composing-components.md
mixonic: In general I think you can better introduce the functionality here.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+{{#user-profile user=user as |profile saveUser|}}
+  {{user-avatar url=profile.imageUrl onchange=saveUser}}
+{{/user-profile}}
+```
+
+As you can see from this example, the consumer was able to hook into the
+`{{user-profile}}` component's `saveUser` action, which we passed down
+with the `(action "saveUser")` subexpression. This subexpression allows us
+to bundle the action and the target into one value so we don't have to
+specify a `targetObject` property.
```
mixonic: ~~subexpression~~, ~~targetObject~~ I don't think they will knoe what that means.

`(action` reads a property off the actions hash of the current context, then creates a closure over that function and any arguments you pass.

source/components/composing-components.md
```diff
@@ -0,0 +1,260 @@
...
+{{/user-profile}}
+```
+
+This is very useful since we are reusing the existing event hooks that
+are provided on these elements. This makes components much more reusable due to how we can compose
+small components that do one thing well.
+
+### Wrapping Up
+
+So remember that composing components is all about isolating functionality into reusable
```
mixonic: ~~So remember that~~

mixonic:
@knownasilya sorry this took me so long to look at. I am in theory :+1: on a merge if we get it polished up. @trek gets veto power if he wants to wait on an alternative TOC, but the worst that happens is that is gets refactored into that new TOC when it arrives IMO.

knownasilya:
Will have a look at the comments and make the changes as soon as I can, hopefully tonight.

source/components/composing-components.md
knownasilya: Yeah removing it works just fine. Thanks.

knownasilya:
@mixonic @michaelrkn all set

source/components/composing-components.md
```diff
@@ -0,0 +1,261 @@
...
+
+But what use is it to just output the same thing over and over? Don't we want to customize our posts,
+and display the right content? Sure we do. Lets explore the `{{yield}}` helper a bit.
+
+### Data Down
+
+To accomplish composability beyond simple templates, we need to pass data to those templates. This can be done with the `{{yield}}` helper that was introduced above.
+
+The `{{yield}}` defines where the content we defined inside our component block will yield in the component's layout,
+as we saw in the previous section. Apart from that, the yield helper also allows us to send data down by yielding params back to the scope the component was invoked in.
```
mixonic: `as we saw in the previous section` reads awkwardly.

mixonic:
:+1: good improvements! Thanks for pushing on this.

locks:
:cookie: for you @knownasilya :) thanks

knownasilya:
Thanks all for proof reading, couldn't have done it without you!

PR #67, opened 2015-03-19
Changed double quotes to single quotes to avoid confusion on View - INSE...

bradware:
@locks do you mind taking a look at this pr?


locks:
@bware225 I'm actually not sure what the styleguide is here. /ping @trek

trek:
We should standardize on one for sure, but I'm not sure what confusion having both right now would cause.

bware225:
@trek For JS beginners like myself, I was confused if one was needed over the other in certain situations. Not a huge deal just figured it would be good to pick one

locks:
@bradware since the views section of the guide is gone can you close this? sorry it didn't make it in time!

michaelrkn:
This can be closed. cc @rwjblue

PR #73, opened 2015-03-21
Add fenced code block with language wherever necessary.

tundal45:
Currently, you building the guides on master fails because guides have code either not fenced in blocks or don't have language specified. This fixes the guides so that they build. https://github.com/emberjs/guides/pull/53 will add better error message for this situation in the future and will also run both the tests & the build the guide as a part of CI build.


source/templates/development-helpers.md
```diff
@@ -35,7 +35,7 @@ For example, if you are wondering why a specific variable isn't displaying in
 your template, you could use the `{{debugger}}` helper. When the breakpoint is
 hit, you can use the `templateContext` in your console to lookup properties:
 
-```
+```javascript
```
tundal45: This is technically console but `javascript` felt right.

trek:
Bwahahaha. Was there only 1?

PR #74, opened 2015-03-21
restore getting ember page

michaelrkn:
it was accidentally deleted by #65


trek:
@locks was it accidental? I don't we want to support (in the main guides) a way of getting ember except via the cli/bower.

locks:
@trek I thought I'd remove it since it's already addressed in https://github.com/emberjs/guides/pull/7#issuecomment-84101952, and then move that section to where most appropriate.

michaelrkn:
Oh oops. I just cloned the repo for the first time yesterday, saw a 404 when I clicked the first "Next up" link, and assumed it was an accident. Looks like the update to add the new install instructions got merged in a bit after I ran into this :)

PR #75, opened 2015-03-21
Clean redo of https://github.com/emberjs/guides/pull/7

bsclifton:
1) Added Getting Started section which has instructions on how to install Ember using NPM and also contains a glossary
2) Updated styles to fix how the guides look on Windows (the font in the sidebar was too large and looked bad on Windows)

### DIFF
--- a/source/ember-cli/index.md
+++ b/source/ember-cli/index.md
@@ -1,0 +1,58 @@
+Installing Ember is *easy*! And our new install process includes *Ember CLI*, Ember's build tool
+
+* State of the art asset management (including combining, minifying, and versioning)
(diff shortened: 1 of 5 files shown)


source/ember-cli/index.md
```diff
@@ -0,0 +1,58 @@
+Installing Ember is *easy*! And our new install process includes *Ember CLI*, Ember's build tool
```
rwjblue: I'm not a huge fan of new here.  In 6 months, this guide will still be live and it will not be "new"....

bsclifton: With versioned guides, it's no biggie- next version we can remove the new and even some of the enthusiasm too (I may have went overboard with the italics and exclamations)

PR #76, opened 2015-03-22
Turn cli section into getting started

trek:


trek:
This is on 1.11 already.

PR #78, opened 2015-03-22
Clean up install instructions

michaelrkn:
I've done a few things here:

* Cleaned up the wording to make the instructions more to the point
* Fixed grammar
* Moved the instructions for Node to the top, so that people who don't have Node see them before they run into problems
* Added a note about installing Watchman
* A few other minor cleanup things

Let me know if this looks okay!


michaelrkn:
Too out-of-date; closing. Maybe we can make this better some day.

PR #83, opened 2015-03-24
Improve documentation about `action` `on` attr.

Serabe:
Addresses the problem expressed in emberjs/ember.js#10713


locks:
@Serabe I believe this PR is wrong, you should instead be updating the actions section to use dasherized names.

locks:
cc @michaelrkn

michaelrkn:
@locks I believe you are correct - see for example http://emberjs.com/api/classes/Ember.TextSupport.html. If so, this PR should be closed. cc @rwjblue @trek

Serabe:
Then, [`action`helper documentation](http://emberjs.com/api/classes/Ember.Handlebars.helpers.html#method_action) needs to be updated as well, since I took the information from there and it still wrong. Not sure if I documentation PRs are accepted in ember.js, though.

michaelrkn:
@Serabe good catch. It's probably worth opening a PR to see. https://github.com/emberjs/ember.js/blob/v1.13.3/packages/ember-routing-htmlbars/lib/keywords/action.js#L9

Serabe:
@michaelrkn, thanks. I'll open another PR for the guides too.

PR #86, opened 2015-03-25
Remove getting ember from nav

joostdevries:
Gives a 404 right now


trek:
Why not just remove?

joostdevries:
saw a lot of other stuff commented out but I since these pages no longer exists I guess removing makes more sense. fixed.

PR #89, opened 2015-03-25
Added "Building your app"

joostdevries:
Briefly explains how to create their first app and points to ember-cli.com for advanced usage.


michaelrkn:
This is great! I think we should make it a lot more fleshed out :) Instead of saying "Run `ember help generate`", I think we should talk through each of the generators and what they do.

joostdevries:
Agree but that could be done in a next iteration/separate pr.

joostdevries:
I would be *really* happy if this section makes it in before the next release

michaelrkn:
Makes sense!

source/ember-cli/building-your-app.md
```diff
@@ -0,0 +1,26 @@
+The Ember guides will take you through the various aspects of your applciation and the framework. To make your work easy, we've got two tips for you:
+
+## 1. Use the built-in code-generators
+
+Ember is a framework that favors Convention over Configuration. Because you might not now all conventions straight away, we've made sticking to them very easy by providing code generation for practically every type of module your app could need. To get an overview of all, run:
```
knownasilya: 'now' should be 'know'

source/ember-cli/building-your-app.md
```diff
@@ -0,0 +1,26 @@
...
+
+```bash
+ember help generate
+```
+
+## 2. Use ES6 modules and imports properly
+
+Ember uses [ES6 modules](http://eviltrout.com/2014/05/03/getting-started-with-es6.html) to organize your application's code. When you're not using the built-in generators, make sure you're always importing any module you're using. This includes importing `Ember` from the 'ember' module and `DS` from the 'ember-data' packages. Failing to do so will result in errors in your console (`Ember is not defined`).
```
knownasilya: Maybe also link to http://jsmodules.io/

joostdevries:
@trek @locks updated.

joostdevries:
also see #415

michaelrkn:
@joostdevries can you close this in favor of the approach we discussed on #415?

PR #92, opened 2015-03-26
Fixing missing Ember. variable

aaronbhansen:
Was following the example locally and noticed you need Ember.Handlebars.Utils


rwjblue:
Thanks!

PR #94, opened 2015-03-27
Fixed version picker for IE

bsclifton:
Fixed a bug w/ the version selection (was causing error in IE; the select object doesn't have a selectedOptions property in IE).  Also includes updates to the README for Windows devs.


trek:
Can you PR just the bugfix against v1.10.0 and v1.11.0 as well. Thanks.

bsclifton:
Done :smile:

PR #102, opened 2015-03-27
Update classes-and-instances.md

stefanpenner:
missing super yo + quick TL;DR of when to call super

### DIFF
--- a/source/object-model/classes-and-instances.md
+++ b/source/object-model/classes-and-instances.md
@@ -97,6 +97,7 @@ instances:
 ```js
 Person = Ember.Object.extend({
   init: function() {
+    this._super.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


source/object-model/classes-and-instances.md
```diff
@@ -111,10 +112,13 @@ Person.create({
 
 If you are subclassing a framework class, like `Ember.View` or
 `Ember.ArrayController`, and you override the `init` method, make sure
-you call `this._super()`! If you don't, the system may not have an
+you call `this._super.apply(this, arguments)`! If you don't, the system may not have an
```
trek: "!" is probably overkill.

trek:
This went stale after feedback. Closing for now.

PR #105, opened 2015-03-27
Removing a " in the ES6 link in the glossary

tommyp:
There was a `"` at the end of the URL which made the link incorrect. This commit removes it, fixing the URL.


tommyp:
I'm not sure why this has 2 additions/deletions. I removed that one character using the Github.com editor so ¯\\\_(ツ)_/¯

rwjblue:
Thanks!

PR #110, opened 2015-03-27
Encourage decorator-style Ember.computed/Ember.observer

stefanpenner:
lets encourage best-practices.
TL;DR we want to move away from prototype extensions.

So for today, lets tackle two simple ones.
- `.property` -> `Ember.computed`
- `.observes` -> `Ember.observer`

Further discussions about all other PROTOTYPE_EXTENSIONS, are more complicated and as they may result in development effort. They will go through the formal RFC process.

Topics for later discussion:

- [ ] `Ember.A`
- [ ] `observer(..., fn).on`

[edit: wording changes, based on feedback]

### DIFF
--- a/source/concepts/what-is-ember-js.md
+++ b/source/concepts/what-is-ember-js.md
@@ -104,11 +104,11 @@ var president = Ember.Object.create({
   firstName: "Barack",
   lastName: "Obama",
 
-  fullName: function() {
+  fullName: Ember.computed(function() {
     return this.get('firstName') + ' ' + this.get('lastName');
 
   // Call this flag to mark the function as a property
-  }.property()
+  })
 });
 
 president.get('fullName');
@@ -124,15 +124,18 @@ You can tell Ember about these dependencies like this:
 
 ```javascript
 var president = Ember.Object.create({
-  firstName: "Barack",
-  lastName: "Obama",
+  init: function() {
+    this._super.apply(this, arguments);
+    this.set('firstName', 'Barack');
+    this.set('lastName', 'Obama');
+  },
 
-  fullName: function() {
+  fullName: Ember.computed('firstName', 'lastName', function() {
     return this.get('firstName') + ' ' + this.get('lastName');
 
-  // Tell Ember that this computed property depends on firstName
-  // and lastName
-  }.property('firstName', 'lastName')
+    // Tell Ember that this computed property depends on firstName
+    // and lastName
+  })
 });
 ```
 
--- a/source/cookbook/working_with_objects/continuous_redrawing_of_views.md
+++ b/source/cookbook/working_with_objects/continuous_redrawing_of_views.md
@@ -28,18 +28,22 @@ another interval is triggered each time the property increases.
 
 ```app/services/clock.js
 export default Ember.Object.extend({
-    pulse: Ember.computed.oneWay('_seconds').readOnly(),
-    tick: function () {
-      var clock = this;
-      Ember.run.later(function () {
-        var seconds = clock.get('_seconds');
-        if (typeof seconds === 'number') {
-          clock.set('_seconds', seconds + (1/4));
-        }
-      }, 250);
-    }.observes('_seconds').on('init'),
-    _seconds: 0
-  });
+  init: function() {
+    this._super.apply(this, arguments);
+    this._seconds = 0;
+  },
+
+  pulse: Ember.computed.oneWay('_seconds').readOnly(),
+  tick: Ember.observer('_seconds', function () {
+    var clock = this;
+    Ember.run.later(function () {
+      var seconds = clock.get('_seconds');
+      if (typeof seconds === 'number') {
+        clock.set('_seconds', seconds + (1/4));
+      }
+    }, 250);
+  }).on('init')
+});
 ```
 
 ### Binding to the `pulse` attribute
@@ -69,20 +73,20 @@ a handful of properties used as conditions in the Handlebars template.
 
 ```app/controllers/interval.js
 export default Ember.ObjectController.extend({
-    secondsBinding: 'clock.pulse',
-    fullSecond: function () {
-      return (this.get('seconds') % 1 === 0);
-    }.property('seconds'),
-    quarterSecond: function () {
-      return (this.get('seconds') % 1 === 1/4);
-    }.property('seconds'),
-    halfSecond: function () {
-      return (this.get('seconds') % 1 === 1/2);
-    }.property('seconds'),
-    threeQuarterSecond: function () {
-      return (this.get('seconds') % 1 === 3/4);
-    }.property('seconds')
-  });
+  secondsBinding: 'clock.pulse',
+  fullSecond: Ember.computed('seconds', function () {
+    return (this.get('seconds') % 1 === 0);
+  }),
+  quarterSecond: Ember.computed('seconds', function () {
+    return (this.get('seconds') % 1 === 1/4);
+  }),
+  halfSecond: Ember.computed('seconds', function () {
+    return (this.get('seconds') % 1 === 1/2);
+  }),
+  threeQuarterSecond: Ember.computed('seconds', function () {
+    return (this.get('seconds') % 1 === 3/4);
+  })
+});
 ```
 
 A controller for a list of comments, each comment will have a new clock
--- a/source/cookbook/working_with_objects/continuous_redrawing_of_views.md
+++ b/source/cookbook/working_with_objects/continuous_redrawing_of_views.md
@@ -27,6 +27,9 @@ of the interval. Since the `tick` method observes the incremented property
 another interval is triggered each time the property increases.
 
 ```app/services/clock.js
+var observer = Ember.observer;
+var on = Ember.on;
+
 export default Ember.Object.extend({
   init: function() {
     this._super.apply(this, arguments);
@@ -34,7 +37,7 @@ export default Ember.Object.extend({
   },
 
   pulse: Ember.computed.oneWay('_seconds').readOnly(),
-  tick: Ember.observer('_seconds', function () {
+  tick: on(observer('_seconds', function () {
     var clock = this;
     Ember.run.later(function () {
       var seconds = clock.get('_seconds');
@@ -42,7 +45,7 @@ export default Ember.Object.extend({
         clock.set('_seconds', seconds + (1/4));
       }
     }, 250);
-  }).on('init')
+  }), 'init')
 });
 ```
 
@@ -96,7 +99,7 @@ comment was created.
 
 ```app/controllers/comment-item.js
 export default Ember.ObjectController.extend({
-    seconds: Ember.computed.oneWay('clock.pulse').readOnly()
+  seconds: Ember.computed.oneWay('clock.pulse').readOnly()
 });
 ```
 
@@ -104,18 +107,18 @@ export default Ember.ObjectController.extend({
 import ClockService from '../services/clock';
 
 export default Ember.ArrayController.extend({
-    itemController: 'commentItem',
-    comment: null,
-    actions: {
-      add: function () {
-        this.addObject(Em.Object.create({
-          comment: this.get('comment'),
-          clock: ClockService.create()
-        }));
-        this.set('comment', null);
-      }
+  itemController: 'commentItem',
+  comment: null,
+  actions: {
+    add: function () {
+      this.addObject(Em.Object.create({
+        comment: this.get('comment'),
+        clock: ClockService.create()
+      }));
+      this.set('comment', null);
     }
-  });
+  }
+});
 ```
 
 ### Handlebars template which displays the `pulse`
--- a/source/controllers/representing-a-single-model-with-objectcontroller.md
+++ b/source/controllers/representing-a-single-model-with-objectcontroller.md
@@ -84,13 +84,13 @@ format for the template:
 
 ```app/controllers/song.js
 export default Ember.ObjectController.extend({
-  duration: function() {
-    var duration = this.get('model.duration'),
-         minutes = Math.floor(duration / 60),
-         seconds = duration % 60;
+  duration: Ember.computed('model.duration', function() {
+    var duration = this.get('model.duration');
+    var minutes = Math.floor(duration / 60);
+    var seconds = duration % 60;
 
     return [minutes, seconds].join(':');
-  }.property('model.duration')
+  })
 });
 ```
 
--- a/source/controllers/representing-multiple-models-with-arraycontroller.md
+++ b/source/controllers/representing-multiple-models-with-arraycontroller.md
@@ -34,12 +34,12 @@ property called `longSongCount` to the controller:
 
 ```app/controllers/songs.js
 export default Ember.ArrayController.extend({
-  longSongCount: function() {
+  longSongCount: Ember.computed('@each.duration', function() {
     var longSongs = this.filter(function(song) {
       return song.get('duration') > 30;
     });
     return longSongs.get('length');
-  }.property('@each.duration')
+  })
 });
 ```
 
@@ -75,11 +75,9 @@ creating an `ObjectController`:
  
 ```app/controllers/song.js
 export default Ember.ObjectController.extend({
-  fullName: function() {
- 
+  fullName: Ember.computed('name', 'artist', function() {
     return this.get('name') + ' by ' + this.get('artist');
- 
-  }.property('name', 'artist')
+  })
 });
 ```
  
--- a/source/cookbook/helpers_and_components/adding_google_analytics_tracking.md
+++ b/source/cookbook/helpers_and_components/adding_google_analytics_tracking.md
@@ -40,12 +40,12 @@ var Router = Ember.Router.extend({
 });
 
 Router.reopen({
-  notifyGoogleAnalytics: function() {
+  notifyGoogleAnalytics: Ember.on('didTransition', function() {
     return ga('send', 'pageview', {
         'page': this.get('url'),
         'title': this.get('url')
       });
-  }.on('didTransition')
+  })
 });
 
 export default Router;
--- a/source/cookbook/user_interface_and_interaction/adding_css_classes_to_your_components_based_on_properties.md
+++ b/source/cookbook/user_interface_and_interaction/adding_css_classes_to_your_components_based_on_properties.md
@@ -17,9 +17,9 @@ You can also set the class name based on a computed property.
 
 ```js
 classNameBindings: ['isActive'],
-isActive: function() {
+isActive: Ember.computed('someAttribute', function() {
   return 'active';
-}.property('someAttribute')
+})
 ```
 
 Another way would be to bind the class name to a bound property.
--- a/source/cookbook/user_interface_and_interaction/basic_form_validations.md
+++ b/source/cookbook/user_interface_and_interaction/basic_form_validations.md
@@ -16,11 +16,12 @@ been focused. The component expects to get the validation result as the
 export default Ember.Component.extend({
   beenFocused: false,
   valid: null,
-  hasError: function() {
+  hasError: Ember.computed('valid', 'beenFocused', function() {
     if (this.get('beenFocused')) {
       return !this.get('valid');
     }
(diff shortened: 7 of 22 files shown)


eviltrout:
Really? Every ember project I've browsed seems to use them for convenience? When did the tide change?

balinterdi:
@stefanpenner Should prototype extensions be avoided because they introduce a performance penalty (or in plain English: because they are slow)? I guess the same then goes for `observes`, which are best avoided anyway when possible.

stefanpenner:
> Really? Every ember project I've browsed seems to use them for convenience? When did the tide change?

this is part of a larger effort to kill prototype extensions in general. Basically, its extremely dubious to pollute globals as we currently do.

Also, ES7 will give us annotations, which partially unlocks our path to using the new Class syntax provided by ES6. Transitioning from the `computed('foo', 'bar', fn)` syntax to the proposed annotation syntax is super mechanical.

> @stefanpenner Should prototype extensions be avoided because they introduce a performance penalty (or in plain English: because they are slow)? I guess the same then goes for observes, which are best avoided anyway when possible.

nah, no performance difference for these prototype extensions.

cibernox:
Will we need to use Ember.A() all over the place?

tomdale:
This is a very big change and I think we need to be a little bit more deliberate in the messaging, *if* we want to make this recommendation at all.

I personally find code written with prototype extensions easier to write and parse. Also remember that disabling prototype extensions has a significant performance cost due to the need to `Ember.A()` observed arrays. That said, prototypes _can_ cause compatibility issues and I am always happy to revisit the decision if the ecosystem has sufficiently changed.

Personally, I would prefer that we frame this less as "let's stop using prototype extensions" and more "let's start aligning our syntax with ES7". Ideally, we could migrate folks over to the annotations API directly with Babel, and only have one mass migration instead of two.

stefanpenner:
> Will we need to use Ember.A() all over the place?

for now, some future RFC process will likely dig into that in more detail.

The above changes, `Ember.computed` over `function(){}.property()` are already supported, and have zero additionally side-affects.

stefanpenner:
> I personally find code written with prototype extensions easier to write and parse. Also remember that disabling prototype extensions has a significant performance cost due to the need to Ember.A() observed arrays.

yes, a more detailed RFC can tackle this one. `Ember.A()` is a special case, that will require more effort and energy. But this puts us on a better path

tomdale:
@stefanpenner Suggesting that everyone `A()` arrays is very large change that will trip up new developers and has the potential to cause performance regressions. I am :-1: on this change which seems hasty; I would be interested in a more thorough transition plan that takes these externalities into account.

> Ember.A() is a special case, that will require more effort and energy. But this puts us on a better path

I think this incremental step will cause more chaos than whatever benefits it provides, which are IMO pretty marginal.

stefanpenner:
> This is a very big change and I think we need to be a little bit more deliberate in the messaging, if we want to make this recommendation at all.

I argue that this isn't a big change, this is merely encouraging a different syntax that we already support.

stefanpenner:
> @stefanpenner Suggesting that everyone A() arrays is very large change that will trip up new developers and has the potential to cause performance regressions. I am  on this change which seems hasty; I would be interested in a more thorough transition plan that takes these externalities into account.

This PR has nothing to do with this. We are merely encouraging using `computed` and `observer` over there prototype extension variants

tomdale:
> This PR has nothing to do with this. We are merely encouraging using computed and observer over there prototype extension variants

Ah, I see. Like I said above, I would rather just wait until we have annotations in Babel (which seems very close) and have people migrate en masse at once, rather than have an annoying transition period where new developers will find examples with `.property()`, `computed()` and the ES7 annotation syntax.

stefanpenner:
> Ah, I see. Like I said above, I would rather just wait until we have annotations in Babel (which seems very close) and have people migrate en masse at once, rather than have an annoying transition period where new developers will find examples with .property(), computed() and the ES7 annotation syntax.

this already exists, ^^ helps move people towards just `computed` rather then all 3 syntaxes.

cibernox:
In fact I already favour Ember.computed/Ember.observer over the prototype extensions, but the observable interface un EmberArray is massively convenient. As Tom said, it might scare newcomers

stefanpenner:
> but the observable interface un EmberArray is massively convenient

this effort does not touch this, some future investigation can dig into those.

rwjblue:
>  I personally find code written with prototype extensions easier to write and parse.

@tomdale - I completely disagree.  It is **much** easier to read when you know the context (normal function, computed property, or observer) of a function before you get to the end.  With `Ember.computed('someProp', function() { });` the reader immediately knows that it is a computed property **AND** what dependent keys will cause invalidation.  Whereas with `function() { }.property('someProp')` you have to read the entire function (or skip to the end and then start back at the beginning) before you know it is a computed with specific DK's.

karlfreeman:
For the computed properties change would it not be best to start off with re-writing the [computed properties section](https://github.com/emberjs/guides/blob/master/source/object-model/computed-properties.md) first? Potentially providing a clear explanation as to why its a "best-practice" going forward? and a nod towards the _current_ way people are writing it? (e.g removing the seciont on [alternate invocation](https://github.com/emberjs/guides/blob/master/source/object-model/computed-properties.md#alternate-invocation) and providing a replacement for the prototyped way of doing it)

stefanpenner:
> For the computed properties change would it not be best to start off with re-writing the computed properties section first?

we can do that as part of this PR

karlfreeman:
@stefanpenner Sure, I guess in my eyes it might be a better straw man to knock down than just the syntax update. :+1:.

stefanpenner:
All this input is great, tweet + pr is succeeding in getting eyes and feedback!

karlfreeman:
@stefanpenner The more I cruise through the docs, the more I realise theres a general inconsistency with how prototype extensions are mentioned / referenced. Perhaps a PR before this one could be a tightening up on that front. [computed properties' "alternative invocation"](https://github.com/emberjs/guides/blob/master/source/object-model/computed-properties.md#alternate-invocation) vs [observer's "without prototype extensions"](https://github.com/emberjs/guides/blob/master/source/object-model/observers.md#without-prototype-extensions). Which ideally would link back to the [disabling prototype extensions](https://github.com/emberjs/guides/blob/master/source/configuring-ember/disabling-prototype-extensions.md) section.

stefanpenner:
> Perhaps a PR before this one could be a tightening up on that front. computed properties' "alternative invocation" vs observer's "without prototype extensions".

I'm curious why this that can't be single consolidated effort? That would help ensure the tone/intent is unified on this topic.

karlfreeman:
@stefanpenner Its close as it [stands right now](https://github.com/emberjs/guides/blob/master/source/configuring-ember/disabling-prototype-extensions.md#functions) so I think its a matter of simply moving the two sections in the computed properties and observer pages into one place. Instead of trying to explain them differently in different places?

stefanpenner:
> @stefanpenner Its close as it stands right now so I think its a matter of simply moving the two sections in the computed properties and observer pages into one place. Instead of trying to explain them differently in different places?

that sounds good, I will try to word smith after my next two meetings. If you have any ideas feel free to submit a PR to my branch, or leave words here in comment.

karlfreeman:
@stefanpenner Both pages could then signpost towards a single place for an _potential_ "upcoming" syntax change? Or not ;).

stefanpenner:
> Both pages could then signpost towards a single place for an potential "upcoming" syntax change? Or not ;).

both syntaxes are currently available, this merely demonstrates that one that enables ember to be a better citizen in the JS ecosystem.

martndemus:
👍, the proposed form reads much better. 

Also, If we do `observer` and `computed`, please also consider including `on`!

bcardarella:
At DY we've worked without the prototype extensions for several projects and personally I find it to be more readable. However, I agree with @tomdale with the issues around Ember.Array. Personally I suspect this will increase the WTFs/minute for people getting into Ember. With the effort around 2.x being to lower the barrier of entry it seems odd to then take a step backwards in this regard.

It seems that most everyone here is in favor of suggesting the `Function.prototype` extensions not be used. Is there a compromise where we can selectively turn on/off certain extensions, or extensions per type, rather than all or nothing? I realize there is a desire to avoid collisions with other libraries but as a first step this might be the best path.

thejameskyle:
Dropping this here for those out of the loop: https://github.com/wycats/javascript-decorators

This would be the (most-likely) ES7-aligned syntax:

```js
import {View, computed} from 'ember';

export default class MyView extends View {
  @computed('foo')
  get bar(foo) {
    return foo + 'bar';
  } 
}
```

The decorators would then get transpiled by Babel to this:

```js
_createDecoratedClass(MyView, [{
  key: "bar",
  decorators: [computed("foo")],
  get: function (foo) {
    return foo + "bar";
  }
}]);
```

source/concepts/what-is-ember-js.md
```diff
@@ -104,11 +104,11 @@ var president = Ember.Object.create({
   firstName: "Barack",
   lastName: "Obama",
 
-  fullName: function() {
+  fullName: Ember.computed(function() {
     return this.get('firstName') + ' ' + this.get('lastName');
 
   // Call this flag to mark the function as a property
```
kimroen: This comment isn't accurate now.

source/concepts/what-is-ember-js.md
```diff
@@ -124,15 +124,18 @@ You can tell Ember about these dependencies like this:
...
 
-  fullName: function() {
+  fullName: Ember.computed('firstName', 'lastName', function() {
     return this.get('firstName') + ' ' + this.get('lastName');
 
-  // Tell Ember that this computed property depends on firstName
-  // and lastName
-  }.property('firstName', 'lastName')
+    // Tell Ember that this computed property depends on firstName
+    // and lastName
```
kimroen: Same here.

source/concepts/what-is-ember-js.md
```diff
@@ -124,15 +124,18 @@ You can tell Ember about these dependencies like this:
 
 ```javascript
 var president = Ember.Object.create({
-  firstName: "Barack",
-  lastName: "Obama",
+  init: function() {
+    this._super.apply(this, arguments);
+    this.set('firstName', 'Barack');
```
cibernox: Is this kind of initialization encouraged? I thought that the *shape deoptimization* happened when defining attributes at a class level (using `Ember.Object.extend()`, but defining attributes with `create` was fine.

Kuzirashi:
I can only say that people who use CoffeeScript are already familiar with `Ember.computed` instead of `property` in whole projects as it's considered best practice in CS. However I didn't know that one can use `Ember.observer().on()` and when I needed to chain `observer` with `on` I've always used `(->).observes().on()`.

chadhietala:
All reviews I do, I tell people not to use the prototype extension. By and large people on my team feel it's much more clear; and becomes more clear with the new syntax, that the scan line is much more clearer.

Since `ArrayControllers` and `ArrayProxies` will go away, what is the advantage of having a "custom" array like thing? It's hard to explain to people that Ember has a custom Array.  This is a different issue altogether, but I think it's another opportunity to cognitive load.

stefanpenner:
> Since ArrayControllers and ArrayProxies will go away, what is the advantage of having a "custom" array like thing? It's hard to explain to people that Ember has a custom Array. This is a different issue altogether, but I think it's another opportunity to reduce cognitive load.

I promise to write a detailed RFC on both the reasons for the current state, and a potential path forward.

abuiles:
For the interested, we just extended ember-watson to help you with this migration

```
npm install ember-watson@0.2.0 --save-dev
ember convert-prototype-extensions
```

stefanpenner:
> npm install ember-watson@0.2.0 --save-dev
> ember convert-prototype-extensions

rad!

stefanpenner:
@abuiles it should likely not use a top level command name like that, maybe something namespaced

j-carvalho:
Don't forget about the .on notation as well. Ember.on usage is also required. This plugin still doesn't cover this use case. 

Also, there are some places in the docs for observers where .observes('name').on('init') is specified, when we want the observer to be called on the initialization.

How should we handle these cases?

Ember.on('init', Ember.observer('name'... )) or Ember.observer('name, Ember.on('init', ...)) ?

stefanpenner:
> Also, there are some places in the docs for observers where .observes('name').on('init') is specified, when we want the observer to be called on the initialization.

we could make `Ember.observer(..., fn).on` work without prototype extensions

MiguelMadero:
> move away from prototype extensions.

Is there an idea of the timeframe for this? 

I think it makes sense to update the docs and change defaults so they're off by default would make sense, but completely removing them seems like unnecessary overhead for a lot of existing apps for 2.0. We could probably move that to an Add-on.

source/cookbook/working_with_objects/continuous_redrawing_of_views.md
```diff
@@ -27,19 +27,26 @@ of the interval. Since the `tick` method observes the incremented property
...
+  pulse: Ember.computed.oneWay('_seconds').readOnly(),
+  tick: on(observer('_seconds', function () {
+    var clock = this;
+    Ember.run.later(function () {
+      var seconds = clock.get('_seconds');
+      if (typeof seconds === 'number') {
+        clock.set('_seconds', seconds + (1/4));
+      }
+    }, 250);
+  }), 'init')
```
totaltrek: http://emberjs.com/api/classes/Ember.html#method_on
on (eventNames, func) 

isn't this backward? Shouldn't it be 
Ember.on('init, Ember.observer('_seconds', function(){...}));

eccegordo:
@stefanpenner what is the syntax for `volatile()`. 

Is it just tacked on at the end still?

This
```
status: function() {
  // return status
}.property('foo', 'bar').volatile()
```
becomes this

```
status: Ember.computed('foo', 'bar', function() {
  // return status
}).volatile()
```

stefanpenner:
> Is it just tacked on at the end still?

yes

mixonic:
*ping* would really love to have this merged.

tomdale:
I retract whatever rejections I may have had and am :+1: on migrating towards non-prototype-using examples, with the exception as discussed above about arrays. If `ember-watson` can do this transition for folks automatically I see no reason to push back.

bcardarella:
@tomdale aren't most of the array prototype extensions pulled out anyway now?

mixonic:
@bcardarella like `pushObject`? That and similar methods should still be present.

bcardarella:
@mixonic I'm referring to https://github.com/emberjs/ember.js/commit/62916e8f68b93f73b53e6ebd4254ff6560e6ae59

martndemus:
http://emberjs.jsbin.com/devesujowo/9/edit?html,js,console
I agree with @tomdale.

I also think the guide needs a section on Arrays, explaining that the vanilla `push`, `shift` and `pop` don't fly with Ember, that they need to use `pushObject`, `shiftObject` or `popObject` instead.

mixonic:
@bcardarella ah, I believe that PR just removes support for ES3 polyfills (IE8). We will continue to extend native array types with pushObject etc for the foreseeable future AFAIK.

source/concepts/what-is-ember-js.md
locks: and wouldn't `Ember.on('init', function() {` be favoured so the user doesn't have to call super?

mixonic: `Ember.on('init', function() {` has issues:

* The event system it relies up on is not performant.
* Using events obscures the order functions are called in during hook execution.
* It is not as ES2015-safe (native classes have a super, but no events).

Fwiw, Babel make it easy to use super anyway:

```js
  init() {
    this._super(...arguments);
    // some thing
  }
```

locks: Shouldn't that be `init(...arguments) {`? I thought `arguments` the keyword was "gone" from ES2015.
Thanks for the clarification :)

mixonic: I know nothing about `arguments` being removed. I would be shocked?!

martndemus: Its a best practice according to [AirBnB](https://github.com/airbnb/javascript).

`Never use arguments, opt to use rest syntax ... instead.`

But also: `Never name a parameter arguments. This will take precedence over the arguments object that is given to every function scope.`

mixonic: @martndemus Babel's transpiler does `this._super(...arguments)` just fine:

* [using super and ...arguments](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&playground=true&code=function%20bob()%20%7B%0A%20%20return%20this.fmt(...arguments)%3B%0A%7D)

The alternative introduces an array and a bunch of boilerplate:

* [using super and reset ..calledWithArgs](http://babeljs.io/repl/#?experimental=false&evaluate=true&loose=false&spec=false&playground=true&code=function%20bob(...bobArgs)%20%7B%0A%20%20return%20this.fmt(...bobArgs)%3B%0A%7D)

AirBnB's best practices are an interesting point to consider, but from a practical approach I disagree with the suggestion. super is an excellent place to use `...arguments`.

I would agree with their suggestion if you plan to manipulate arguments in any way though- then you should absolutely make an array instead of chaning the arguments object itself.

stefanpenner: most runtimes special case `method.apply(this, arguments)` to a more optimal form of passing `arguments` through without needing to fully materialize it.

thejameskyle: arguments is "deprecated" in ES2015, Babel handles this fine:

```js
function bob(...args) {
  return this.fmt(...args);
}
```
```js
function bob() {
  return this.fmt.apply(this, arguments);
}
```

More in depth explanation why I put that in quotes: https://esdiscuss.org/notes/2015-03-25

mixonic: the alternative I linked to above now shows the ideal behavior as @thejameskyle describes. Not sure if I read the output incorrectly or if there was a deploy of the repl.

**So,** using `...` at all times seems reasonable to me actually.

stefanpenner: @thejameskyle Ya, but its not going away in our life times.

stefanpenner: I don't think it is actually deprecated. I think those comments may have mislead you.

SaladFork:
>> Is it just tacked on at the end still?

> yes

So the proposed looks like:

With Extensions | Without Extensions
---------------------- | -------------------------
`function () { ... }.property('foo')` | `Ember.computed('foo', function () { ... })`
`function () { ... }.observes('foo')` | `Ember.observer('foo', function () { ... })`
`function () { ... }.on('bar')` | `Ember.on('bar', function () { ... })`¹
`function () { ... }.volatile()` | `(function () { ... }).volatile()`
`function () { ... }.readOnly()` | `(function () { ... }).readOnly()`

¹ Though the discussion suggests that `on('bar')` might also be maintained, at least if added to an 
`Ember.computed` or `Ember.observer`

It seems odd that we're moving from a consistent interface to one that can take one of two forms.

Of course, this mostly becomes a non-issue with ES7 decorators where `@computed` and `@readOnly` are similar.

pixelhandler:
One valuable note regarding the convention to move away from reliance on prototype extensions in the default documentation is to promote addon development. Addon development requires development to support *no* prototype extensions by default. In addition to the default documentation supporting the `Ember.computed` syntax over `.property()` it will be important do document when it's beneficial to use prototype extension e.g. with `Em.A()`. I for one really like a common way to write ember code in apps and addons  :)

stefanpenner:
> it will be important do document when it's beneficial to use prototype extension e.g. with Em.A().

will shortly go away entirely.

> One valuable note regarding the convention to move away from reliance on prototype extensions in the default documentation is to promote addon development.

yes

michaelrkn:
This PR has led to some great discussion, as well as some other PRs that have moved the state of the docs forward. However, the PR itself is out of date and mostly changes files that don't exist any more. I've just opened up #455 to summarize what needs to be done so that somebody can pick up the remaining work. If a couple of you who have been involved in this discussion can make sure that the issue correctly describes what needs to be done, that would be great! Thanks a bunch!

Issue #82, opened 2015-03-24
Update bound attribute syntax


michaelrkn:
I believe Ember 1.11 removes the need for `bind-attr`, so https://github.com/emberjs/guides/blob/master/source/templates/binding-element-attributes.md and https://github.com/emberjs/guides/blob/master/source/templates/binding-element-class-names.md should be removed. 

Since inline `if`s are much saner to use for the new attribute binding syntax - 

```
<div class="{{color}} {{if isEnabled 'active' 'disabled'}}"></div>
```

I think it would make sense to add this example to https://github.com/emberjs/guides/blob/master/source/templates/conditionals.md. 

The only thing I'm not sure about is the part on [Adding data attributes](https://github.com/emberjs/guides/blob/master/source/templates/binding-element-attributes.md#adding-data-attributes). Maybe this should be moved to the Views section somewhere?

kellysutton:
I'll take a crack at this. PR incoming.

michaelrkn:
Thanks for helping out with this, @kellysutton!

PR #111, opened 2015-03-27
Update bound attribute syntax

kellysutton:
Addresses https://github.com/emberjs/guides/issues/82


data/guides.yml
```diff
@@ -13,6 +13,12 @@
     - title: "Glossary"
       url: "glossary"
 
+- title: "Getting Ember"
```
trek: What's this entry for?

kellysutton: Whoops. I think that got pulled in from a different branch. I'll fix it.

source/templates/binding-element-attributes.md
```diff
@@ -6,11 +6,11 @@ to an image:
 
 ```handlebars
 <div id="logo">
-  <img {{bind-attr src=logoUrl}} alt="Logo">
+  <img src={{logoUrl}} alt="Logo">
 </div>
 ```
 
-This generates the following HTML:
+This generates the following HTML when the `logoUrl` property is set in the controller:
```
trek: :-1: on this change. That could equally come from the component.

source/templates/binding-element-attributes.md
```diff
@@ -18,25 +18,16 @@ This generates the following HTML:
...
 ```
 
-If you use `{{bind-attr}}` with a Boolean value, it will add or remove
-the specified attribute. For example, given this template:
+### Binding multiple values
 
-```handlebars
-<input type="checkbox" {{bind-attr disabled=isAdministrator}}>
-```
+Multiple values can also be bound to a single attribute:
```
trek: Active voice, please.

source/templates/binding-element-attributes.md
```diff
@@ -18,25 +18,16 @@ This generates the following HTML:
...
+<div class="{{priority}} {{placement}}"></div>
 ```
 
-If `isAdministrator` is `false`, Handlebars will produce the following:
-
-```html
-<input type="checkbox">
-```
+For further control over how these values are applied, see the [inline-if syntax
+documentation on the Conditionals page](/templates/conditionals/#toc_inline-if-syntax).
```
trek: absolute urls do not work with our publication style. They need to be relative.

source/templates/conditionals.md
```diff
@@ -35,3 +35,23 @@ To only render a block if a value is falsy, use `{{#unless}}`:
 to invoke a helper with a portion of your template. Block expressions look like
 normal expressions except that they contain a hash (#) before the helper name,
 and require a closing expression.
+
+## Inline-if syntax
+
+We can also use the inline-if to keep code terse:
```
trek: s/We/You. Subject changed from "you" used earlier.

michaelrkn:
Is `binding-element-attributes.md` even necessary now? It seems self-evident that you can include a Handlebars expression as an attribute.

kellysutton:
@michaelrkn I definitely think `binding-element-class-names.md` is no longer necessary. Given how important binding attribute values are, I'm for including the `binding-element-attributes.md` for newcomers.

kellysutton:
Also, correct me if I'm wrong: There's no way with HTMLBars to do "boolean" HTML5 attributes liked `disabled`, `checked`, etc.

michaelrkn:
What about just providing an example of using a Handlebars expression to provide the value of an attribute on the _Handlebars Basics_ page, right below "This would look up the firstName and lastName properties from the controller, insert them into the HTML described in the template, then put them into the DOM."?

Kuzirashi:
@kellysutton 
You can do in HTMLBars, Ember.js 1.11-beta5 this:
~~~~ HTML
<input type="submit" disabled={{model.isntValid}} value="..."></input>
~~~~

trek:
Going to merge because there's more 1.11 work that needs doing

PR #114, opened 2015-03-27
Proof of concept old version warnings

wagenet:
*DO NOT MERGE YET*

Uses JS to display warnings when viewing outdated versions of the guides. In this case, outdated means that it's a release version that is not the latest release. The presence of beta (and canary in the event that we decide to do it) will not trigger a warning.

One major downside to this approach is that the messages are only displayed if JS is enabled.

Currently requires a change in version.json structure so we'd have
to update all releases to support this. This is probably a good
idea anyway since we'd want the messages to display on old versions.

TODO:
* [ ] Style messages


trek:
Seems fine, but I'm beginning to think that we should move shared code like this to https://github.com/emberjs/guides.emberjs.com and access via javascript.

wagenet:
@trek works for me. Not sure if/when I'll have time to help with this, but I'm happy to fix up this PR if necessary to go with that.

locks:
@trek @wagenet what's the way forward here?

michaelrkn:
Closed in favor of #563.

PR #115, opened 2015-03-27
Testing Guide: Correct statement about what parameters are automatically provided to registered helpers.

kategengler:
Correct statement about what parameters are automatically provided to registered helpers.


rwjblue:
Thank you!

PR #119, opened 2015-03-27
Document how to find the end of the loading event.

wbyoung:
I found it difficult to discover how to actually be notified of when the end of a transition occurred after receiving the loading event. I believe this is the correct way, and if so, I think it'd be helpful for others to see this example.

This was discussed briefly on emberjs/website#1566.


locks:
lgtm. @michaelrkn ?c

michaelrkn:
looks fine but is beyond my expertise :)

locks:
@rwjblue merge plox

PR #129, opened 2015-03-28
Use pod structure throughout Query Params examples.

cball:
This uses pod structure throughout the example filenames. Most of them were already that way, but there were a few that were not.

I also fixed a typo specifying `controller:application` rather than `controller:articles`.


trek:
The rest of the guide are non-pods until that becomes the default, so the change so be in the other direction.

cball:
Alright. Re-opened with those changes: https://github.com/emberjs/guides/pull/140.

PR #132, opened 2015-03-29
correct url to action-bubbling.png

nonmanifold:
URL to action-bubbling.png should be relative, not absolute

### DIFF
--- a/source/templates/actions.md
+++ b/source/templates/actions.md
@@ -115,7 +115,7 @@ When an action is triggered, but no matching action handler is
 implemented on the controller, the current route, or any of the
 current route's ancestors, an error will be thrown.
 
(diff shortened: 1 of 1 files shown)


wagenet:
@rwjblue was this cherry-picked into the individual branches as well?

rwjblue:
Nope, it was only updated on master.

wagenet:
@rwjblue this seems like a regression to the 1.10 guides. I'd think we should fix regressions when they happen.

rwjblue:
I do not disagree, someone should PR this against v1.10 and v1.11 branches then.

wagenet:
@nonmanifold could you do that?

nonmanifold:
@wagenet I have no idea how to do it, sorry.

PR #133, opened 2015-03-29
Fix models/index.md

purezen:
The guides currently read ' [Ember Data][emberdata] ' in the introduction to models page. I have replaced it with link to the supposed url.


purezen:
:smiley:

PR #136, opened 2015-03-29
Fixes 404's on routing pages

Mongey:
The "routing guide", "ObjectController" & "ArrayController" links currently 404 on the [routing/generating-objects](http://guides.emberjs.com/v1.10.0/routing/generated-objects/) page.

"Specifying a Route's Model" link on [routing/setting-up-a-controller](http://guides.emberjs.com/v1.10.0/routing/setting-up-a-controller/) also 404's.


trek:
These changes will also result in 404s. Each build of the site is published into a directory, so we cannot use absolute urls. Can you make these relative urls?

Mongey:
@trek whoops. Amended commit :recycle:

PR #137, opened 2015-03-30
Swaps out Nokogiri implementation for semi custom encoder

danmcclain:
There is no way to insert a clean patch into CodeRay's encoder, so I had to copy/pasta some of the implementation to inject the changes


danmcclain:
@trek Let me know what you think, feels bad, but really no better way to do it :unamused:

trek:
Why can we not PR against CoderRay to just get this into their library?

rwjblue:
This is mostly to make the custom line highlighting work, and the internal coderay structure is not really setup to make this sort of extension easy.

IMHO, the modifications of the HTML output via Nokogiri was much cleaner and easier to deal with longer term.  I cannot reason about the internal coderay structure, but manipulating HTML with nokogiri is pretty simple.

danmcclain:
This would be a lot of effort to bring into CodeRay, and that assumes that they'd want to accept it at all. It's a lot of hackery to get to a place to do this work. If you look, I'm still splitting the code on new lines inside of CodeRay, because it is no clear spot to put the line wrapping code (there is a `begin_line` method that never gets called.

The nokogiri solution is cleaner in my opinion, and CodeRay is not easy to patch the way it is currently implemented

trek:
Closed in favor of #144

PR #139, opened 2015-03-30
How to create flash message alerts

ianwdunlop:
Cookbook recipe for creating rails  style flash messages in your ember app to alert users to things that are happening in your app.

### DIFF
--- a/source/cookbook/helpers_and_components/creating_flash_messages.md
+++ b/source/cookbook/helpers_and_components/creating_flash_messages.md
@@ -1,0 +1,132 @@
+### Problem
+You want Rails style flash messages to appear in your application and alert the user that something has happened.
+
+### Solution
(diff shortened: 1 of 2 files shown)


joostdevries:
Personally I'd rather have people using an addon for this, like https://github.com/aexmachina/ember-notify

trek:
Thanks for the submission. We've decided to move away from cookbook (they'll be gone for 2.0) now that the ecosystem of blog posts, books, and addons better supports this style of learning and reuse.

jbrown:
@ianwdunlop Just to throw another example out there for future reference, have a look at [this blog post](https://medium.com/delightful-ui-for-ember-apps/adding-flash-messages-to-an-ember-app-437b13e49c1b) and [addon](https://github.com/poteto/ember-cli-flash).

PR #144, opened 2015-03-30
Implements highlighting

danmcclain:
At the end of the filename or language add `{1,2,-4,+5}`,
`+` signifies that the line should be marked as added
`-` signifies that the line should be marked as removed
line numbers are otherwise highlighted

Does not currently handle ranges

Also updates colors in code blocks, fixing contrast

**Bringing it back**

@trek

### DIFF
--- a/lib/highlighter.rb
+++ b/lib/highlighter.rb
@@ -1,5 +1,6 @@
 require "redcarpet"
 require "coderay"
+require "nokogiri"
 
 module Highlighter
   class MissingLanguageError < StandardError; end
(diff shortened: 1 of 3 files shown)


danmcclain:
@trek Still interested in this version? Created new PR to bring it back

trek:
Still wondering about https://github.com/emberjs/guides/pull/137#issuecomment-87652562

trek:
Bummed CodeRay isn't better organized for this, but let's move on.

PR #148, opened 2015-03-30
Document Ember.Service

bf4:
stubbing something for discussion

### DIFF
--- a/source/understanding-ember/dependency-injection-and-service-lookup.md
+++ b/source/understanding-ember/dependency-injection-and-service-lookup.md
@@ -227,3 +227,31 @@ Injections can be made onto all of Ember's major framework classes including com
  -->
 
 Dependency injection and service lookup are two powerful tools in your Ember.js toolset, and every mature Ember application will require their use.
+
+### Dependency Injection with `Ember.Service`
+
+```JavaScript
+import Registry from "container/registry";
+import Service from "ember-runtime/system/service";
+import EmberObject from "ember-runtime/system/object";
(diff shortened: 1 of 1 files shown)


bf4:
@mitchlloyd @rwjblue any feedback on what should go in the guides here appreciated.  I haven't actually used a service before.

mitchlloyd:
@bf4 Reading over the current document, it seems like there needs to be a pretty big overhaul. Probably the documentation needs to go something like this:

* What is dependency injection?
  - Example of route getting the router injected should be updated for routing service, but that depends on https://github.com/emberjs/ember.js/pull/10596
* What is a service and what is it good for? (use cases, audio example in here is good but needs to be updated to use a service and not use `render`)
* Using Ember.inject to inject services (and controllers I suppose)
* Ember's register & inject (what is there appears valid)

bf4:
@mitchlloyd If you're up to writing it, I can close this. I really just wanted to start a discussion and maybe learn a thing or two.

I don't think I have enough context or experience to expand on your bullet points in any useful way. :smile: If you want  to collaborate in code, I suppose you can open a PR into my branch.

mitchlloyd:
I think there just needs to be an issue to document services if there isn't
one already. I have had many people ask me "where are services documented"?

bf4:
I decided to make the issue as a PR in case that it was close to what I wrote that I could take it home.  (I also prefer PRs to 'hey could you do this?' :)

locks:
@bf4 can you close this? like @mitchlloyd commented it's not a light undertaking and it's easier to coordinate internally

PR #157, opened 2015-04-01
Correct falsy --> false

viaro-vinicio:


trek:
"falsy" is accurate. The statement above explains why it's not just "false"

PR #158, opened 2015-04-01
Testing component blocks

knownasilya:
Basically taken from https://gist.github.com/code0100fun/f5922b6f44eee2abc21d 

Thanks @code0100fun and https://emberweekend.firebaseapp.com/episodes/the-weekend-strikes-back

### DIFF
--- a/source/testing/testing-components.md
+++ b/source/testing/testing-components.md
@@ -201,6 +201,41 @@ test('trigger external action when button is clicked', function(assert) {
 });
 ```
 
+### Testing component blocks and defaults
+
+Often you might define a component which works in block form:
+
+```app/templates/components/my-component.hbs
+{{#if template}}
+  {{yield}}
+{{else}}
+  {{defaultContent}}
+{{/if}}
+```
+
+Now this shows `{{defaultContent}}` when using `{{my-component}}` otherwise it shows the content within `{{#my-component}}custom content{{/my-component}}`.
+
+To test this functionality, we can do the following:
+
+```tests/unit/components/my-component-test.js
+test('component with optional block', function(assert) {
+  assert.expect(1);
+ 
+  var component = this.subject();
+  
+  Ember.run(function() {
+    component.set('template', function() {
+      return 'Something else';
+    });
+    component.set('defaultContent', 'foo');
+  });
+  
+  assert.trimEq(this.$().text(), 'Something else');
+});
+```
+
+Notice that `template` is actually a function that returns a string.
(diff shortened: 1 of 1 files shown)


source/testing/testing-components.md
```diff
@@ -201,6 +201,41 @@ test('trigger external action when button is clicked', function(assert) {
 });
 ```
 
+### Testing component blocks and defaults
+
+Often you might define a component which works in block form:
+
+```app/templates/components/my-component.hbs
+{{#if template}}
```
rwjblue: We intend to add `hasBlock?` helper for this, I am unsure that we should suggest this (though checking `template` will likely always work).

trek:
Going to close from lack of reply/activity. Feel free to re-open if we can address @rwjblue's question.

knownasilya:
What was the question? lol. Looks like we are waiting on `hasBlock?` to be implemented? Seems silly to wait on that and not have any documentation in the mean time. We can always update the docs once that has landed.

rwjblue:
Accessing `this.get('template')` triggers a deprecation in 1.13, and `hasBlock` is enabled in 1.13.

rwjblue:
Also, setting `template` does nothing in 1.13.

knownasilya:
@rwjblue so just replace template with hasBlock?

knownasilya:
What's the alternative to setting `template`?

rwjblue:
> so just replace template with hasBlock?

Yes, without the question mark:

```hbs
{{#if hasBlock}}
  Stuff
{{/if}}
```

knownasilya:
@rwjblue what about setting the template? What's the new API for that?

rwjblue:
There isn't one.  The template for a component is supplied by the block provided by the calling location.

Please explain the intent of changing the template of a component, maybe I have a better answer/explaination...

knownasilya:
For testing integration between different components, e.g.

```hbs
{{#video-player as |playing|}}
  {{ad-overlay ad=myAd playing=playing}}
{{/video-player}}
```

Or even static data, since you want to test if your component is behaving correctly when in block mode.

rwjblue:
For testing, you should be using the newer style component integration tests.  We need more documentation for it though.

Added in https://github.com/switchfly/ember-test-helpers/pull/38, you can do something like this [example from liquid-fire](https://github.com/ef4/liquid-fire/blob/new-test-helpers/tests/integration/helpers/liquid-if-test.js).

PR #162, opened 2015-04-01
Fix code block file paths

HeroicEric:
Fixes a few incorrect file paths for code blocks


trek:
:-1: shorthand link-to. We avoid using it elsewhere.

HeroicEric:
Interesting. I took that part out.

Why against the shorthand though?

trek:
Neither for nor against, but we prefer to keep samples as similar as possible, and block form is the lowest common denominator for `link-to`, so we use it preferentially

PR #166, opened 2015-04-02
Remove reference to emq.globalize() in v1.10.0

ryanmurakami:


trek:
We probably won't republish 1.10 for a while however (if ever)

PR #167, opened 2015-04-02
Remove reference to emq.globalize() in v1.11.0

ryanmurakami:


trek:
I didn't see a matching request for master, which is the more important change to get in.

PR #175, opened 2015-04-03
Remove global `App` variable

swashcap:
The `App` variable, previously used as the Ember application's global namespace, is no longer used in the guides in favor of ES6 modules. The `App` would be undefined in this case.


rwjblue:
Hmm, looks like this PR is pulling in all commits from master to v1.11.

We just need the specific change from  fe1037c.

swashcap:
> Hmm, looks like this PR is pulling in all commits from master to v1.11.

Ah, shoot. How do I pick just that one commit? `cherry-pick` locally and do another PR?

rwjblue:
Yeah, I think that is easiest.  You could also likely use the github online file editor and just make the change there too.

swashcap:
See #176.

PR #176, opened 2015-04-03
Remove global `App` variable

swashcap:
The `App` variable, previously used as the Ember application's global namespace, is no longer used in the guides in favor of ES6 modules. The `App` would be undefined in this case.

(See emberjs/guides@fe1037c12ad5fd8b25394147a1257d2af470c49d for duplicate on `master`.)


rwjblue:
Perfect!

PR #177, opened 2015-04-03
Fix link to "specifying a routes model"

joostdevries:


joostdevries:
If ok I'll also open to 1.10 and 1.11.

rwjblue:
@joostdevries - Yes, please.

PR #180, opened 2015-04-03
Add missing `DS.` to model example.

jackfranklin:
Just spotted this when reading the docs - I think this change is the right one but I'm not too familiar (hence reading the tutorial!) so please do correct me if I'm wrong!


rwjblue:
Thanks!

Issue #117, opened 2015-03-27
Version dropdown is empty if JS is disabled


wagenet:
In theory most of the guides functionality should work without JS, however, the version selector just shows up blank. If JS isn't available, we should at least not display it.

PR #187, opened 2015-04-03
Display version links for users without JS

HeroicEric:
- Displays version links for users without JS enabled
- Hides version select for users without JS enabled

without js

I also moved `source/versions.json` to `data/versions.yml` to make this work. I'm not sure if anything else is depending on that being there but it didn't look like it.


HeroicEric:
Fixes #117

source/layout.erb
```diff
@@ -62,9 +62,23 @@
 
     <div id="content-wrapper">
       <div id="sidebar-wrapper">
-        <select class="version-select" style="width: 100%"></select>
-        <div id="sidebar">
+        <div id="versions">
+          <noscript>
+            <ul>
+              <% data.versions.each do |version| %>
```
trek: How would previously published revisions of the site know about the new set of versions?

source/javascripts/app/guides/version-select.js.erb
```diff
@@ -1,16 +1,12 @@
 var currentRevision = '<%= data.search.revision %>';
```
trek: This will work only in development, sadly, and would not work on publication.

trek:
Thanks for contributing. This PR goes a bit further than the issued reported in #117 and cannot be merged as-is.
 
> I'm not sure if anything else is depending on that being there but it didn't look like it.

The JSON file included in this repo is only for testing purposes. The actual file used is https://github.com/emberjs/guides.emberjs.com/blob/master/snapshots/versions.json and available to all published revisions. 

We just don't support this feature for users with JavaScript enabled.

HeroicEric:
Thanks for the info. 




I'll update to simply hide the dropdown as you mention in the issue.



—————————————————————


Eric Kelly

Lead Developer at Greenfield

@HeroicEric all over the internet

HeroicEric:
Closing in favor of #192

PR #192, opened 2015-04-04
Hide version select when JS disabled

### LINKED ISSUE #117: Version dropdown is empty if JS is disabled

HeroicEric:
Fixes #117 

Hides the version select when javascript is not enabled

### DIFF
--- a/source/stylesheets/site.css.scss
+++ b/source/stylesheets/site.css.scss
@@ -298,6 +298,10 @@ p {
   }
 }
 
+.version-select {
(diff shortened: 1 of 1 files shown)


trek:
Can you add a comment in the CSS about how the select get's shown? This seems like a refactoring hazard as the `display` value gets reset elsewhere but you won't know that looking here.

PR #194, opened 2015-04-04
Updated reference to middleman-swiftype

bsclifton:
Updated reference to middleman-swiftype; it now generates the search.json on build. The new version of middleman-swiftype has rspec tests around all of the functionality we use. I updated README.md to show new usage (simply using `middleman build`).


trek:
Awesome! Thanks for all the hard work on this.

PR #197, opened 2015-04-04
Remove references to shorthand Em

HeroicEric:
Replaces references to `Em.*` with `Ember.*`.

I think seeing these would be confusing to new people coming in.


rwjblue:
:+1:

Issue #196, opened 2015-04-04
Block params version of each


HeroicEric:
[Templates: Displaying a List of Items](http://guides.emberjs.com/v1.11.0/templates/displaying-a-list-of-items/) doesn't mention anything about using block params with the `each` helper.

Should the current examples be updated to use block params or should block params be shown in addition to the `each .. in ..` version?

rwjblue:
Each...in is deprecated (or will be very soon) in 1.12.0-beta.2 (along with with..as). We should definitely make sure we are not using the deprecated syntax in these guides.

HeroicEric:
Closing this since those ^^ were merged

PR #198, opened 2015-04-04
Use block params version of each

HeroicEric:
related to https://github.com/emberjs/guides/issues/196

This replaces references to `{{#each ... in ...}}` with `{{#each ... as |...|}}`.


rwjblue:
You can use `itemController` with `{{each}}`: http://emberjs.jsbin.com/rwjblue/383/edit?html,js,output

HeroicEric:
:dizzy_face: I only tried putting it at the end. I'll get those too :dash:

HeroicEric:
I think I got them all now.

rwjblue:
LGTM

PR #202, opened 2015-04-05
Update templateContext to context

anthonycollini:
`templateContext` was renamed to `context` a long time ago, and there is no mention of `typeOfTemplateContext` in the repo. This change updates the doc to reflect current master.


trek:
cc @mmun or someone for accuracy. I have no idea.

mmun:
This should just document the `get(<path>)` helper. We should remove references to `view` and `context` in the Ember code as well.

michaelrkn:
@locks are you familiar with this `get(<path>)` helper? If so, maybe you can put together an alternate PR so we can get this content up to date.

michaelrkn:
Oh, I just came across http://emberjs.com/api/classes/Ember.Handlebars.helpers.html#method_debugger, which has a really nice explanation and seems up-to-date. I've opened #417 to pull that info straight into the guides here.

@rwjblue can you close this and take a quick look at #417?

PR #203, opened 2015-04-05
Describe dashes in helper names

anthonycollini:
Example is incorrect, mentioning makeBoundHelper without a dasherized name and making the user assume that they can use it immediately in their template.


rwjblue:
FYI - I reverted in https://github.com/emberjs/guides/pull/352, and listed a few resources.

PR #209, opened 2015-04-05
Fix location value in last example

EPadronU:


rwjblue:
Looks like this is duplicate of https://github.com/emberjs/guides/pull/208, which was just landed.

rwjblue:
Thanks for taking the time to submit the patch though!

EPadronU:
I'm so ashamed. Thank you for your time :)

2015-04-05 18:30 GMT-04:30 Robert Jackson <notifications@github.com>:

PR #210, opened 2015-04-06
Fix link to EC6 support page

Manc:
Typo in link currently leads to a 404 page


data/guides.yml
```diff
@@ -376,8 +376,8 @@
       url: "the-view-layer"
     - title: "Managing Asynchrony"
       url: "managing-asynchrony"
-    - title: "Keeping Templates Up-to-Date"
-      url: "keeping-templates-up-to-date"
+    # - title: "Keeping Templates Up-to-Date"
+    #   url: "keeping-templates-up-to-date"
```
trek: ?

Manc: Sorry about putting 2 different things into one pull request. I can separate it if you like...

I commented the lines above out, because there is no page "Keeping Templates Up-to-Date" and clicking on the link in the menu of the documentation leads to a 404 page, too.

trek:
Then just remove the entries? I believe they're already gone on master.

PR #215, opened 2015-04-08
Glossary: some edits to CDN entry

jamesarosen:
Previously, the entry said that the app developer uploaded assets to the CDN. There are some CDNs that act that way, but more act as caching proxies. In that model, the app developer sets up the CDN as a proxy to their origin and uses the CDN URLs in their app.


hjdivad:
LGTM.  More accurate, but still just as clear.

PR #225, opened 2015-04-12
Describe how to run your tests

roryokane:
Not currently described in any of the Testing guides. New users will probably guess the existence of `ember test`, but they will miss out on `ember test --server` and visiting `/tests` in the browser without being told about them.

### DIFF
--- a/source/testing/index.md
+++ b/source/testing/index.md
@@ -37,7 +37,7 @@ supported through third-party addons.
 
 ### How to Run Your Tests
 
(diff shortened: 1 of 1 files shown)


rwjblue:
We should not suggest hitting /tests/ while running 'ember server' (the assets are built in the development environment which means they are potentially out of sync with 'ember test').

:+1: to suggesting 'ember test' though.

roryokane:
How’s this?

PR #239, opened 2015-04-18
Trigger search on enter. Fixes #125

### LINKED ISSUE #125: `enter` key on search doesn't trigger a search for the current term
Entering a search term and pressing enter seems to simply add an empty query parameter to the URL (e.g. `http://guides.emberjs.com/v1.11.0/?`) and refresh the page.

If an item is chosen from the dropdown, it works as expected and transitions to the correct page.

jpadilla:
Uses search behavior directly from swiftype like done in [emberjs/website](https://github.com/emberjs/website/blob/master/source/javascripts/app/swiftype.js) while keeping autocomplete feature from the jQuery plugin.

Only caveat with this approach is that I don't see how we can customize overlay rendering URLs to include the `revision`.

Also, mentioned to @trek the idea of migrating to the [new configurable search widget](https://swiftype.com/documentation/tutorials/upgrade_to_configurable_search_widget). Unfortunately couldn't give this a try without an install key(not the engine key) and I didn't manage to get any crawling done on a sample app. I'd be down on getting that done if anyone cares on setting up the configuration and sending over an install key.

### DIFF
--- a/source/javascripts/swiftype.js.erb
+++ b/source/javascripts/swiftype.js.erb
@@ -1,5 +1,25 @@
-$(function() {
+var Swiftype = window.Swiftype || {};
+(function() {
+  Swiftype.key = '<%= data.search.engine_key %>';
+  Swiftype.inputElement = '#st-search-input';
+  Swiftype.resultContainingElement = '#st-results-container';
+  Swiftype.attachElement = '#st-search-input';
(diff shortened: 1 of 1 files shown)


trek:
Can you open a new Issue or the configurable widget and flesh that out into a [Quest Issue](https://medium.com/@trek/source-quest-ff7d227d8fed)

Issue #235, opened 2015-04-14
Actions.md broken links


Kilowhisky:
Didn't know quite the right place to report this.

This is giving a 404:
http://guides.emberjs.com/images/template-guide/action-bubbling.png

This link is dead:
http://guides.emberjs.com/guides/understanding-ember/the-view-layer/#toc_adding-new-events

IanVS:
That same dead link is also on http://guides.emberjs.com/v1.11.0/templates/actions/#toc_specifying-the-type-of-event

nicolechung:
I'd like to fix this, but I'm a little confused about where (i.e. tag or master) to fix these.

trek:
Master.

nicolechung:
Ok will do. Give me the weekend.

jpadilla:
Those dead links should probably link to http://emberjs.com/api/classes/Ember.View.html#toc_event-names now

mostwanted1976:
still broken:

http://guides.emberjs.com/v1.10.0/templates/actions/ > "listed in the View guide"

prikhi:
This was fixed & can be closed? Just needs #325 so that it'll show up.

rwjblue:
Thanks for confirming!

PR #240, opened 2015-04-19
link and image fix

nicolechung:
Not sure why the image is broken on the actions page, it works when I build it and the link is the same as the live site.

I notice sometimes people use  tags instead of Markdown image text so I tried that.

### DIFF
--- a/source/templates/actions.md
+++ b/source/templates/actions.md
@@ -118,7 +118,7 @@ When an action is triggered, but no matching action handler is
 implemented on the controller, the current route, or any of the
 current route's ancestors, an error will be thrown.
 
-![Action Bubbling](../images/template-guide/action-bubbling.png)
+![Action Bubbling](../../images/template-guide/action-bubbling.png)
 
 This allows you to create a button that has different behavior based on
 where you are in the application. For example, you might want to have a
@@ -168,7 +168,7 @@ You can specify an alternative event by using the `on` option.
 You should use the normalized event names [listed in the View guide][1].
 In general, two-word event names (like `keypress`) become `keyPress`.
 
-[1]: ../../understanding-ember/the-view-layer/#toc_adding-new-events
+[1]: http://emberjs.com/api/classes/Ember.View.html#toc_event-names
 
 ### Specifying Whitelisted Modifier Keys
 


source/templates/actions.md
```diff
@@ -117,7 +117,7 @@ When an action is triggered, but no matching action handler is
 implemented on the controller, the current route, or any of the
 current route's ancestors, an error will be thrown.
 
-![Action Bubbling](/images/template-guide/action-bubbling.png)
+<img src="/images/template-guide/action-bubbling.png" alt="Action Bubbling">
```
trek: This needs to be relative or it will break when published image lives at `v1.12.0/images/template-guide/action-bubbling.png` and not `/images/template-guide/action-bubbling.png`

source/templates/actions.md
```diff
@@ -167,7 +167,7 @@ You can specify an alternative event by using the `on` option.
 You should use the normalized event names [listed in the View guide][1].
 In general, two-word event names (like `keypress`) become `keyPress`.
 
-[1]: /guides/understanding-ember/the-view-layer/#toc_adding-new-events
+[1]: http://emberjs.com/api/classes/Ember.View.html#toc_event-names
```
trek: why did this change from guides to API?

nicolechung: Link change was suggested in the issue? https://github.com/emberjs/guides/issues/235

source/templates/actions.md
nicolechung: You mean, back the markdown? (Both versions use relative links, both img tag and markdown version).

Not sure why that link is not working on the actual live site because it is not broken when I build it with:

-Action Bubbling

source/templates/actions.md
nicolechung: In the View Guide, the only documentation I see about handling events is this:
http://guides.emberjs.com/v1.11.0/views/handling-events/

Which links to this:
http://emberjs.com/api/classes/Ember.View.html#toc_event-names

```
You should use the normalized event names listed in the View guide. In general, two-word event names (like keypress) become keyPress.
```

trek:
I'm pretty sure that image tag will not work deploy.

nicolechung:
Changed it back?

trek:
The url that page appears as would be `templates/actions/`, which is implicitly interpreted as `templates/actions/index.html`. I suspect the image is broken because it needs to go back two directories (`actions`, and `templates`) before going forward to `images`

nicolechung:
I really can't tell what is going on.

For example, in models/index.md (same level of nesting) this works:
```
![Diagram showing process for finding an unloaded record after the payload has returned from the server](../images/guides/models/finding-loaded-record-diagram.png)
```

However in actions.md, writing this (file has the same level of nesting, i.e. models/index.md vs templates/actions.md)

```
![Action Bubbling](../images/template-guide/action-bubbling.png)
```

breaks the image link.

trek:
That's because of the "pretty urls" feature of middleman

`/models/index.md` resolves to `/models/index.html` (depth of two)

`templates/actions.md` resolves to `templates/actions/index.html` (depth of three).

It's very confusing.

nicolechung:
Oh got it. Let me try something else

nicolechung:
Seems to work on my local now with ../../images...

trek:
Cool, can you squash this into one commit?

PR #252, opened 2015-04-22
Made "Routes" header plural, for consistency

eostrom:
Consistent with "Templates," "Components," and "Models."

I left "Router" alone, because there's only one.


eostrom:
Actually, I have several suggested changes for this page:

1. Consistent pluralization in headers.
2. Reorder the terms so that it's easier to follow the connections between them (e.g., "routes" next to "router").
3. Introduce "controllers" here so that it's less mysterious on the following page (naming conventions).

Do these sound reasonable? If so, would you prefer separate pull requests, one pull request with multiple commits, or just one commit that does all three?

trek:
One PR is fine.

PR #254, opened 2015-04-23
Move "handlebars basics" before "application template"

michaelrkn:


rwjblue:
Makes sense to me, :+1:

Issue #249, opened 2015-04-21
Fail tests when data/guides.yml references guides that don't exist


trek:
We recently removed a guide but neglected to remove its entry from the `data/guides.yml` file, which lead to broken links.

Interestingly https://github.com/emberjs/guides/blob/7ba7f4bfdab0bd4bf7176fbbbc1fdefeeba40656/lib/toc.rb#L48 would still generate a url, but without the version slug included.

We should detect attempts to link to missing guides or chapters and fail the build

PR #266, opened 2015-04-26
Raise exception when data/guides.yml references guides that don't exist

michaelrkn:
Addresses #249.

Feedback welcome!

Notes:

* Not sure if this is the approach @trek had in mind but seemed the simplest way to go about it. If you'd prefer a different approach, just let me know what you're thinking.
* Not sure I covered all cases. Somebody who knows this code better than me should take a look :)
* `allow(File).to receive(:exist?).and_return(true)` feels a bit blunt.
* In `toc.rb`, `middleman_url` variable has a couple variations, is declared and then overwritten, and seems redundant with `url` on line 45. This seems ripe for refactoring... if I'm not missing something, I'll open an issue and take a stab at cleaning it up. I'm hesitating because it appends `.html` to the end but I don't actually see the `.html` anywhere in the app, so I'm worried I'm missing something.

### DIFF
--- a/lib/toc.rb
+++ b/lib/toc.rb
@@ -31,7 +31,11 @@ module TOC
           requested_guide_url = slugs[0]
           current = (guide.url == requested_guide_url)
 
-          middleman_url = "/#{guide.url}/#{guide.chapters[0].url}.html"
+          middleman_base_url = "/#{guide.url}/#{guide.chapters[0].url}"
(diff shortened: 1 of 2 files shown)


trek:
I believe `link_to` wants the `.html` ending or it doesn't generate the link correctly.

trek:
And thanks! I've closed the related issue.

PR #267, opened 2015-04-26
Fixed link to 'Specifying a route's model'

cosmix:


source/models/index.md
```diff
@@ -45,7 +45,7 @@ $ npm rm ember-data --save-dev
 
 Learning to use Ember Data is easiest once you understand some of the
 concepts that underpin its design.
-
+sp
```
trek: ?

cosmix:
No idea how that crept in there. Apologies, the patch should only include the corrected link. Will revise.

trek:
I also think the `../..` is required when linking relatively between two sections. Middleman's "clean urls" means that the actual link is to `routing/specifying-a-routes-model/` which is implicitly `routing/specifying-a-routes-model/index.html`, which is why you need to move back two relative segments.

cosmix:
I see. However, it currently breaks the guide on guides.emberjs.com. See http://guides.emberjs.com/v1.11.0/models/

I made another Pull Request. If you feel this is a mistake, please do reject/close.

trek:
Ah, yes, when linking from and index to another section, you only need on `../`

PR #271, opened 2015-04-28
Updated fixture adapter docs links

kamilogorek:
Hey,
I'm not sure about the reason, but all thinks in form of [name][linkno] are not working properly, and they're pointing to eg. http://guides.emberjs.com/guides/models/the-rest-adapter instead of http://guides.emberjs.com/v1.11.0/models/the-rest-adapter/

Let me know if this PR should be pointed at different branch.

Cheers,
Kamil

### DIFF
--- a/source/models/the-fixture-adapter.md
+++ b/source/models/the-fixture-adapter.md
@@ -22,7 +22,7 @@ export default DS.FixtureAdapter.extend();
 
 #### Define Your Model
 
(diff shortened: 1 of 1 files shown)


trek:
I think this is because the relative links are going back one directory too far from their current location.

bf4:
Rebase or close?

michaelrkn:
This file no longer exists, and the PR should be closed. cc @rwjblue

PR #278, opened 2015-05-02
Refactor TOC tests

michaelrkn:
These tests had a ton of duplication, especially of the YAML data, that made them very difficult to read and work with. This PR standardizes all the tests to use the same YAML data, and cleans up other problems and duplication I found as well.


rwjblue:
:+1: - Thanks for putting the time into this!

michaelrkn:
My pleasure! :)

PR #283, opened 2015-05-05
noticeable phrasing

da-vaibhav:
Change in phrasing. 
Earlier sentencing got me confused while skimming through. This one is better and more noticeable while reading, IMO.


trek:
This won't merge cleanly anymore. Can you resubmit? Thanks

PR #284, opened 2015-05-06
computer properties is not yet introduced

baijum:
A link to chapter on computer properties would be good as this topic is not yet introduced until this point.


trek:
This link is absolute and will break when we publish a version of the site. Can you make this relative and resubmit?

PR #290, opened 2015-05-07
path to calendar helper

baijum:


trek:
Can you rebase? This won't merge cleanly now.

baijum:
@trek It looks like the section has been removed (Custom View Helpers), so this patch is not relevant.

PR #291, opened 2015-05-07
Update link that went to 404

MrJadaml:
The link to specifying a routes model was broken.


trek:
This is linking to a specific version of the guides and would 404 on all other versions. Can you make this relative and resubmit?

PR #292, opened 2015-05-07
Came across another 404ed link

MrJadaml:


trek:
This links to a specific version of the guides, which would when viewing the guide on other versions. Can you make this relative and resubmit?

PR #299, opened 2015-05-12
Fix links in guides about testing and Ember Inspector

XrXr:
Previously, the links point to pages that do not exist. Using relative path in
the URL fixes the issue.


teddyzeenny:
:+1:

PR #300, opened 2015-05-13
Update action handler guide for preventDefault parameter

anobin:


trek:
Looks like this needs a quick rebase.

locks:
@anobin poke

michaelrkn:
Not trying to take credit for others' work, but since there hasn't been a reply from @anobin, I've just opened #416 to move his work onto an up-to-date commit. If this looks good, we can merge #416 and close this. cc @rwjblue

PR #302, opened 2015-05-13
several dead-link fix and ES6 -> ES5 syntax

wismer:
I found a few dead links in the v1.11.0 guide and a syntax inconsistency regarding the two leading ECMA versions (5 & 6).

### DIFF
--- a/source/models/the-fixture-adapter.md
+++ b/source/models/the-fixture-adapter.md
@@ -79,6 +79,6 @@ a uniquely identifiable field. By default, Ember Data assumes this key
 is called `id`. Should you not provide an `id` field in your fixtures, or
 not override the primary key, the Fixture Adapter will throw an error.
 
-[1]: /v1.11.0/models/defining-models
-[2]: /v1.11.0/models/finding-records
-[3]: /v1.11.0/models/the-rest-adapter
+[1]: ../defining-models
+[2]: ../finding-records
+[3]: ../the-rest-adapter
--- a/source/models/the-fixture-adapter.md
+++ b/source/models/the-fixture-adapter.md
@@ -40,7 +40,7 @@ In order to attach fixtures to your model, you have to use `reopenClass` method 
 fixtures:
 
 ```app/models/documenter.js
-let Documenter = DS.Model.extend({
+var Documenter = DS.Model.extend({
   firstName: DS.attr( 'string' ),
   lastName: DS.attr( 'string' )
 });
@@ -52,7 +52,7 @@ Documenter.reopenClass({
   ]
 });
 
-export default Documenter
+export default Documenter;
 ```
 
 That's it! You can now use all of methods for [Finding Records][2] in your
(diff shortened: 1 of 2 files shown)


source/models/the-fixture-adapter.md
```diff
@@ -40,7 +40,7 @@ In order to attach fixtures to your model, you have to use `reopenClass` method
 fixtures:
 
 ```app/models/documenter.js
-let Documenter = DS.Model.extend({
```
wismer: Only instance of ES6 syntax in the guides. I think it could become a little confusing to those that do not know about ES6.

source/models/the-fixture-adapter.md
```diff
@@ -79,6 +79,6 @@ a uniquely identifiable field. By default, Ember Data assumes this key
 is called `id`. Should you not provide an `id` field in your fixtures, or
 not override the primary key, the Fixture Adapter will throw an error.
 
-[1]: /guides/models/defining-models
```
wismer: I'm unsure about these; I don't know if there is a root-template for these links that automatically matches the selected version, but those broken links are live.

rwjblue: They need to be relative paths.

rwjblue: I believe they should be `../defining-models`, but you'll have to test to confirm...

source/templates/actions.md
```diff
@@ -115,7 +115,7 @@ When an action is triggered, but no matching action handler is
 implemented on the controller, the current route, or any of the
 current route's ancestors, an error will be thrown.
 
-![Action Bubbling](/images/template-guide/action-bubbling.png)
```
rwjblue: Should be relative

wismer: fixed to relative

wismer:
@rwjblue added the relative paths. Works when I run it locally.

source/templates/actions.md
```diff
@@ -165,7 +165,7 @@ You can specify an alternative event by using the `on` option.
 You should use the normalized event names [listed in the View guide][1].
 In general, two-word event names (like `keypress`) become `keyPress`.
 
-[1]: /guides/understanding-ember/the-view-layer/#toc_adding-new-events
+[1]: /understanding-ember/the-view-layer/#toc_adding-new-events
```
rwjblue: this should also be relative

source/templates/actions.md
```diff
@@ -115,7 +115,7 @@ When an action is triggered, but no matching action handler is
 implemented on the controller, the current route, or any of the
 current route's ancestors, an error will be thrown.
 
-![Action Bubbling](/images/template-guide/action-bubbling.png)
+![Action Bubbling](/../images/template-guide/action-bubbling.png)
```
rwjblue: Need to be relative

wismer:
Apologies for the false starts. I hope I got it this time.

rwjblue:
No worries!

rwjblue:
Thanks for sticking through it, we definitely appreciate the help!

Issue #206, opened 2015-04-05
[Proposal] When you change ember versions try to stay on the same section of the guide


varblob:
Currently when you change the version of the ember guide it resets you to the root of the guide, which is not ideal.  

I have a couple proposals for how to try and approach this with the full expectation that my understanding of how things are setup maybe totally wrong.

# The simple fallback approach

When a user switches version attempt to match based on the title of the guide, if this fails match based on the section, if this fails goto root.  

It may also make sense to add some sort of non title identifier to guide pages so small title changes don't cause fallback.

# The manual mapping approach

In the data.yml if a guide goes away or changes in some significant way then put a manual mapping of the closest page.  I would imagine that there should always be some sort of mapping possible since during the deprecation of the old feature there is normally discussion of how to upgrade / update.  If there is no one to one mapping for a page then it would seem an excellent opportunity to put an upgrade guide page that explains to users the how to upgrade or simply points them to more than one guide if there is a one to many mapping.  

# Other side benefits

If the mapping issue is solved, especially if it's intermixed with pages helping with upgrading, a link to the new version can be placed more prominently on out of date pages encouraging users to upgrade.

trek:
Hey @varblob we'd need to see concrete details on how something like this can be implemented to demonstrate the idea has been thought through at a technical level. It's a good idea, but needs _actionable_ steps for people help out or :+1: the work.

There are _many_ technical questions that would need identifying and answering before we could do this.

e.g.
>When a user switches version attempt to match based on the title of the guide

Match where? Where is this data stored and how? How is it looked up? 

> In the data.yml if a guide goes away or changes in some significant way then put a manual mapping of the closest page.

We remove occurs in `v1.11.0/foos`  and map to the closest page `v1.12.0/new-foos`, which seems possibly easy? What happens when we've release five more revisions. Where is the data stored that the system knows  `v1.11.0/foos` now points to `v1.17.0/new-foos`. What if there is another change between the two? `v1.11.0/foos` -> `v1.13.0/new-foos` -> `v1.15.0/newer-and-imrproved-foos`. How does ` `v1.11.0/foos` know to point to `v1.15.0/newer-and-imrproved-foos`?

Where do we store data so it can mutate over time and remain manageable?

varblob:
I totally understand the need for more detail I'm hoping this is the place to try and get there.

I will poke further at this in the evening.  Do you think you could point me towards a good entry point to try and understand better what is being done at the moment especially around how the mds are served up?

[...]

PR #306, opened 2015-05-17
Stay on same page when switching versions

michaelrkn:
Also updated 404 to be more helpful in case page doesn't exist in the version switched to.

To test the JS code, I went to http://guides.emberjs.com/v1.11.0/concepts/core-concepts/, ran the relevant code:

```js
var versionSelectBox = $(".version-select");
versionSelectBox.on('change', function() {
  var version = this.value;
  window.location = window.location.href.replace(currentRevision, version);
});
```

and then manually switched from v11 to v10.

### DIFF
--- a/source/404.html.erb
+++ b/source/404.html.erb
@@ -5,7 +5,9 @@ title: 404 - Not Found
 <h1>404</h1>
 
 <p>
-  This page wasn't found. If you were looking for documentation, please try
-  the <a href="/guides">Guides</a> section of the site. If you expected
-  something else to be here, please <a href="https://github.com/emberjs/website/issues">file a ticket</a>.
+  This page doesn't exist in this version of the Guides.
+  Try hitting the back button to go back to where you were,
+  or go to the <a href="/">Guides homepage</a>.
+  If you expected something else to be here, please
+  <a href="https://github.com/emberjs/website/issues">file a ticket</a>.
 </p>
--- a/source/javascripts/app/guides/version-select.js.erb
+++ b/source/javascripts/app/guides/version-select.js.erb
@@ -5,7 +5,7 @@ $.ajax('/versions.json')
     var versionSelect = $(".version-select");
     versionSelect.select2({
         data: data
-    })
+    });
     versionSelect.select2('val', currentRevision);
 
     versionSelect.on('change', function() {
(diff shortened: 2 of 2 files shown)


michaelrkn:
Closes #206.

rwjblue:
What happens if the guide page you were on in an older version, does not exist in a newer one?

michaelrkn:
You'll get the 404 page, which I updated to be a bit more helpful:

```
This page doesn't exist in this version of the Guides.
Try hitting the back button to go back to where you were,
or go to the <a href="/">Guides homepage</a>.
If you expected something else to be here, please
<a href="https://github.com/emberjs/website/issues">file a ticket</a>.
```

I'm open to other ideas, but this seemed like the most straightforward way to handle it.

rwjblue:
Going to the root of the new version (`/v1.xx.0/`) seems better (vs 404 page) when the specific URL doesn't work.  if we can do it of course...

michaelrkn:
I agree that would be the ideal approach. However, there are trade-offs...

The simplest thing development-wise is to handle the logic client-side, like this:

```
var versionSelectBox = $(".version-select");
versionSelectBox.on('change', function() {
  var version = this.value;
  var newUrl = window.location.href.replace(currentRevision, version);
  
  var exists = $.ajax({type: 'HEAD', url: newUrl});
  
  if (exists === 200) {
    window.location = newUrl;
  } else {
    window.location.href.split(currentRevision)[0] + version;
  }
});
```

However, that adds in the extra HEAD request before making the request for the actual page, which is a bit inefficient.

The alternative is to go server-side. I assume we'd still want to display a 404 for "regular" invalid requests, so that users don't get confused if they try to go to a non-existent URL that they believe should exist and end up on the version root instead. So, we'd need to add something to the request indicate when a request comes from switching versions. Then we'd need to build a Middleman extension to redirect if the page doesn't exist.

What do you think?

wagenet:
Would there be a big issue with doing a pre-emptive HEAD request? In most cases the user won't change the dropdown immediately. We could also display a warning somewhere that they're looking at an old version.

wagenet:
I also wonder if there's a way to improve the SEO for this too. Right now whenever I Google I just get v1.10 stuff.

michaelrkn:
Pre-emptive requests: We'd have to make several requests, one for each
version. We'd also need to handle what happens if a user tries to change
versions before all of the HEAD requests finish. We gain with front-loading
the requests, but we lose with having to make lots of requests and
introducing more code complexity. For people on slow connections, I'm not
sure they'd actually gain anything. What do you think?

Warning for old version: Great idea. Can you open an issue for it?

SEO: Also worth opening an issue for.

trek:
Looks good! Thanks.

PR #311, opened 2015-05-20
fixes broken link when previous_chapter.url is an empty string

roseweixel:
This broken link results in a 404:
This broken link results in a 404


tomdale:
@roseweixel Thank you Rose!

PR #320, opened 2015-05-24
Fix 'which will resolves' phrase (line 8)

kgautreaux:
I know this is rewritten on master and reads much better in the new version, but the current phrasing for the deployed 1.11.0 branch bothered me so I thought I'd take a crack at fixing it. :-)


tomdale:
@trek @rwjblue How do we handle updates to already-deployed branches?

trek:
We typically don't, in the same way we don't update already released versions of Ember unless there is something incredibly broken that requires it.

kgautreaux:
Okay, thanks for the clarification. Cheers.  

--------------  
Keith

PR #327, opened 2015-05-27
Minor error fix in the actions helper chapter

enthrops:


tomdale:
I don't think this change is correct?

enthrops:
Whoops, misread that, sorry

PR #333, opened 2015-06-01
Update example so nothing is hidden behind flag

smeriwether:
Part of the this code snippet was hidden behind the "HTML" flag of the code box. Adding a new line makes it so all code is viewable.


tomdale:
@smeriwether Just in case this happens again, would you be interested in submitting a PR that adds overflow:scroll to examples like this?

PR #335, opened 2015-06-04
Updated description of the {{debugger}} template helper

aflores:
this is a minor documentation change about the {{debugger}} template. I think this is applicable since HTMLBars was introduced but for sure since 1.11

### DIFF
--- a/source/templates/development-helpers.md
+++ b/source/templates/development-helpers.md
@@ -27,7 +27,7 @@ you the ability to inspect the current rendering context:
 Just before the helper is invoked two useful variables and a helper are defined:
 
 * `view` The current view.
-* `context` The current context. This is likely a controller.
+* `context` The current context. This is likely a component.
 * 'get(<path>)` A helper to lookup properties.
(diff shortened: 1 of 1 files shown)


aflores:
I think PR#312 by nflbeezy is also related to this but in his case he only talks about 'context' maybe that applied to an earlier version

source/templates/development-helpers.md
```diff
@@ -24,17 +24,17 @@ you the ability to inspect the current rendering context:
...
 {{debugger}}
 ```
-Just before the helper is invoked two useful variables are defined:
+Just before the helper is invoked two useful variables and a helper are defined:
 
-* `templateContext` The current context that variables are fetched from. This
-  is likely a controller.
-* `typeOfTemplateContext` A string describing what the templateContext is.
+* `view` The current view.
+* `context` The current context. This is likely a controller.
```
rwjblue: "likely a component" (lets future proof :smile: )

PR #337, opened 2015-06-05
Removes usage of views in guides

tomdale:
As of Ember 2.0, views are deprecated and new users should not use them. For the 2.0 documentation, let's make sure we move examples and prose to refer to templates and components rather than views.


rwjblue:
Rebase please?

tomdale:
@rwjblue Rebased.

PR #338, opened 2015-06-05
Removes unused guide sections

tomdale:
The guides.yml file has accumulated some cruft from guides that have been commented out but not deleted. As we approach 2.0, this seems like a good opportunity to garbage collect guides that have not received much love recently or that may not be as relevant for new users given the programming model changes.


trek:
YAAAASSSS.

PR #340, opened 2015-06-06
Add `key` semantics to the `{{#each}}` helper guide

martypenner:
Documentation is hard, but I figure a starting point is better than nothing, especially with 1.13.0 arriving soon! I welcome any critique; I'm not very good at writing (yet).

refs https://github.com/emberjs/ember.js/issues/11117

### DIFF
--- a/source/templates/displaying-a-list-of-items.md
+++ b/source/templates/displaying-a-list-of-items.md
@@ -38,14 +38,13 @@ the DOM will be updated without having to write any code.
   ```
   When this `{{#each}}` re-renders, Ember will match up the previously rendered
   items (and reorder the generated DOM elements) based on each item's `id`
-  property.
+  property. Make sure the value you pass to `key` is unique!
+  
   There are a few special values for `key`:
     * `@index` - The index of the item in the array.
     * `@item` - The item in the array itself.  This can only be used for arrays of strings
       or numbers.
     * `@guid` - Generate a unique identifier for each object (uses `Ember.guidFor`).
-
-Make sure the value you pass to `key` is unique!
 
 The `{{#each}}` helper can have a matching `{{else}}`.
 The contents of this block will render if the collection is empty:
--- a/source/templates/displaying-a-list-of-items.md
+++ b/source/templates/displaying-a-list-of-items.md
@@ -23,15 +23,29 @@ The above example will print a list like this:
 
 Like everything in Handlebars, the `{{#each}}` helper is bindings-aware.
 If your application adds a new item to the array, or removes an item,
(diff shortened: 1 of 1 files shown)


rwjblue:
@tomdale and/or @trek would you mind reviewing?  I'd like to land *something* in the 1.13 guides to explain that you should be using `key=` with `{{each}}`...

source/templates/displaying-a-list-of-items.md
```diff
@@ -23,13 +23,21 @@ The above example will print a list like this:
 
 Like everything in Handlebars, the `{{#each}}` helper is bindings-aware.
 If your application adds a new item to the array, or removes an item,
-the DOM will be updated without having to write any code.
+the DOM will be updated without having to write any code. Because
+Glimmer (Ember's rendering engine) will try to re-use DOM whenever
+possible, it needs a way to uniquely identify each loop item in order
+to prevent unexpected results. This is done by passing the `key`
+attribute to the `{{#each}}` helper. The value of the `key` attribute
+will be the property Ember looks up on the loop object, in this case
```
tomdale: "Loop object" is terminology I'm not familiar with; maybe something like "each object being iterated over" or something?

martypenner: That certainly sounds better. However, I just read @rwjblue's API docs, and they sound way better to me. Any objection to using them instead? I'd rather have the guide be more complete than this PR currently suggest. Here is his docs for comparison, with one small change from me:

### Specifying Keys
  The `key` option is used to tell Ember how to determine if the array being
  iterated over with `{{#each}}` has changed between renders. By helping Ember
  detect that some elements in the array are the same, DOM elements can be
  re-used, significantly improving rendering speed and preventing unexpected
  results. For example, here's the `{{#each}}` helper with its `key` set to
  `id`:
  ```handlebars
  {{#each people key="id" as |person|}}
  {{/each}}
  ```
  When this `{{#each}}` re-renders, Ember will match up the previously rendered
  items (and reorder the generated DOM elements) based on each item's `id`
  property.
  There are a few special values for `key`:
    * `@index` - The index of the item in the array.
    * `@item` - The item in the array itself.  This can only be used for arrays of strings
      or numbers.
    * `@guid` - Generate a unique identifier for each object (uses `Ember.guidFor`).

rwjblue: Seems good to me, also @tomdale cleaned up what I had originally written to be good :dancer:

martypenner:
This should do it.

trek:
We can always iterate!

PR #342, opened 2015-06-06
Update Google Analytics guide embed code

seanahern:
The embed code in the current version of this guide does not work with the corresponding Router code. `ga()` is not available to the Router. Updating to the latest embed code provided by Google makes this example work again without editing the Router code.


locks:
While this PR is older than #360, the latter seems more complete. Can you give it a look @seanahern and confirm?

michaelrkn:
This can be closed in favor of #360. @rwjblue

PR #345, opened 2015-06-07
update models section to reflect renamed store methods

csantero:
Based on the changes in https://github.com/emberjs/data/pull/3235 and https://github.com/emberjs/data/pull/3241

### DIFF
--- a/source/models/finding-records.md
+++ b/source/models/finding-records.md
@@ -1,9 +1,9 @@
 The Ember Data store provides a simple interface for finding records of a single
 type through the `store` object's `find` method. Internally, the `store`
-uses `find`, `findAll`, and `findQuery` based on the supplied arguments.
+uses either `findAll` or `findRecord` based on the supplied arguments.
 
 The first argument to `store.find()` is always the record type. The optional second
-argument determines if a request is made for all records, a single record, or a query.
+argument determines if a request is made for all records or just a single record.
 
 ### Finding All Records of a Type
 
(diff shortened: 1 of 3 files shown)


HeroicEric:
there's another reference at https://github.com/emberjs/guides/blob/master/source/models/handling-metadata.md

csantero:
@HeroicEric Thanks, I've pushed a new diff with that change

source/models/finding-records.md
```diff
@@ -1,9 +1,9 @@
 The Ember Data store provides a simple interface for finding records of a single
 type through the `store` object's `find` method. Internally, the `store`
-uses `find`, `findAll`, and `findQuery` based on the supplied arguments.
+uses either `findAll` or `findRecord` based on the supplied arguments.
 
 The first argument to `store.find()` is always the record type. The optional second
-argument determines if a request is made for all records, a single record, or a query.
+argument determines if a request is made for all records or just a single record.
```
tomdale: Isn't this behavior deprecated and people should use `findAll`? @bmac

HeroicEric: Those are being updated separately https://github.com/emberjs/guides/pull/346

trek:
cc @igorT

michaelrkn:
I think this should get merged in so that #346 can move forward with finishing the job (it already needs rebasing). @locks or @rwjblue can you do a final sanity check on this?

locks:
:+1:

PR #348, opened 2015-06-09
Update displaying-the-number-of-incomplete-todos

jdhines:
`this.filterBy` caused a run-time error: "this.filterBy is not a function"
and without `model.@each.isCompleted` the `remaining` property wouldn't get updated.

### DIFF
--- a/source/getting-started/displaying-the-number-of-incomplete-todos.md
+++ b/source/getting-started/displaying-the-number-of-incomplete-todos.md
@@ -17,8 +17,8 @@ actions: {
 },
 
 remaining: function() {
-  return this.filterBy('isCompleted', false).get('length');
-}.property('@each.isCompleted'),
+  return this.get('model').filterBy('isCompleted', false).get('length');
+}.property('model.@each.isCompleted'),
 
 inflection: function() {
   var remaining = this.get('remaining');


trek:
This should have been correct as of 1.10. Are you using a newer version of Ember?

jdhines:
@trek It looks like it's actually Ember 1.8, which is what the [guides](http://guides.emberjs.com/v1.10.0/getting-started/obtaining-emberjs-and-dependencies/) link to. Maybe with Ember 1.10 this would work, but I couldn't even swap 1.10 in place of 1.8 and have the Getting Started app run at all up to the point of "Displaying the Number of Incomplete Todos." So if the guide is written for 1.8, and doesn't run in 1.10, either this PR, and possibly another one to reference the version of Ember used in the guides, needs to be made, or else the guide should use 1.10 and be updated accordingly. 
So this PR is just to make it so that the what's in the guide actually works.

trek:
Can you commit your progress through the app at this point onto Github somewhere?

jdhines:
@trek: here is my [finished tutorial code](https://github.com/jdhines/ember-getting-started).
I did deviate slightly by trying to use a component instead of `itemController` but other than that I tried to follow the instructions as close as possible.

trek:
You're seeing `"this.filterBy is not a function"` because your version the TodosController is an instance of `Ember.Controller` and not `Ember.ArrayController` (which shows up in [this step](http://guides.emberjs.com/v1.10.0/getting-started/creating-a-new-model/)).

`ArrayController`s have a `filterBy` method, but their parent class (`Controller`) do not.

PR #353, opened 2015-06-12
Simplify format for skipping toc entries

michaelrkn:
This PR makes it easier to understand how to skip TOC entries. It also makes skipping guides and chapters consistent, which will pave the way towards arbitrary levels of nesting and significantly cleaning up the code.

See the last section of https://github.com/emberjs/guides/issues/10#issuecomment-108136584 for more details.

### DIFF
--- a/data/guides.yml
+++ b/data/guides.yml
@@ -1,9 +1,9 @@
 - title: "Guides and Tutorials"
   url: 'index'
+  skip_toc: true
   chapters:
     - title: "Ember.js Guides"
       url: ""
-      skip_sidebar: true
 
 - title: "Getting Started"
   url: 'getting-started'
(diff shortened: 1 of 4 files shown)


trek:
The guides are not skipped entirely. They should still appear in the previous/next links at the bottom of each guide, they just don't get their own sidebar entry.

michaelrkn:
Are you commenting on the behavior or the naming?

If behavior, this PR leaves that behavior intact - there are still previous/next links for each guide. This PR does not change any behavior.

If naming, how about namign the key `skip-toc` to indicate that it will not be included in the TOC, even though it will still be included in previous/next links?

trek:
I was commenting on the naming.

michaelrkn:
Okay, I've updated the name to `skip_toc`.

PR #356, opened 2015-06-14
Update Model introduction for Ember Data 1.0

tomdale:
This PR updates the Model section of the guides to update idioms and advice for Ember 2.0 and Ember Data 1.0.

I will be working on updates for the remaining guides. This section (Models) will focus on apps using conventional adapters (e.g., JSON API, Firebase, etc.).

I will be adding an additional section that goes into depth on creating a custom adapter, but in an incremental way. For example, "Right now I'm using $.ajax in my model hook, how do I migrate to Ember Data?" (a very easy process, actually!).


fivetanley:
I guess serializers aren't part of the overall conceptual overview?

tomdale:
@fivetanley Nah, I think the nuance between adapters and serializers is too subtle for the introduction. For most people they are coupled. I am working on a separate "how to write an adapter" guide that will go into detail on serializers.

PR #357, opened 2015-06-14
Add guide for {{each-in}}

tomdale:


source/templates/displaying-the-keys-in-an-object.md
```diff
@@ -0,0 +1,131 @@
...
+      <li>WhistlePig</li>
+      <li>High West</li>
+    </ol>
+  </li>
+</ul>
+```
+
+### A Note About Automatic Updates
+
+Note that the `{{#each-in}}` helper does **not** observe property
```
trek: trolling me?

rwjblue: Dang, sorry didn't see your comment when I merged. I can revert if you'd like?

rwjblue: I think the concept that a rerender is needed when you add props is important information IMHO. Can we remove the 'Note that' lead in from this sentence?

tomdale: ABT — Always Be Trolling

PR #358, opened 2015-06-15
Change each-in section title

tomdale:


source/templates/displaying-the-keys-in-an-object.md
```diff
@@ -57,7 +57,7 @@ The above example will print a list like this:
 </ul>
 ```
 
-### A Note About Automatic Updates
+### Re-rendering
 
 Note that the `{{#each-in}}` helper does **not** observe property
```
rwjblue: Can you remove "Note that" here also?

tomdale: @rwjblue Done

rwjblue:
Thank you!

PR #359, opened 2015-06-16
Command for installing required packages after app generation

osam:
Added "npm install && bower install" otherwise the user will get: Cannot find module 'ember-cli/lib/broccoli/ember-app' etc.


trek:
This already happen as part of the `ember new` execution. If you're not seeing that, please file and issue on https://github.com/ember-cli/ember-cli

PR #360, opened 2015-06-17
Update google analytics guide

alex88:
I've updated the google analytics guide to use new analytics.js instead of ga.js and set the page url as google's best practices https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#multiple-hits


seanahern:
@locks This does looks better than #342 and will be useful if other ga() events are called in a particular route. 

Saves the need for the extra { page: ... } param in an event call (https://developers.google.com/analytics/devguides/collection/analyticsjs/events#implementation)

Only thing I would recommend is preserving the title assignment. It's not critical, but something to note that is being dropped with this commit. Otherwise looks great!

tomdale:
@alex88 Can you update this to retain the `title` property as @seanahern mentioned?

I'm kind of on the fence with these cookbook guides. While they're definitely useful for new users, keeping them up-to-date with things like Google Analytics change their embed code is quite a burden on the docs team.

alex88:
Updated rebased and squashed :wink:

seanahern:
@alex88 Looks good, thanks!

@tomdale I agree, re: embed updates. The embeds used to be cross/backwards-compatible but this one (and #342) was opened to update a breaking change.

mixonic:
Per https://github.com/emberjs/guides/pull/411 the cookbooks no longer live in this repo. Thank you for the contribution, and please consider submitting it to their new community home at https://github.com/emberwatch/emberwatch

PR #362, opened 2015-06-17
Create should be a function

websiddu:


rwjblue:
I agree that `create` should be a function, but there are other CSS changes here also, can you remove those to a separate PR that is specific?

websiddu:
Moved to #384

PR #367, opened 2015-06-20
Switch from App.Model.find to this.store.find

michaelrkn:


trek:
cc @igorT

fivetanley:
confirm. this has not been a thing for quite a while since ember data 1.0 beta.1

rwjblue:
thanks!

PR #371, opened 2015-06-22
`wife` has the income value not `husband`

bryanjswift:
The example previously provided set the income on the `wife` object
during initialization and then updated (`.set`) it on the `husband`
object instead of the `wife` object. This fixes the inconsistency.


rwjblue:
The way it exists demonstrates that `Ember.computed.alias` is a two-way binding, and that changing it on `husband` also updates `wife` upstream.  It does read oddly to me, and I'm inclined to merge this update even though it now only demonstrates a one-way binding.

PR #374, opened 2015-06-23
must be controller.model not just controller

orige:


trek:
How so? This section (and example) are describing ArrayController

>You can treat the Ember.ArrayController as if it was the model itself. The benefit of this is that you can replace the controller's model at any time without having to directly notify templates and components of the change.

So shouldn't the example match this usage?

locks:
I agree with @trek, the example is contextualized. What would need to happen is the section to be updated to the current best practices.

PR #378, opened 2015-06-26
Use "model" instead of "controller" in templates

michaelrkn:
Also use meaningful `each` parameter name instead of "item"

### DIFF
--- a/source/controllers/representing-multiple-models-with-arraycontroller.md
+++ b/source/controllers/representing-multiple-models-with-arraycontroller.md
@@ -95,8 +95,8 @@ export default Ember.ArrayController.extend({
 ```
 
 ```app/templates/songs.hbs
-{{#each controller as |item|}}
-  <li>{{item.fullName}}</li>
+{{#each model as |song|}}
+  <li>{{song.fullName}}</li>
 {{/each}}
 ```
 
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm, these examples are specifically for `Ember.ArrayController`, and I think iterating `controller` is correct in that context (but it definitely should be deprecated).

michaelrkn:
I just double-checked, and both `model` and `controller` work. Even if `controller` isn't deprecated, I think it's a better practice/more future-proof to use `model`. It also makes the examples more consistent for different controllers, making it easier for newbies to understand.

rwjblue:
But `model` and `controller` do not have the same meaning.  `controller` would take other things into account (like `sortProperties` and whatnot).

rwjblue:
But I am :+1: on removing `controller` anyways, sorry for arguing needlessly.

michaelrkn:
Oh, yeah, totally. Good point.

Hey, while we're looking at this guide, I recently noticed that ember-disable-proxy-controllers breaks itemController. Should we remove the reference to itemController here?

michaelrkn:
See https://github.com/emberjs/ember.js/issues/11259 for more details.

rwjblue:
We should remove array controller and item controller from the guides completely (they will not be present in Ember 2.0.0).  @trek - Do you have any objections?

michaelrkn:
That makes sense to me, but I think it's important to replace those sections with info on what to do without an item controller, and how to sort without an array controller (those being the two subtopics covered in http://guides.emberjs.com/v1.12.0/controllers/representing-multiple-models-with-arraycontroller/). 

It might be worth re-writing http://guides.emberjs.com/v1.12.0/controllers/dependencies-between-controllers/ at this time, too.

PR #379, opened 2015-06-26
Move enumerables into object model

michaelrkn:
They were so lonely by themselves. And having so many different sections makes things more overwhelming for newbies.


rwjblue:
@trek - r?

PR #380, opened 2015-06-26
Remove application section

michaelrkn:
Its content is already covered in Naming Conventions. And having so many different sections makes things more overwhelming for newbies.


rwjblue:
@trek - r?

PR #381, opened 2015-06-26
Combine Concepts into Getting Started

michaelrkn:
There's no great reason to have them broken up into two tiny sections. And having so many different sections makes things more overwhelming for newbies.


rwjblue:
@trek - r?

PR #390, opened 2015-06-30
YAML refactor

michaelrkn:
This PR changes the nomenclature from "guides" and "chapters" to simply "pages" that can be nested. I've begun refactoring the code a bit to simplify and allow for arbitrary levels of nesting. After this is merged, I will continue to refactor and improve the ability for arbitrary levels of nesting.


tomdale:
Can you say more about the motivation for this?

michaelrkn:
Oops, sorry, see #10. Last points is to "allow arbitrary levels of nesting". I asked for [feedback](https://github.com/emberjs/guides/issues/10#issuecomment-108136584) on this proposed approach a month ago and got no reply, so hopefully it's ok :) This structure also lets us greatly simplify the code - this is just the first of a few refactors I've been working on.

tomdale:
Seems OK to me, @trek?

trek:
Seems good.

trek:
Ach, should have looked closer. The `index` naming on the introduction to each section is particular to middleman so we can have "pretty" urls.

I might have just missed where this was configured? Can you verify that this change still leave the current url structure in place?

michaelrkn:
You know what? I thought that arbitrary nesting was going to make it impossible to support using `index`, because we couldn't map the URL slugs to pages if we were dropping parts of the URL. But I just spent a bit of time playing around with it, and at least for the last URL segment, I think I was wrong. Why don't you revert this merge and let me open a new PR where I don't change the `index` segment to `intro` - I'm 99% sure I can do that pretty easily.

trek:
:+1: reverting

PR #391, opened 2015-06-30
Remove extra columns in tables from "Rendering with Helpers"

XrXr:
The tables in that article look like this at 1fc83942d9b58a6612cb11928d5e396794491935
extra columns
I think these are leftovers from the 1.13 `Ember.View` deprecation


rwjblue:
Thanks!

PR #394, opened 2015-07-02
Update Polyfill in getting-started

rmi22186:
changed 'which' to 'with'

...providing JavaScript which tests for features...

to:

providing JavaScript with tests for features..

### DIFF
--- a/source/getting-started/glossary.md
+++ b/source/getting-started/glossary.md
@@ -46,7 +46,7 @@ A validation tool which checks for common issues in your JavaScript. You'd usual
 
 
 ## Polyfill
-This is a concept that typically means providing JavaScript which tests for features that are missing (prototypes not defined, etc) and "fills" them by providing an implementation.
+This is a concept that typically means providing JavaScript with tests for features that are missing (prototypes not defined, etc) and "fills" them by providing an implementation.
 
 
 ## Promise


tomdale:
I don't think this is incorrect usage, but could probably be revised to be clearer:

> This is a term that describes code that detects features that are missing in the user's browser and "fills" them by providing an implementation.

We may want to provide a link to something like https://remysharp.com/2010/10/08/what-is-a-polyfill as well.

source/getting-started/glossary.md
```diff
@@ -46,7 +46,7 @@ A validation tool which checks for common issues in your JavaScript. You'd usual
 
 
 ## Polyfill
-This is a concept that typically means providing JavaScript which tests for features that are missing (prototypes not defined, etc) and "fills" them by providing an implementation.
+This is a concept that typically means providing JavaScript with tests for features that are missing (prototypes not defined, etc) and "fills" them by providing an implementation.
```
locks: `that tests for features` would be a better alternative. `with tests` changes the meaning of the sentence.

locks:
@rmi22186 can you add the link mentioned?

source/getting-started/glossary.md
locks: another suggestion: `This is a concept that typically means using JavaScript to test for features`

locks:
@rmi22186 poke, can you address my comments and tom's suggestion?

michaelrkn:
I'm going to close this out. If somebody wants to move it forward, feel free to address the above comments and re-open or open a new PR. Thanks!

PR #395, opened 2015-07-02
Update Ember.LinkView -> Ember.LinkComponent

truenorth:
Related: https://github.com/emberjs/ember.js/pull/11594

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -58,10 +58,10 @@ renders the following HTML:
 ```
 
 There are two ways to enable support for data attributes. One way would be to add an
-attribute binding on the view, e.g. `Ember.LinkView` or `Ember.TextField` for the specific attribute:
+attribute binding on the view, e.g. `Ember.LinkComponent` or `Ember.TextField` for the specific attribute:
 
 ```javascript
-export default Ember.LinkView.reopen({
+export default Ember.LinkComponent.reopen({
   attributeBindings: ['data-toggle']
 });
 


rwjblue:
I would prefer to suggest folks export their own link-to component vs reopening like this.

tomdale:
I agree with @rwjblue: we should have folks subclass rather than reopen the framework class. Is that what you had in mind?

truenorth:
Seems like a lot of work (a lot of code changes within an app) just to add a new attribute binding to some/all links, relative to reopening.

Can either of you guys provide me with an example of how this could be easy? I'm thinking of how you could retrofit many/all of a large app's `{{link-to}}`s with a `{{my-link-component}}`, such that it's a mechanical refactoring. 

It would seem that you'd have to build a new handlebars helper in order to continue using the same positional args

rwjblue:
The following should do it for 1.13+:

    // app/components/-link-to.js

    export default Ember.LinkComponent.extend();

truenorth:
awesome. That is quite simple. I'll add it to the guides

rwjblue:
Heh, make sure it works first :smiley:

mixonic:
@rwjblue fwiw I'm unsure I agree. `-link-to` is clearly meant to be private, and `reopen` is a public documented API. I understand the elegance in using a `-link-to` component name, but we should make it `link-to` or something if it is a publicly supported strategy.

truenorth:
One additional question -- how would one do this using globals instead of ES6?

michaelrkn:
I believe we're not discussing globals in the guides anymore. 

@truenorth @mixonic @rwjblue what's the way forward here?

tomdale:
All discussion of globals should be extracted to a transition guide that explains the legacy globals mode, and explains the rules for mapping from module examples to globals.

locks:
Thanks @truenorth. Some of the concerns raised will be addressed in the ongoing review for 2.0.

PR #396, opened 2015-07-03
Remove broken link

jmeas:
I'm not sure what this link was supposed to be pointing at, but it no longer works. I removed the link in this PR, but maybe it should either be updated, or the entire line should be removed?


tomdale:
Thanks for bringing this to our attention! That domain belongs to @wycats IIRC. I will ping him about renewing it…

tomdale:
It looks like the domain expired but it has since been renewed and should be back up shortly. Sorry about that!

PR #402, opened 2015-07-06
To keep disco dead

wakeupapril:
'Aborting transition' example was excluding the wrong decades.


locks:
@michaelrkn

michaelrkn:
LGTM, although I think we should also consider how we might be able to revive disco (perhaps a separate PR would be appropriate)

PR #403, opened 2015-07-07
Updates sample with new get/set computed property syntax

locks:


source/object-model/computed-properties.md
```diff
@@ -93,17 +93,16 @@ Person = Ember.Object.extend({
...
 
-  fullName: function(key, value, previousValue) {
-    // setter
-    if (arguments.length > 1) {
+  fullName: Ember.computed('firstName', 'lastName', function(key, value, previousValue) {
+    get(key) {
+      return this.get('firstName') + ' ' + this.get('lastName');
+    },
+    set(key, value) {
       var nameParts = value.split(/\s+/);
```
locks: This could be replaced with `let [firstName, lastName] = value.split(/\s+/);`.

source/object-model/computed-properties.md
```diff
@@ -93,17 +93,16 @@ Person = Ember.Object.extend({
   firstName: null,
   lastName: null,
 
-  fullName: function(key, value, previousValue) {
-    // setter
-    if (arguments.length > 1) {
+  fullName: Ember.computed('firstName', 'lastName', function(key, value, previousValue) {
+    get(key) {
+      return this.get('firstName') + ' ' + this.get('lastName');
```
locks: This could be replaced with:

    return `${this.get('firstName')} ${this.get('lastName')}`;

Or maybe

    let firstName = this.get('firstName');
    let lastName = this.get('lastName');

    return `${firstName} ${lastName}`;

locks:
How hard do we want to push babel.js stuff? See line notes.

rwjblue:
This seems good to me, as is.

locks:
That's what I suspected, hence the "conservative" PR. zankyou :+1:

PR #404, opened 2015-07-07
Try to explain `this._super` more explicitly

locks:
There were issues raised about the clarity of the example.
The developer has a Python background and was confused about `thing` being the name of the positional argument in both `say` methods, since they "are different things because you append the string).
This is my attempt at making the example clearer while adding as concise an explanation as possible.


locks:
@michaelrkn

michaelrkn:
lgtm

PR #406, opened 2015-07-08
YAML refactor, take 2

michaelrkn:
This is a redo of #390 that has a bit of a lighter touch so that `index` URLs can still be "prettified" by taking off that last section of the slug. 

The only functionality that changes is that "inner" links in the TOC structure now link to `#`, rather than the innermost nested page. For example (note the first `<a>` tag):

```html
<ol class="toc-level-0 selected">
  <li class="toc-level-0 ">
    <a href="#">Getting Started</a>
    <ol class="toc-level-1">
      <li class="toc-level-1 "><a href="../getting-started/">Installing Ember</a></li>
      <li class="toc-level-1 "><a href="../getting-started/glossary/">Glossary</a></li>
```

The way that I've written this code using recursion would be difficult to changes the `#` links to link to the innermost page instead. It didn't seem worth the effort to try, given that the current implementation using JavaScript to expand and collapse the sections disables the link anyways.


mixonic:
@michaelrkn would love to get this in, but it will need a rebase since I killed the cookbooks.

michaelrkn:
Rebased and conflict resolved.

PR #407, opened 2015-07-08
Updated Helper Guide

tomdale:
The updates the Handlebars helper guide to use the new syntax introduced as part of the Glimmer work, as well as generally expanding on the concepts and content.

### DIFF
--- a/source/templates/writing-helpers.md
+++ b/source/templates/writing-helpers.md
@@ -1,4 +1,4 @@
-Handlebars helpers allow you to add additional functionality to your
+Helpers allow you to add additional functionality to your
 templates beyond what is included out-of-the-box in Ember. Helpers are
 most useful for transforming raw values from models and components into
 a format more appropriate for your users.
@@ -12,8 +12,8 @@ instead of a floating point dollar value][currency-stackoverflow].
 [currency-stackoverflow]: http://stackoverflow.com/a/3730040
 
 However, if we display dollar values to our users as "100¢" instead of
-"$1.00", they may be very confused. We can write a Handlebars helper to
-help us format these values into the appropriate human-readable form.
+"$1.00", they may be very confused. We can write a helper to
+format these values into the appropriate human-readable form.
 
 Let's create a `format-currency` helper that takes an integer count of
 cents and turns it into formatted dollars.
@@ -85,7 +85,7 @@ their name.
 
 ### Helper Arguments
 
-You can pass one or more arguments to Handlebars helpers which can be used
+You can pass one or more arguments to  which can be used
 inside the function. In the above example, we passed the amount in cents
 as the first and only argument.
 
@@ -135,7 +135,7 @@ example, let's abandon our American-centric ways and update our
 `format-currency` helper to take an optional configuration for which
 currency symbol to display.
 
-Handlebars helpers allow you to pass named arguments as a JavaScript
+Helpers allow you to pass named arguments as a JavaScript
 object that contains the name of the argument along with an associated
 value.  The order in which named arguments are supplied does not affect
 functionality.
@@ -147,7 +147,7 @@ helper:
 {{format-currency 350 sign="£"}}
 ```
 
-Now our helper will print pounds sterling rather than US dollars:
+We'd like our helper to print pounds sterling rather than US dollars:
 
 ```handlebars
 £3.50
@@ -182,13 +182,9 @@ world"}}
 ```app/helpers/my-helper.js
 import Ember from "ember";
 export default Ember.Helper.helper(function(params, namedArgs) {
-  let option1 = namedArgs.option1;
-  let option2 = namedArgs.option2;
-  let option3 = namedArgs.option3;
-
-  console.log(option1); // => "hello"
-  console.log(option2); // => "world"
-  console.log(option3); // => "goodbye cruel world"
+  console.log(namedArgs.option1); // => "hello"
+  console.log(namedArgs.option2); // => "world"
+  console.log(namedArgs.option3); // => "goodbye cruel world"
 });
 ```
 
@@ -380,4 +376,3 @@ would just see this:
 Welcome back! <b>&lt;script
 type="javascript"&gt;alert('pwned!');&lt;/script&gt;</b> has joined the channel.
 ```
-
--- a/source/templates/writing-helpers.md
+++ b/source/templates/writing-helpers.md
@@ -1,62 +1,383 @@
-Sometimes, you may use the same HTML in your application multiple times. In those cases, you can register a custom helper that can be invoked from any Handlebars template.
+Handlebars helpers allow you to add additional functionality to your
+templates beyond what is included out-of-the-box in Ember. Helpers are
+most useful for transforming raw values from models and components into
+a format more appropriate for your users.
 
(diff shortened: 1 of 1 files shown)


tomdale:
@trek @locks @michaelrkn @wycats @mmun @mixonic This is ready for review.

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
-Sometimes, you may use the same HTML in your application multiple times. In those cases, you can register a custom helper that can be invoked from any Handlebars template.
+Handlebars helpers allow you to add additional functionality to your
```
mixonic: These should probably just be "Helpers" or "Ember helpers".

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+instead of a floating point dollar value][currency-stackoverflow].
 
-```app/helpers/highlight.js
-export default Ember.Handlebars.makeBoundHelper( function(value, options) {
-  var escaped = Ember.Handlebars.Utils.escapeExpression(value);
-  return new Ember.Handlebars.SafeString('<span class="highlight">' + escaped + '</span>');
+[currency-stackoverflow]: http://stackoverflow.com/a/3730040
+
+However, if we display dollar values to our users as "100¢" instead of
+"$1.00", they may be very confused. We can write a Handlebars helper to
```
mixonic: -Handlebars

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+help us format these values into the appropriate human-readable form.
```
mixonic: "help us format" -> "format"

mixonic:
@tomdale "Adding Services" notably skips over `recompute()`. Intentional? At the least, it seems that section could mention a change to the logged-in state would not trigger a recomputation.

Additionally it may be nice to mention that helpers can be called as nested helpers, for example passing a formatted currency to a component.

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+Let's now implement the helper. Helpers are just functions that take
+one or more inputs and return a single output that should be put into
+the HTML.
+
+To add a new helper, create a file with the name of the helper you want
+(e.g. `format-currency.js`) in your application's `helpers` directory.
+You can also have Ember generate the file for you from the command line:
+
+```shell
+ember generate helper format-currency
```
michaelrkn: prepend `$` for shell prompt?

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+```
+
+That file should export a function wrapped with `Ember.Helper.helper()`:
+
+```app/helpers/format-currency.js
+import Ember from "ember";
+
+export default Ember.Helper.helper(function(params) {
+  let value = params[0],
```
michaelrkn: Is `let` necessary, or can we stick with the more common `var`?

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+      dollars = Math.floor(value / 100),
+      cents = value % 100,
+      sign = '$';
+
+  if (cents.toString().length === 1) { cents = '0'+cents; }
+  return sign+dollars+'.'+cents;
```
michaelrkn: Maybe add spaces around `+` operators to make it easier to read.

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+
+In this example, we can pass a `sign` argument to our `format-currency`
+helper:
+
+```handlebars
+{{format-currency 350 sign="£"}}
 ```
 
-### Dependencies
+Now our helper will print pounds sterling rather than US dollars:
```
michaelrkn: "We'd like our helper to print pounds..." (trying not to imply that it already will somehow magically do this)

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+```app/helpers/my-helper.js
+import Ember from "ember";
+export default Ember.Helper.helper(function(params, namedArgs) {
+  let option1 = namedArgs.option1;
+  let option2 = namedArgs.option2;
+  let option3 = namedArgs.option3;
+
+  console.log(option1); // => "hello"
+  console.log(option2); // => "world"
+  console.log(option3); // => "goodbye cruel world"
```
michaelrkn: What about just `console.log(namedArgs.option3);`?

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+to `Ember.Helper.helper`.
+
+In fact, we can refactor the above stateless helper into a stateful
+helper just by making the function into a `compute` method on the class:
+
+```app/helpers/format-currency.js
+import Ember from "ember";
+
+export default Ember.Helper.extend({
+  compute(params, hash) {
```
michaelrkn: Maybe this is some fancy new syntax I don't know, but does `compute` not need to be a key whose value is a function?

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+
+    if (cents.toString().length === 1) { cents = '0'+cents; }
+    return sign+dollars+'.'+cents;
+  }
+});
+```
+
+This is exactly equivalent to the example above. You can think of the
+function version as a shorthand for the longer class form if it does not
+require any state.
```
michaelrkn: Why would we do this? Under what scenarios should a helper be stateless vs stateful?

source/templates/writing-helpers.md
rwjblue: This is ES6 concise object syntax, you can read a little bit about it [here](https://leanpub.com/understandinges6/read#leanpub-auto-method-initializer-shorthand).

michaelrkn: I suspected :) Do we want to start sprinkling new ES6 syntax throughout the guides as we update them, or update everything to be consistent in one go and then use that syntax moving forward?

locks: My take: anything written anew should make full use of ES6, for `1.13` docs we update any sections we touch since we're at it, for `2.0` given the major change everything should be consistent.
The "guide assumpts" in the intro sections ties into this.

mixonic: @michaelrkn I like the balance Tom struck here. He described destructuring in small steps, but let some simpler syntax items like concise object syntax come in without explanation.

Part making Ember-CLI an idiomatic True Path is embracing ES2015 as brought to us by babel.

michaelrkn: :thumbsup:

Maybe we should consider pointing ES6 noobs like me to a good external overview of new syntax from https://github.com/emberjs/guides/blob/master/source/getting-started/glossary.md.

source/templates/writing-helpers.md
```diff
@@ -1,62 +1,383 @@
...
+
+export default Ember.Helper.extend({
+  compute(params, hash) {
+    let value = params[0],
+        dollars = Math.floor(value / 100),
+        cents = value % 100,
+        sign = hash.sign === undefined ? '$' : hash.sign;
+
+    if (cents.toString().length === 1) { cents = '0'+cents; }
+    return sign+dollars+'.'+cents;
```
chadhietala: These plus signs need some air Tom.

PR #409, opened 2015-07-08
Adding clarification on #each array behavior

rancerbeta:
I experienced much pain trying to find out why changing an attribute that was an array was not updating the #each helper result. 

It is not clear from the documentation that such adding and removing needs to be done with pushObject. I hope this change will help save new Ember users some time.

If this is not the correct place to put such documentation, please let me know an alternative.

### DIFF
--- a/source/templates/displaying-a-list-of-items.md
+++ b/source/templates/displaying-a-list-of-items.md
@@ -23,7 +23,9 @@ The above example will print a list like this:
 
 The `{{#each}}` helper is bindings-aware.  If your
 application adds a new item to the array, or removes an item, the DOM
-will be updated without having to write any code.
+will be updated without having to write any code. Note that a [].push()
+will not update the helper. Adding items need to be done with [].pushObject, 
+and related Ember Array mixins so Ember knows to update the resulting HTML.
 
 ### Specifying Keys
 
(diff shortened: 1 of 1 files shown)


locks:
lgtm. cc @michaelrkn

michaelrkn:
Good idea - thanks for contributing! 

I'd suggest you put backticks around code (eg `[].push()`) to make it easier to read, and change the text 

"and related Ember Array mixins so Ember knows to update the resulting HTML"

to:

"and related Ember Mutable Array methods](http://emberjs.com/api/classes/Ember.MutableArray.html) so that Ember can observe the change"

Just to be a little more accurate and so people know where to find the additional documentation.

rancerbeta:
Sure! I'll put up the change in a bit.

On Fri, Jul 10, 2015 at 8:04 AM, Michael Kaiser-Nyman <
notifications@github.com> wrote:

> Good idea - thanks for contributing!
>
> I'd suggest you put backticks around code (eg [].push()) to make it
> easier to read, and change the text
>
> "and related Ember Array mixins so Ember knows to update the resulting
> HTML"
>
> to:
>
> "and related Ember Mutable Array methods](
> http://emberjs.com/api/classes/Ember.MutableArray.html) so that Ember can
> observe the change"
>
> Just to be a little more accurate and so people know where to find the
> additional documentation.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/emberjs/guides/pull/409#issuecomment-120430956>.
>



-- 
Michael C Chang

Issue #412, opened 2015-07-10
Vulnerable gem versions (found by bundle-audit)


olleolleolle:
Here is the current output of bundle-audit on this repo.

```
guides
bundle-audit
Insecure Source URI found: git://github.com/LeonB/middleman-swiftype.git
Insecure Source URI found: git://github.com/dgeb/coderay.git
Name: activesupport
Version: 4.1.9
Advisory: CVE-2015-3226
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/ruby-security-ann/7VlB_pck3hU
Title: XSS Vulnerability in ActiveSupport::JSON.encode
Solution: upgrade to >= 4.2.2, ~> 4.1.11

Name: activesupport
Version: 4.1.9
Advisory: CVE-2015-3227
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/rubyonrails-security/bahr2JLnxvk
Title: Possible Denial of Service attack in Active Support
Solution: upgrade to >= 4.2.2, ~> 4.1.11, ~> 3.2.22

Name: rack
Version: 1.6.0
Advisory: CVE-2015-3225
Criticality: Unknown
URL: https://groups.google.com/forum/#!topic/ruby-security-ann/gcUbICUmKMc
Title: Potential Denial of Service Vulnerability in Rack
Solution: upgrade to >= 1.6.2, ~> 1.5.4, ~> 1.4.6

Name: redcarpet
Version: 3.2.2
Advisory: 120415
Criticality: Unknown
URL: http://danlec.com/blog/bug-in-sundown-and-redcarpet
Title: redcarpet Gem for Ruby markdown.c parse_inline() Function XSS
Solution: upgrade to >= 3.2.3

Vulnerabilities found!
```

PR #413, opened 2015-07-10
Avoid vulnerable gems

olleolleolle:
- this passes bundle-audit
- and the middleman site still runs

See #412


locks:
@rwjblue merge this and close #412 please

PR #414, opened 2015-07-10
[Typo] Mis-take or mis-understanding ? :-)

imbrou:
*(I don't think this one needs a lot of explanation.)*
Basically just a typo in the example, or I just don't understand how you bind your `app/routes/top-charts-choose.js` file to the `this.route('choose', { path: '/' });` route you defined just upper.
By the way, a **big thumb up for this awesome guide** : it is **definitely very nice to read**, and I am so happy to notice that this is the *very first* doubt that I have about what I red from the beginning !

### DIFF
--- a/source/routing/redirection.md
+++ b/source/routing/redirection.md
@@ -74,7 +74,7 @@ Router.map(function() {
 });
 ```
 
-```app/routes/top-charts-choose.js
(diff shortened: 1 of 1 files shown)


locks:
You are correct, it is a typo indeed :) Thanks for the PR.

@rwjblue merge

PR #418, opened 2015-07-10
Fix camelized event names in template/actions.

Serabe:
Event names were camelized in actions instead of dasherized. This fix also
expects emberjs/ember.js#11711 (or similar) to fix the API documentation as
well.

This is the actual fix for emberjs/ember.js#10713 continuing the discussion in
emberjs/guides#83.


michaelrkn:
Awesome! Thanks for closing the loop on this. cc @rwjblue

PR #420, opened 2015-07-11
More 1.13 Cleanup

mixonic:
* Moar ES2015
* More controller cleanup
* Update DI guide to talk about services instead of needs


mixonic:
I will merge this in a few hours if there is no feedback. Want to keep moving.

PR #423, opened 2015-07-11
Remove "Extending Built-In Controls" section from Input Helpers

alefteris:
This section linked to the views guide that has since been removed.


rwjblue:
Thanks!

PR #427, opened 2015-07-13
Fixes #349

### LINKED ISSUE #349: The rest adapter docs is hard to follow
The rest adapter docs `http://guides.emberjs.com/v1.10.0/models/the-rest-adapter/` is using too many different pairs of names interchangeably, and this is simply confusing.

The actions table is explaining the endpoints for `store.find('post', 1)` so the table should show `/posts/123` instead of `/people/123`, etc.

Then the `inflector.irregular('formula', 'formulae');` example should be replaced by `inflector.irregular('person', 'people');` because this `person` vs `people` example is used again below.

locks:


mixonic:
Thank you sir.

locks:
:heart_decoration:

Issue #399, opened 2015-07-05
Naming Conventions Table CSS exceeds 100% width


ashleygwilliams:
on http://guides.emberjs.com/v1.12.0/concepts/naming-conventions/
table exceeds 100% width, overflow is cut off, cannot read template column

<img width="814" alt="screen shot 2015-07-05 at 5 44 43 pm" src="https://cloud.githubusercontent.com/assets/1163554/8513531/cd01c2f4-233d-11e5-89fb-ad8fce5ff829.png">

in Chrome 43.0.2357.130 (64-bit)

pmasucci:
So I would love to fix this bug, but I'm wondering if someone can point me in the right direction as far as best practices go (I'm about as junior as you can get). The problem can be solved by wrapping the table in a div with overflow:scroll. Is it okay to add that right into naming-conventions.md and then add a .scroll class in the css?

Everything in the .md files that I've looked through so far seems to be purely html with no classes so I didn't want to mix those up if it isn't okay.

eccegordo:
Just say no to scroll overflow. :smile: 

How about a tree view to represent the convention?

<table>
    <tr>
        <th>Route Name</th>
        <th>Convention</th>

    </tr>
    <tr>
        <td><code>posts</code></td>
        <td>
            <pre>
app
 ├── controllers/
 │   └── post.js  
 ├── routes/
 │   └── post.js  
 └── templates/
     └── posts.hbs
            </pre>
        </td>
    </tr>

    <tr>
        <td><code>posts.favorites</code></td>
        <td>
            <pre>
app
 ├── controllers/
 │   └── posts/  
 │      └── favorites.js  
 ├── routes/
 │   └── posts/
 │      └── favorites.js  
 └── templates/
     └── posts/
        └── favorites.hbs
          </pre>
        </td>
    </tr>

    <tr>
        <td><code>posts.post</code></td>
        <td>
            <pre>
app
 ├── controllers/
 │   └── posts/  
 │      └── post.js  
 ├── routes/
 │   └── posts/
 │      └── post.js  
 └── templates/
     └── posts/
        └── post.hbs
            </pre>
        </td>
    </tr>
</table>

Here is the markup

```

<table>
    <tr>
        <th>Route Name</th>
        <th>Convention</th>

    </tr>
    <tr>
        <td><code>posts</code></td>
        <td>
            <pre>
app
 ├── controllers/
 │   └── post.js  
 ├── routes/
 │   └── post.js  
 └── templates/
[...]

PR #430, opened 2015-07-15
[BUGFIX] #399 Improve Naming Conventions table

eccegordo:
The naming conventions table in older guides cuts off the text. And in the current guide making the text smaller is not very elegant.

This commit represents the convention as tree view. Which takes up more vertical space but is easier to visualize at a glance. And doesn't suffer the scroll overflow or font size issues.

I believe this should also be back ported to old versions of the guides. Because #399 was reported again v1.12.0 version of guides

Before

conventions-before

After

conventions-after

### DIFF
--- a/source/getting-started/naming-conventions.md
+++ b/source/getting-started/naming-conventions.md
@@ -202,33 +202,61 @@ export default Ember.Router.extend().map(function(){
 Here are the naming conventions for each of the routes defined in
 this router:
 
-<table class="extensive">
-  <thead>
-  <tr>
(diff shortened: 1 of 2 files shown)


rwjblue:
I like this :+1:, thank you!

rwjblue:
Will leave open (for a few hours) so others can review and confirm also...

michaelrkn:
Looks great!

michaelrkn:
@rwjblue can you merge and close #399?

PR #431, opened 2015-07-15
Use run loop in the example code describing the use of the run loop

ksin:
Tiny fix. Referring to this page of the documentation:
http://guides.emberjs.com/v1.10.0/understanding-ember/run-loop/

### DIFF
--- a/source/understanding-ember/run-loop.md
+++ b/source/understanding-ember/run-loop.md
@@ -92,10 +92,12 @@ However, if we have the run loop in the above code, the browser will only rerend
 
 ```javascript
 var user = User.create({ firstName: 'Tom', lastName: 'Huda' });
-user.set('firstName', 'Yehuda');
-user.set('lastName', 'Katz');
-user.set('firstName', 'Tom');
-user.set('lastName', 'Huda');
+Ember.run(function() {
+  user.set('firstName', 'Yehuda');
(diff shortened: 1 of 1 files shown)


locks:
Hm, this whole section is rather strange. I think the goal was to kind of say that there's an implicit runloop for Ember and that that code block will get scheduled in it, but it ends up confusing.

@michaelrkn are we keeping this section? might need a review and a check for update-ness.

michaelrkn:
I don't know if we're keeping it. It feels like maybe it's outside the scope of the guides... does the average Ember developer need to know about the run loop in order to successfully write Ember apps? Any thoughts, @trek?

ksin:
@michaelrkn As an average Ember developer, my input is that I would like to see as much documentation on it as possible. I feel like anyone building a non-trivial Ember app will quickly run into problems where understanding the run loop better will be very helpful.

michaelrkn:
Thanks for the input, @ksin! I definitely agree it should be documented - the question is, should it be covered in the guides or just elsewhere (eg API docs)?

locks:
@ksin thank you for raising the issue, but this needs more discussion. I've opened a meta issue and tagged you and @michaelrkn if you want to get looped in.

ksin:
Thanks @locks . Hope to get it resolved soon. :)

PR #432, opened 2015-07-16
Use post.js instead of posts.js

aamerabbas:
It seems to me that this route is referring to something that is called with a dynamic segment, so it should be post.js instead of posts.js. The sample above this also uses post.js, so it seems it should be consistent with that.

### DIFF
--- a/source/routing/rendering-a-template.md
+++ b/source/routing/rendering-a-template.md
@@ -28,7 +28,7 @@ export default Ember.Route.extend({
 If you want to use a different controller than the route handler's
 controller, pass the controller's name in the `controller` option:
 
-```app/routes/posts.js
+```app/routes/post.js
 export default Ember.Route.extend({
   renderTemplate: function() {
     this.render({ controller: 'favoritePost' });


locks:
This seems to contradict the router at the top and the previous code samples.

aamerabbas:
That's a good point. Then why is the example directly above this post.js? Shouldn't that be posts.js?

Let me know if I'm missing something here (I'm still new to Ember, so I could be confused). However, if you agree, I'll submit a pull request to change the example above this from post.js to posts.js.

Thanks

michaelrkn:
@aamerabbas You're right, it is confusing. I think it would be easier to understand if everything were singular. Could you update your PR to make the entire page consistently singular? Thank you!

locks:
This whole file seems problematic, where does https://github.com/emberjs/guides/pull/432/files#diff-67d6c402852bae1c8abbeda3b8734a32R70 even come from?

michaelrkn:
@locks I think you have to use your imagination for how the template might look

michaelrkn:
Since this has been sitting for a week, I've opened #463 and am closing this out.

PR #434, opened 2015-07-17
Switch from setupController to a simpler model hook

michaelrkn:
Keeps things more straightforward so that the reader can focus on the issue at hand, rather than getting potentially confused by a less-common way of setting the model.

### DIFF
--- a/source/controllers/representing-multiple-models.md
+++ b/source/controllers/representing-multiple-models.md
@@ -1,14 +1,10 @@
 The model of a controller can represent several records as well as a single
-one. Here we will customize the default routing logic and set the `model`
-property in a route's `setupController` method.
-
-In our route, we setup our `SongsController` to represent the songs in a
-playlist:
+one. Here, our route's `model` hook returns an array of songs:
 
 ```app/routes/songs.js
 export default Ember.Route.extend({
-  setupController(controller, playlist) {
-    controller.set('model', playlist.get('songs'));
+  model() {
+    return this.store.findAll('song');
   }
 });
 ```


source/controllers/representing-multiple-models.md
```diff
@@ -1,15 +1,11 @@
...
-playlist:
+one. Here, our router's `model` hook returns an array of songs:
 
 ```app/routes/songs.js
-  setupController(controller, playlist) {
-    controller.set('model', playlist.get('songs'));
-  }
+ model: function() {
+   this.store.findAll('song');
```
irnc: This should be `return this.store.findAll('song');`

michaelrkn:
Fixed! Thanks for the heads up, @irnc!

source/controllers/representing-multiple-models.md
```diff
@@ -1,15 +1,11 @@
...
-In our route, we setup our `SongsController` to represent the songs in a
-playlist:
+one. Here, our router's `model` hook returns an array of songs:
 
 ```app/routes/songs.js
-  setupController(controller, playlist) {
-    controller.set('model', playlist.get('songs'));
-  }
+ model: function() {
```
knownasilya: Probably keep the shorthand version, i.e. `model() {`

michaelrkn:
does this look right @knownasilya?

knownasilya:
:+1:

michaelrkn:
Thanks for the help! @rwjblue can you merge this?

source/controllers/representing-multiple-models.md
```diff
@@ -1,14 +1,10 @@
 The model of a controller can represent several records as well as a single
-one. Here we will customize the default routing logic and set the `model`
-property in a route's `setupController` method.
-
-In our route, we setup our `SongsController` to represent the songs in a
-playlist:
+one. Here, our router's `model` hook returns an array of songs:
```
HeroicEric: `router's` -> `route's`?

michaelrkn: good catch! thank you. fixed

PR #438, opened 2015-07-19
{{get}} helper documentation

jmurphyau:
This PR is to add the {{get}} helper to the guides.

### DIFF
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -60,7 +60,7 @@
       url: "actions"
     - title: "Input Helpers"
       url: "input-helpers"
-    - title: "Get Helper"
+    - title: "The Get Helper"
       url: "get-helper"
     - title: "Development Helpers"
       url: "development-helpers"
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -60,6 +60,8 @@
       url: "actions"
     - title: "Input Helpers"
       url: "input-helpers"
+    - title: "Get Helper"
+      url: "get-helper"
     - title: "Development Helpers"
       url: "development-helpers"
     - title: "Rendering with Helpers"
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -61,7 +61,7 @@
     - title: "Input Helpers"
       url: "input-helpers"
     - title: "The Get Helper"
-      url: "get-helper"
+      url: "the-get-helper"
     - title: "Development Helpers"
       url: "development-helpers"
     - title: "Rendering with Helpers"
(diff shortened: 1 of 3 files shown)


data/pages.yml
```diff
@@ -60,6 +60,8 @@
       url: "actions"
     - title: "Input Helpers"
       url: "input-helpers"
+    - title: "Get Helper"
```
michaelrkn: Maybe "The Get Helper"?

source/templates/get-helper.md
```diff
@@ -0,0 +1,38 @@
...
+          <td>{{get person field}}</td>
+        {{/each}}
+      </tr>
+    {{/each}}
+  </tbody>
+
+</table>
+```
+
+The `{{get}}` helper can also be used with the `{{input}}` helper to allow you to edit a field. You need to use the `{{mut}}` helper to achieve this:
```
michaelrkn: Isn't this only applicable once one-way binding is the default, which is not currently the case?

jmurphyau: Nah - the get helper is one way and to allow two way you need to use mut... Maybe I could explain this a bit better?

data/pages.yml
jmurphyau: Yep

tomdale:
@michaelrkn For context, the way we did two-way bindings in a backwards compatible way with the new 1WD system is by parsing templates and "auto-muting" them by rewriting the AST.

So really under the hood all component bindings are one-way by default, and then a template like this:

```hbs
{{my-component foo=bar}}
```

Is rewritten to this:

```hbs
{{my-component foo=(mut bar)}}
```

Because that happens statically, it means new features like `{{get}}` will opt you into the new mutable behavior, even if you're still using curly components with 2WB by default.

source/templates/get-helper.md
```diff
@@ -0,0 +1,38 @@
...
+
+```handlebars
+<div>
+  <label>{{fieldName}}</labe>
+  {{input value=(mut (get person fieldKey)) typ='text'}}
```
et: typo 'typ'

michaelrkn:
Thanks for the clarification! I would bet that this syntax will be really confusing to many people without any other context or background in the guides. 1-way binding lands with angle bracket components and routable components in 2.1, right? I'd suggest either 1) adding some text to the bottom noting that this syntax looks weird and linking to a blog post that explains about 1-way binding and `mut` (maybe https://github.com/emberjs/rfcs/pull/15 ?), or 2) just acknowledging that this may cause confusing for people who aren't all up-to-date on future plans for Ember but that it's not worth tackling when the guides will cover the topic in the release after next.

PR #439, opened 2015-07-20
Typo in {{#each}} helper

mxchai:
Using {{#each todos as |todo|}} fails to load the model in index.hbs. {{#each model.todos as |todo|}} should be the correct version.


rwjblue:
:+1:

rwjblue:
Thank you!

PR #440, opened 2015-07-20
Update config.rb to use pages instead of guides and chapters

michaelrkn:


michaelrkn:
@trek or @rwjblue, can you review and if good merge so that we can get the build fixed?

PR #446, opened 2015-07-21
Add language to code fence for `get` function

michaelrkn:
Without the language, the site won't build.


michaelrkn:
@trek or @rwjblue, can you merge so that we can get the build fixed?

rwjblue:
I just enabled Travis (which will run specs and `middleman build`), can you amend and push to make sure Travis agrees?

Issue #455, opened 2015-07-22
Encourage decorator-style Ember.computed/Ember.observer


michaelrkn:
There seems to be general consensus that we should encourage people to use decorator-style methods instead of prototype extensions, except in the case of Array extensions. We should change all places in the guides that use `.property` to `Ember.computed`, and `.observes` to `Ember.observer`. For example, this:

```javascript
var president = Ember.Object.create({
  firstName: "Barack",
  lastName: "Obama",

  fullName: function() {
    return this.get('firstName') + ' ' + this.get('lastName');
  }).property()
});
```

becomes:

```javascript
var president = Ember.Object.create({
  firstName: "Barack",
  lastName: "Obama",

  fullName: Ember.computed(function() {
    return this.get('firstName') + ' ' + this.get('lastName');
  })
});
```

trek:
:+1:

PR #457, opened 2015-07-22
Encourage decorator-style Ember.computed/Ember.observer

HeroicEric:
- [x] Use `Ember.computed` over `property()`
- [x] Use `Ember.observer` over `.observes()`

fixes https://github.com/emberjs/guides/issues/455

### DIFF
--- a/source/object-model/computed-properties.md
+++ b/source/object-model/computed-properties.md
@@ -2,7 +2,7 @@
 
 In a nutshell, computed properties let you declare functions as properties. You create one by defining a computed property as a function, which Ember will automatically call when you ask for the property. You can then use it the same way you would any normal, static property.
 
-It's super handy for taking one or more normal properties and transforming or manipulating their data to create a new value. 
+It's super handy for taking one or more normal properties and transforming or manipulating their data to create a new value.
 
 ### Computed properties in action
 
@@ -14,9 +14,9 @@ Person = Ember.Object.extend({
   firstName: null,
   lastName: null,
 
-  fullName: function() {
+  fullName: Ember.computed('firstName', 'lastName', function() {
     return this.get('firstName') + ' ' + this.get('lastName');
-  }.property('firstName', 'lastName')
+  })
 });
 
 var ironMan = Person.create({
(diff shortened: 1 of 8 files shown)


source/object-model/computed-properties.md
```diff
@@ -26,20 +26,28 @@ var ironMan = Person.create({
...
-  fullName: Ember.computed('firstName', 'lastName', function() {
-    return this.get('firstName') + ' ' + this.get('lastName');
-  })
+fullName: function() {
+  return this.get('firstName') + ' ' + this.get('lastName');
+}.property('firstName', 'lastName')
 ```
 
+This is possible because Ember extends the `function` prototype. This is no
+longer the recommended syntax.
```
HeroicEric: Wasn't sure whether or not to just completely remove this and the other one like it for `.observes()`

michaelrkn: @stefanpenner @mixonic @tomdale do any of you have any thoughts about this?

michaelrkn:
Awesome!!! Let's get some feedback on your question and then get this merged in!

source/object-model/computed-properties.md
michaelrkn: Ooh, I have an idea. We could take this out of here and move https://github.com/emberjs/guides/blob/master/source/configuring-ember/disabling-prototype-extensions.md into this Object Model section. We'll just need to change the wording a bit to make it flow right. Does that seem okay?

tomdale: I am very much not in favor of integrating caveats about prototype extensions into the main flow.

I'd like to see us pick a happy path and talk only about that happy path in the primary documentation. Anything that deviates from the happy path can be moved to an auxiliary guide. If we make non-proto-extension syntax the default, let's update to that everywhere and create a legacy appendix describing how that syntax works for folks who need to work on code bases that use prototype extension.

locks: I second removing proto-extensions references from the guides, especially since we have versioning. If people think they were removed and can't use them, well we're halfway to our goal ;)

HeroicEric:
Updated and removed the inline references to the old syntax in the guides for computed properties and observers.

PR #461, opened 2015-07-22
Update usage of deprecated ArrayController

zacstewart:
Replace usage of the deprecated ArrayController with Controller.

### DIFF
--- a/source/object-model/classes-and-instances.md
+++ b/source/object-model/classes-and-instances.md
@@ -113,10 +113,10 @@ Person.create({
 // alerts "Stefan Penner, reporting for duty!"
 ```
 
-If you are subclassing a framework class, like `Ember.View`, and you override
-the `init` method, make sure you call `this._super()`! If you don't, the system
-may not have an opportunity to do important setup work, and you'll see strange
-behavior in your application.
+If you are subclassing a framework class, like `Ember.Component`, and you
+override the `init` method, make sure you call `this._super()`! If you don't,
+the system may not have an opportunity to do important setup work, and you'll
+see strange behavior in your application.
 
 ### Accessing Object Properties
 
--- a/source/object-model/classes-and-instances.md
+++ b/source/object-model/classes-and-instances.md
@@ -113,11 +113,10 @@ Person.create({
 // alerts "Stefan Penner, reporting for duty!"
 ```
 
-If you are subclassing a framework class, like `Ember.View` or
-`Ember.ArrayController`, and you override the `init` method, make sure
-you call `this._super()`! If you don't, the system may not have an
-opportunity to do important setup work, and you'll see strange behavior
-in your application.
+If you are subclassing a framework class, like `Ember.View`, and you override
+the `init` method, make sure you call `this._super()`! If you don't, the system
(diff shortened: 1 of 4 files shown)


zacstewart:
We may want to update this in all the guides before merging.

michaelrkn:
Thanks for the PR! Yes, could you go through all the guides and remove references to `ArrayController`?

source/getting-started/naming-conventions.md
```diff
@@ -178,10 +178,14 @@ template with the app's instance of `app/controllers/post.js`.
...
-`ArrayController`. Otherwise, you get an `ObjectController`. As of 1.11,
-`ObjectController`s are deprecated and if you try to use its proxy feature you
-will get a warning. Please, take a look at the [deprecations guide](http://emberjs.com/deprecations/v1.x/) for more
-detailed information.
+`ArrayController`. Otherwise, you get an `ObjectController`.
+
+As of 1.11, `ObjectController`s are deprecated and if you try to use its proxy
+feature you will get a warning. Likewise, as of 1.13, `ArrayController`s are
+deprecated.  However, unlike `ObjectController`, using the proxying behavior of
+an auto-generated `ArrayController` will not throw a deprecation message.
```
locks: are you sure this is correct?

locks: this is also incomplete/incorrect, since `ember-cli` includes an addon that removes the generation of `ObjectController` and `ArrayController`. I would suggest opening a separate PR for this.

source/object-model/classes-and-instances.md
```diff
@@ -113,11 +113,10 @@ Person.create({
 // alerts "Stefan Penner, reporting for duty!"
 ```
 
-If you are subclassing a framework class, like `Ember.View` or
-`Ember.ArrayController`, and you override the `init` method, make sure
-you call `this._super()`! If you don't, the system may not have an
-opportunity to do important setup work, and you'll see strange behavior
-in your application.
+If you are subclassing a framework class, like `Ember.View`, and you override
```
locks: since `Ember.View` is also deprecated can you use `Ember.Controller` instead?

zacstewart: I thought about that, but from looking in `Ember.Controller`, I didn't see that it had an `init` method. I think a subsequent PR should remove references to `Ember.View`, but I'm not sure of how to handle this paragraph.

source/getting-started/naming-conventions.md
zacstewart: It seems to be what is explained [here](http://emberjs.com/deprecations/v1.x/#toc_arraycontroller), but maybe I misunderstand it.

locks: it's incomplete because it fails to mention the `ember-disable-proxy-controllers` addon.

locks: please open a separate PR for this change so we can discuss it proper, thanks :)

source/object-model/classes-and-instances.md
locks: actually, can you change it to `Ember.Component`? it's the most direct update

PR #462, opened 2015-07-23
Updates for shorthand method literals syntax

locks:


locks:
@michaelrkn confirm and merge?

source/getting-started/naming-conventions.md
```diff
@@ -91,11 +91,13 @@ to the template.
...
 ```app/routes/favorites.js
+import ajax from 'ic-ajax';
+
-  model: function() {
+  model() {
     // the model is an Array of all of the posts
     // fetched from this url
-    return $.ajax('/a/service/url/where/posts/live');
+    return ajax('/a/service/url/where/posts/live');
```
michaelrkn: This is a thing I'm not familiar with - you don't need to prepend `$.` to `ajax`? What is it called on?

michaelrkn:
wow! nice work!

source/getting-started/naming-conventions.md
locks: see https://github.com/emberjs/guides/pull/462/files#diff-414c60298bd3dd9a7aa15654913a5bddR94

locks: I added the import since it's probably not changing anytime soon.

michaelrkn: ah missed that. thanks!

PR #463, opened 2015-07-23
Be consistent in "post" vs "posts"

michaelrkn:


michaelrkn:
@locks can you review?

locks:
What's confusing about https://github.com/emberjs/guides/pull/432/files#diff-67d6c402852bae1c8abbeda3b8734a32R70 to me is that the previous snippets have two named outlets, neither of which is `post`. Might open a separate issue.

PR #467, opened 2015-07-24
Makes `app/router.js` code sample more alike ember-cli's file

locks:
The first time I saw the code I got a bit confused because we're not used to seeing it like that. Hopefully this way is less surprising for the reader.


locks:
@michaelrkn review+merge

PR #469, opened 2015-07-24
Updates `Ember.View` reference to `Ember.Component`

locks:


locks:
I know there's another `Ember.View` reference, but it's hopefully being dealt with in https://github.com/emberjs/guides/pull/461/files#r35377980

PR #474, opened 2015-07-24
Modify data attribute bindings example to support bound attrs

sohara:
This fixes an issue in the existing code example for adding a `data-` attribute to an core Ember View, such as LinkView. Currently the code supports assigning strings to data attributes just fine. But if one tries to bind a property a data attribute on a view, e.g.:
```handlebars
 {{#link-to route id data-result-number=dataResultNumber}}
```
You end up with an anchor such as:
```html
<a id="ember1670" class="ember-view" href="/xyz/123" data-result-number="1" data-result-numberBinding=[Object object]> 
```
This modification to the example avoids that problem.

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -90,7 +90,7 @@ export default Ember.View.reopen({
 
     // bind attributes beginning with 'data-'
     Ember.keys(this).forEach(function(key) {
-      if (key.substr(0, 5) === 'data-') {
+      if (key.substr(0, 5) === 'data-' && key.substr(key.length - 7) !== 'Binding') {
         self.get('attributeBindings').pushObject(key);
       }
     });


mmun:
The entire `data-*` section should be removed.

locks:
@sohara can you address the previous comments? If not I'll submit a separate PR and close this issue.

sohara:
@mmun @locks I'm open to documenting other approaches to adding data attribute support to core Ember views. Is that what you had in mind for your PR? I agree that there's nothing inherently special about data attributes but I'm guessing the use cases that devs run up against when trying to add unsupported attributes to core Ember views most often involve data attributes. I don't think we should remove the section unless it's replaced by something else. Adding attributes that aren't supported out of the box to core Ember views is something that a lot of devs will want/need to do, so some kind solution should be documented imo.

michaelrkn:
We're just going to remove this whole section per @mmun's suggestion. @sohara I'd suggest opening an issue to discuss adding documentation on what you're trying to do here, or better yet, write a blog post and then open an issue linking to it, so that the content gets out there without any of us holding it up, and then we can talk about incorporating it.

mmun:
I agree with @sohara that we should document how to setup some attributeBindings. We just don't need to document the dynamic case.

PR #475, opened 2015-07-25
The action name should be "select" only

szines:
> At the moment "selectPost" is there as action name in template, however only `select()` in the component js.
> This was the original version:

```app/templates/component/show-posts.hbs
<button {{action "selectPost" model}}>Select Post</button>
```

You can implement a function that responds to the button being clicked
by adding an `actions` hash to your component with a method called
`select`:

```app/components/show-posts.js
  actions: {
    select(post) {
      // do your business.
    }
  }
});
```


michaelrkn:
Thanks for catching this!

PR #480, opened 2015-07-28
Add `import`s and fix an example

ericdouglas:
With the `import Ember`, everyone will be able to just copy the code and have a *running* example.


michaelrkn:
Thanks for your contribution!

ericdouglas:
@michaelrkn :smile:

PR #481, opened 2015-07-28
Get version from url instead of hard-coding for swiftype

michaelrkn:
This will eliminate one more step in the build/deploy process.

### DIFF
--- a/data/search.yml
+++ b/data/search.yml
@@ -1,3 +1,1 @@
 engine_key: 'ha9YM1kgBdihxrq_8ZHK'
-document_type: 'guides-v1-10-0'
-revision: 'v1.10.0'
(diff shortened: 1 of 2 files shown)


rwjblue:
If it is removing a step from deployment process, can you update the README also?

michaelrkn:
See https://github.com/emberjs/guides.emberjs.com/pull/6.

trek:
hold up:

```
var slugs = window.location.href.split('/');
var currentRevision = slugs[slugs.length - 2];
var documentType = 'guides-' + currentRevision.replace(/\./g,'-');
```

certainly works for `http://guides.emberjs.com/v1.13.0`

but does it work for 'http://guides.emberjs.com/v1.13.0/ember-inspector/installation/' where the version is located in the `slugs.length - 4` position?

PR #482, opened 2015-07-28
[Components] Positional Params

knownasilya:
Added the idea of positional params to the passing properties section, since it was missing previously.

### DIFF
--- a/source/components/passing-properties-to-a-component.md
+++ b/source/components/passing-properties-to-a-component.md
@@ -84,8 +84,33 @@ create a component for each item and bind it to each model in the loop.
 <script src="http://static.jsbin.com/js/embed.js"></script>-->
 
 If you are using the `{{component}}` helper to render your component, you can
-pass properties to the chosen component the same manner:
+pass properties to the chosen component in the same manner:
 
 ```handlebars
 {{component componentName title=title name=name}}
+```
+
+### Positional Params
+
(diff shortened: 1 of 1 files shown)


source/components/passing-properties-to-a-component.md
```diff
@@ -84,8 +84,33 @@ create a component for each item and bind it to each model in the loop.
...
+
+```app/components/x-visit.js
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  positionalParams: ['name', 'model']
+});
+```
+
+This will expose the first parameter as the `{{name}}` attribute and the second as `{{model}}`. They will also be available as regular attributes in your source via `this.get('name')` and `this.get('model')`.
```
michaelrkn: How about "This will expose the first parameter _in your template_ as the ...", and "They will also be available as regular attributes in your _component_ via ..."?

source/components/passing-properties-to-a-component.md
```diff
@@ -84,8 +84,33 @@ create a component for each item and bind it to each model in the loop.
...
+
+In addition to mapping the parameters to attributes, you can also have an arbitrary number of parameters by setting `positionalParams`
+to a string, e.g. `positionalParams: 'params'`. This will allow you to access those params like so:
```
michaelrkn: How about "This will allow you to access those params _as an array_ like so:"?

knownasilya:
Changes made, thanks :+1:

michaelrkn:
Thanks a bunch for your contribution!

Issue #483, opened 2015-07-28
Remove JSBins


michaelrkn:
Rationales: They can't be PR'd, there is no indication if they get broken, it's not obvious when they are out of date, and they are either owned by somebody who might be hard to track down or need a shared login that doesn't currently exist. 

We need to go through the Guides and remove all JSBins, replacing with code examples where appropriate.

# Checklist

- [x] https://github.com/emberjs/guides/blob/master/source/components/passing-properties-to-a-component.md - https://github.com/emberjs/guides/pull/489
- [x] https://github.com/emberjs/guides/blob/master/source/components/wrapping-content-in-a-component.md - https://github.com/emberjs/guides/pull/490
- [x] https://github.com/emberjs/guides/blob/master/source/components/handling-user-interaction-with-actions.md - https://github.com/emberjs/guides/pull/491
- [x] https://github.com/emberjs/guides/blob/master/source/components/index.md - https://github.com/emberjs/guides/pull/492
- [x] https://github.com/emberjs/guides/blob/master/source/components/customizing-a-components-element.md - https://github.com/emberjs/guides/pull/488
- [x] https://github.com/emberjs/guides/blob/master/source/components/sending-actions-from-components-to-your-application.md - https://github.com/emberjs/guides/pull/493
- [x] https://github.com/emberjs/guides/blob/master/source/routing/specifying-a-routes-model.md - https://github.com/emberjs/guides/pull/495
- [x] https://github.com/emberjs/guides/blob/master/source/components/defining-a-component.md - https://github.com/emberjs/guides/pull/494

# More checklist

* [ ] There is a script in the layout file that has sth to do with jsbins, probably can be removed once this gets merged – didn't see any difference locally after removing it. [link](https://github.com/emberjs/guides/blob/6d1ba3eefa223fa3ba056f1c219b09cc0cdceba7/source/layout.erb#L114)
* [x] There is a link to a broken JSBin here in the guide for loading and error substates: [link](https://github.com/emberjs/guides/blob/4ddd8393466609b45e40492e7dc0b11c75bd400e/source/routing/loading-and-error-substates.md#L240) (https://github.com/emberjs/guides/pull/510)
* [ ] There's also a long list of JSBin examples in the guide for query params: [link](https://github.com/emberjs/guides/blob/dcf3c6ca0e139d9b68338dae3d862bc214351199/source/routing/query-params.md#L328)
[...]

PR #488, opened 2015-07-29
Remove JSBin from components/customizing-a-components-element

mamut:
Meta PR: https://github.com/emberjs/guides/issues/483

This bin was commented out since march.


locks:
That was easy! :)

PR #497, opened 2015-07-30
Add Ember.RSVP.hash

hank-harvey:


source/controllers/representing-multiple-models.md
```diff
@@ -50,3 +52,39 @@ Now we can use this property in our template:
 
 {{longSongCount}} songs over 30 seconds.
 ```
+
+### Representing Several Models
+
+Multiple models can be returned through `Ember.RSVP.hash`. To continue with the above example, if we also wanted to return the `models/album.js` records then our `model` hook would look like this:
```
locks: You mention `Ember.RSVP.hash` but you don't explain why it is necessary for the model hook to work properly. Say something like "we can also return a hash from the model hook, but we need rsvp.hash for it to resolve all of the promises etc"

locks: `model:album`, not `models/album.js`

source/controllers/representing-multiple-models.md
```diff
@@ -50,3 +52,39 @@ Now we can use this property in our template:
...
+  model() {
+    return Ember.RSVP.hash({
+      songs: this.store.findAll('song'),
+      albums: this.store.findAll('album')
+    });
+  }
+});
+```
+
+In the `songs` template, we can specify both models and use the `{{#each}}` helper to display
```
locks: You could also add the `longSongCount` computed property from the previous example, to make a parallel about how you have to use `model.songs.@each` instead of `model.@each`.

source/controllers/representing-multiple-models.md
```diff
@@ -50,3 +52,53 @@ Now we can use this property in our template:
...
+    });
+  }
+});
+```
+
+The controller with the computed property would then be:
+
+```app/controllers/songs.js
+export default Ember.Controller.extend({
+  longSongCount: Ember.computed('model.@each.duration', function() {
```
locks: should be `model.songs.@each.duration`

source/controllers/representing-multiple-models.md
```diff
@@ -50,3 +52,53 @@ Now we can use this property in our template:
...
+    let songs = this.get('model.songs');
+    let longSongs = songs.filter((song) => {
```
locks: this can be shortened to `songs.filter(song => song.get('duration') > 30);`

source/controllers/representing-multiple-models.md
```diff
@@ -50,3 +50,60 @@ Now we can use this property in our template:
...
+  model() {
+    return Ember.RSVP.hash({
+      songs: this.store.findAll('song'),
+      albums: this.store.findAll('album')
+    });
+  }
+});
+```
+
+The controller with the computed property would then be:
```
locks: would then be

locks:
Can you address the comment and then squash the commits?

locks:
:boom: :hammer:

PR #499, opened 2015-07-30
Revert "Get version from url instead of hard-coding for swiftype"

trek:
Reverts emberjs/guides#481

### DIFF
--- a/data/search.yml
+++ b/data/search.yml
@@ -1,1 +1,3 @@
 engine_key: 'ha9YM1kgBdihxrq_8ZHK'
+document_type: 'guides-v1-10-0'
+revision: 'v1.10.0'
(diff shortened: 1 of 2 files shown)


michaelrkn:
I'd suggest we just replace `slugs[slugs.length - 2]` with `slugs[3]`. Getting the version from JavaScript lets us avoid having to update `search.yml` every time we deploy. I want to make it as easy and foolproof as possible to deploy, so we don't end up with broken search like we have on the last two releases, and so if we do need to re-cut a release again (though i hope we won't), it will be easier.

michaelrkn:
Please do not merge until we have decided whether or not to use #513 instead.

michaelrkn:
Closed in favor of #513.

PR #504, opened 2015-07-30
Replace versions with note in versions.json

michaelrkn:
At least a couple people have been confused about why versions.json isn't updated in this repo. This change replaces the dummy version numbers with a note documenting that the file is updated in the deployment process. In development, the user sees this note where the version numbers would be in the drop-down menu.

### DIFF
--- a/source/versions.json
+++ b/source/versions.json
@@ -1,1 +1,1 @@
-["v1.10.0", "v1.9.0"]
+["Versions will appear here when deployed"]


locks:
I don't find the note itself clear, what does it mean?

michaelrkn:
It means that as part of the deploy process, the version numbers are added and will appear in the select menu. Would it be more clear to say "Versions will appear here when deployed"?

locks:
> Would it be more clear to say "Versions will appear here when deployed"?

:+1:

michaelrkn:
Updated!

PR #510, opened 2015-07-30
Removed JSBin from routing/loading-and-error-substates

acorncom:
Meta PR #483 

Probably worth checking me on the naming of this route, I haven't done nested routes like this much ;-)


acorncom:
The above pull request fixes Mamut's second checkbox on the meta PR #483

PR #511, opened 2015-07-31
Fixed typo

mkurian:


michaelrkn:
Heads up - #508 would remove this text anyways, so let's decide whether to merge that first.

locks:
@michaelrkn let's merge yours.

@mkurian thank you for your contribution, sorry about the conflict with #508.

PR #512, opened 2015-07-31
Update handlebars-basics.md wrong link

nicobot:
Link for naming conventions was using a "concepts" prefix which was not leading to the correct page.


michaelrkn:
Heads up - #508 would remove this text anyways, so let's decide whether to merge that first.

locks:
@michaelrkn let's merge yours.

@nicobot  thank you for your contribution, sorry about the conflict with #508.

PR #513, opened 2015-08-02
Search

michaelrkn:
@trek noted that the approach #481 uses will break search if the domain structure of the site changes. Rather than revert to the previous approach of having to update the version number before building (#499), which adds an extra step and thus increases the odds of deploying a build with broken search, this PR simply adds a placeholder string for the version number, which then the `publish-search` task of https://github.com/emberjs/guides.emberjs.com can replace with the version number of the snapshot being deployed (see https://github.com/emberjs/guides.emberjs.com/pull/9).


mixonic:
Being somewhat "ballsy" in merging this. Seems good.

Issue #410, opened 2015-07-09
Merge ember-cli.com & emberjs.com guides


joostdevries:
We can start moving content over from ember-cli.com with the goal to merge the two documentations into one. I've split up ember-cli.com into "user guides" and "extending" so we can first focus on the "user guides". Below are some first thoughts on a toc

I propose to add a section to the guide, right under "getting started" called "Developing your app"
- Configuration
- Development server
- Managing dependencies
- Building your app

In getting started, we should add:
- Upgrading

Finally, at the very bottom of the guide, I think a "Common issues" section with dedicated sub-sections (eg. for CLI) would make sense.

joostdevries:
@locks @trek I think we need to agree on the toc before we/others start creating PR's for this.

joostdevries:
cc @tomdale

eccegordo:
Will ember cli ever have anything that approaches the API docs that ember has? Could be useful for addon developers. Not sure if the ember cli source code is instrumented to support automatic generation (e.g. yuidoc stuff). ping @stefanpenner 

As for TOC I think there also needs to be a dedicated section: 

*Getting Started with Node, NPM and Ember CLI*

Some folks know Ember but are new to the node ecosystem so coming into ember cli will be confusing if they don't have some grounding in how it works. Even installing node can be confusing, which is why I personally recommend NVM.

I started some notes in a gist but happy to beta test and contribute to a guide if someone has PR in process for this.

https://gist.github.com/eccegordo/ad908734e7798bf6fa86

joostdevries:
@eccegordo Quite a few modules Ember CLI are actually documented and there is a script to build those docs, I believe we just never got around to actually publishing them. https://github.com/ember-cli/ember-cli/issues/2764

@stefanpenner I think the main issue with the Ember CLI API's is that we might not have all `@private` markers in place and we **don't** want people using private API's in Addons.

locks:
@eccegordo I would rather make the assumptions explicit at the head of the guides and link appropriately to relevant documentation elsewhere, for example: https://docs.npmjs.com/, http://jsmodules.io/, https://babeljs.io/docs/learn-es2015/, https://leanpub.com/exploring-es6/read. I believe "How Ember CLI works with Node and NPM ecosystem" is/will be explained along the guides as we explain how to use it.

@joostdevries mark everything `@private` like Ember.js did and then review as necessary.

joostdevries:
[...]

PR #515, opened 2015-08-02
Added section on proxying to `Connecting to an http server`

joostdevries:
#410

### DIFF
--- a/source/testing/testing-models.md
+++ b/source/testing/testing-models.md
@@ -1,10 +1,10 @@
-_Unit testing methods and computed properties follows previous patterns shown 
+_Unit testing methods and computed properties follows previous patterns shown
 in [Unit Testing Basics] because DS.Model extends Ember.Object._
 
 [Ember Data] Models can be tested using the `moduleForModel` helper.
 
-Let's assume we have a `Player` model that has `level` and `levelName` 
-attributes. We want to call `levelUp()` to increment the `level` and assign a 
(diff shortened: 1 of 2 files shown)


joostdevries:
I've gone over this content again and the section on mocking is a bit of a train wreck IMO. Furthermore, if we want to encourage a happy path I'd say that an approach with pretender/mirage is favorable.

source/testing/testing-models.md
```diff
@@ -83,6 +83,23 @@ relationships, so you probably don't need to duplicate those tests.  You could
...
+
+```package.json
+{
+  "APIMethod": "proxy",
+  "proxyURL": "http://apiserver.dev:3000",
+  "proxyPath": "/api"
+}
+```
+
+This proxies all requests for a URL starting with `proxyPath` to the proxy server.
```
michaelrkn: I think it would be a bit clearer to say "This proxies all requests for a URL starting with `/api` to the proxy server."

michaelrkn: Also, doesn't this content apply to the development server as well? I'm not sure testing is the right place to introduce this - maybe this text should go in the Models section somewhere, and then be referenced here?

joostdevries:
moved to "Connecting to an http server"

michaelrkn:
Sweet! Now can you squash the commits and I'll get this merged?

joostdevries:
:+1:

PR #517, opened 2015-08-02
Add "Building for production" page to "Getting started"

joostdevries:
#410

### DIFF
--- a/source/getting-started/building-for-production.md
+++ b/source/getting-started/building-for-production.md
@@ -1,0 +1,341 @@
+Ember can automatically build assets for production environments. In these cases,
+assets will be minified, versioned and the production versions of libraries will be
+used.
+
+### Raw Assets
+
+* `public/assets` vs `app/styles`
+
+To add images, fonts, or other assets, place them in the `public/assets` directory. For
+example, if you place `logo.png` in `public/assets/images`, you can reference it in
+templates with `assets/images/logo.png` or in stylesheets with
+`url('/assets/images/logo.png')`.
+
+### Minifying
+
+The compiled css-files are minified by `broccoli-clean-css` or `broccoli-csso`,
+if it is installed locally. You can pass minifer-specific options to them using
+the `minifyCSS:options` object in your Brocfile. Minification is enabled by
+default in the production-env and can be disabled using the `minifyCSS:enabled`
+switch.
+
+Similarly, the js-files are minified with `broccoli-uglify-js` in the
+production-env by default. You can pass custom options to the minifier via the
+`minifyJS:options` object in your Brocfile. To enable or disable JS minification
+you may supply a boolean value for `minifyJS:enabled`.
+
+For example, to disable minifying of CSS and JS, add in `Brocfile.js`:
+```Brocfile.js
+minifyCSS: {
+  enabled: false
+},
+minifyJS: {
+  enabled: false
+}
+```
+
+### Source Maps
+
+Ember CLI supports producing source maps from your CSS and JS files.
+Source maps are configured by the `sourcemaps` option and
+are disabled in production by default.
+
+In dev, the default setting is equal to:
+
+```Brocfile.js
+sourcemaps: {
+  enabled: true,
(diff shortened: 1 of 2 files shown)


michaelrkn:
This feels like a lot of stuff for one page. I'd suggest we keep it simple in the Getting Started section (see #528), and then break this stuff up into smaller pieces.

source/getting-started/building-for-production.md
```diff
@@ -0,0 +1,341 @@
...
+  init: (userId) ->
+    @set 'user', User.findById(userId)
+
+`export default Post`
+```
+
+Note that earlier versions of the transpiler had explicit support for
+CoffeeScript, but that support has been removed.
+
+### EmberScript
```
locks: Remove EmberScript

source/getting-started/building-for-production.md
```diff
@@ -0,0 +1,341 @@
...
+first add [broccoli-ember-script](https://github.com/aradabaugh/broccoli-ember-script) to your
+NPM modules:
+
+```bash
+npm install broccoli-ember-script --save-dev
+```
+
+Note that the ES6 module transpiler is not directly supported with Emberscript, to allow use of ES6 modules use the `` ` `` character to escape raw Javascript similar to the CoffeeScript example above.
+
+### Emblem
```
locks: Remove Emblem

source/getting-started/building-for-production.md
```diff
@@ -0,0 +1,341 @@
...
+
+For [Emblem](http://emblemjs.com/), run the following commands:
+
+```bash
+ember install ember-cli-emblem-hbs-printer
+```
+
+If you're using the older broccoli-emblem-compiler addon, you need to switch over the above ember-cli-emblem-hbs-printer. The older broccoli-emblem-compiler compiles directly to JS instead of Handlebars and therefore is broken on all newer version of HTMLBars.
+
+### Fingerprinting and CDN URLs
```
locks: Separate page

source/getting-started/building-for-production.md
```diff
@@ -0,0 +1,341 @@
...
+  fingerprint: {
+    enabled: false
+  }
+});
+```
+
+Or remove the entry from your `EmberApp` and  `broccoli-asset-rev`
+from your `package.json`.
+
+### Application Configuration
```
locks: Separate Page

source/getting-started/building-for-production.md
```diff
@@ -0,0 +1,341 @@
...
+sourcemaps: {
+  enabled: true,
+  extensions: ['js']
+}
+```
+
+Pass `{enabled: true}` to your EmberApp constructor to enable source maps for javascript.
+Use the `extensions` option to configure CSS.
+
+### Stylesheets
```
locks: Separate page

locks:
We should also be careful about what is linked from the page. For example, Emblem and EmberScript have to play catch up since it's not a core project, so their "updateness" varies.

Maybe link to the category in emberobserver.com instead?

source/getting-started/building-for-production.md
```diff
@@ -0,0 +1,341 @@
+Ember can automatically build assets for production environments. In these cases,
+assets will be minified, versioned and the production versions of libraries will be
+used.
+
+### Raw Assets
```
michaelrkn: Maybe this should be moved to the Managing Dependencies page?

source/getting-started/building-for-production.md
michaelrkn: I think lines 55-69 should be moved to the Managing Dependencies page (with slight modifications to make things flow properly).

source/getting-started/building-for-production.md
michaelrkn: I think all this config stuff should be broken into logically separate pages and put under the Configuring Ember.js section.

michaelrkn:
@joostdevries, can you extract the parts about configuration options and put them under the "Configuring Ember.js" section? Thanks again for all your work putting this together!

PR #522, opened 2015-08-02
Add inline generate command to "The REST Adapter"

joostdevries:
Also added a small subtitle
Refs #410

### DIFF
--- a/source/models/the-rest-adapter.md
+++ b/source/models/the-rest-adapter.md
@@ -5,6 +5,13 @@ associated with each model are conventional; this means that, if you
 follow the rules, you will not need to configure the adapter or write
 any code in order to get started.
 
+If you do need to customize the adapter for your application, you can create
+an Application Adapter which subclasses [DS.RESTAdapter](http://emberjs.com/api/data/classes/DS.RESTAdapter.html) by running:
+
+```bash
+ember generate adapter application
+```
+
 ### URL Conventions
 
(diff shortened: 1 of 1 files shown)


source/models/the-rest-adapter.md
```diff
@@ -5,6 +5,15 @@ associated with each model are conventional; this means that, if you
 follow the rules, you will not need to configure the adapter or write
 any code in order to get started.
 
+#### Customizing the adapter
```
michaelrkn: Let's take away the heading here - it unnecessarily breaks up the page flow, and it doesn't start something that could be considered a new section.

source/models/the-rest-adapter.md
```diff
@@ -5,6 +5,15 @@ associated with each model are conventional; this means that, if you
...
+
+If you do need to customize the adapter for your application, you can create
+an Application Adapter which subclasses [DS.RESTAdapter](http://emberjs.com/api/data/classes/DS.RESTAdapter.html) by running:
```
locks: I thought the default in 1.13 was JSONAPIAdapter?

michaelrkn:
Hey @joostdevries - just want to ping you on these couple comments.

source/models/the-rest-adapter.md
joostdevries: then this entire section needs to be fixed+renamed right?

joostdevries:
fixed the header.

Issue #442, opened 2015-07-20
Add a guide for Services


mattw59:
This should include roughly the following: 
- [ ] What is a service
- [ ] When to use a service, and when not to use one
- [ ] Discuss factories
- [ ] Using initializers with services
- [ ] Using Ember.inject.service()
- [ ] Common types of services (session, user, persistent state rather than temporary state)
- [ ] Fold in existing guide on dependency injection: http://guides.emberjs.com/v1.12.0/understanding-ember/dependency-injection-and-service-lookup/
- [ ] Caution: don't overuse services, this can lead to monoliths of brittle and interdependent code

pixelhandler:
@mattw59 how about adding "How to inject a service into another object", here is an example I've used: https://github.com/pixelhandler/ember-jsonapi-resources/blob/master/tests/dummy/app/models/post.js#L7 sometimes its great to use `application.inject()` with an initializer https://github.com/pixelhandler/ember-jsonapi-resources/blob/master/tests/dummy/app/initializers/post.js#L20 and other times it works to use `Ember.inject.service()` when defining a prototype.

mattw59:
@pixelhandler Sounds good, added a task for this.

michaelrkn:
@mattw59 - We have a goal of getting the 2.0 guides together by Monday, August 3. Do you think you'll be able to have this section ready by then?

mattw59:
@michaelrkn Yes, I'll be sure that something is ready for the team to review by then, preferably before since I'll be on a family vacation Aug 2-4.

michaelrkn:
Awesome!

Poorchop:
I think utils should be mentioned as well. I know that services and utils are two different things, but there are instances where they can both provide a similar means to an end. Utils should probably get their own section so as to avoid confusion, but they could be a subsection of services at the very least.

elbeezy:
@poorchop I disagree that utils should be included in the services section, even in the interim. They belong with helpers IMO as general tooling.

Poorchop:
Yeah that makes sense. I just want to ensure that some mention is made of them because I think that they are a pretty important piece of the framework.

locks:
Isn't a "util" just a module?

MiguelMadero:
@locks I think there's some confusion about when to simply use a module or use services. This is something I've had to clarify multiple times. I think it's important to clarify in the "When not to use services" section that a good alternative is often a module.

[...]

Issue #336, opened 2015-06-05
Add section about initializers


wvteijlingen:
With the advent of instance initializers, I was surprised that there was no mention of initializers in the main guides. The only thing I could find was a small section in the API documentation for Ember.Application.

Maybe adding a section about this to the guides would be nice.

michaelrkn:
@locks what do you think?

locks:
@michaelrkn I think we should document the why in the guides, especially with application/instance initializers, and then punt most of the how to the API docs (like `before`, `after`)

michaelrkn:
It looks like this will be part of #442 sooner or later, so I'm going to close it out now.

PR #532, opened 2015-08-03
Add guide for `Ember.Service`

elbeezy:
With contributions from @mattw59, @SaladFork, @ErikCH

Addresses #442.

Note: This replaces the old DI guide, which may deserve its own (new) section distinct from Services.


source/services/index.md
```diff
@@ -0,0 +1,105 @@
...
+});
+```
+
+### Accessing services
+
+To access a service, inject it with `Ember.inject`:
+
+```app/components/cart-contents.js
+export default Ember.Component.extend({
+  shoppingCart: Ember.inject.service()
```
trek: Can we teach the maximal form first `shoppingCart: Ember.inject.service('shopping-cart')` before showing the shorthand?

trek:
I'm :+1:

source/services/index.md
```diff
@@ -0,0 +1,105 @@
+An `Ember.Service` is a long-lived `Ember.Object` available application-wide. Services are singletons, and only one instance of a given service can exist at a time.
```
locks: what does "application-wide" mean?

locks: the "and" reads weird, since you're giving the definition of what a singleton is.

source/services/index.md
```diff
@@ -0,0 +1,105 @@
...
+
+* Logging
+* User/session authentication
+* Geolocation
+* Third-party APIs
+* Web Sockets
+* Server-sent events or notifications
+* Server-backed API calls that may not fit Ember Data
+
+Because `Ember.Service` doesn't add any sugar to `Ember.Object`, services are not required to follow as many conventions as components or routes. Anything that can be accomplished with an `Ember.Object` can be stored in a service.
```
locks: reader might not know what sugar is

locks: so does it need to follow conventions? which?

locks: "that can be accomplished" -> "can be stored" change of action is confusing

source/services/index.md
```diff
@@ -0,0 +1,105 @@
...
+### Defining Services
+
+To define a service, extend the `Ember.Service` base class:
+
+```app/services/shopping-cart.js
+export default Ember.Service.extend({
+});
+```
+
+Like any `Ember.Object`, a service can have properties and methods of its own.
```
locks: I wonder if it might be worth mentioning CPs just in case? are they introduced by this point in the guides?

source/services/index.md
```diff
@@ -0,0 +1,105 @@
...
+</ul>
+```
+
+The injected property is lazy; the service will not be instantiated until the property is explicitly called. It will then persist until the application exits.
+
+If no argument is provided to `service()`, Ember will use the dasherized version of the property name:
+
+```app/components/cart-contents.js
+export default Ember.Component.extend({
+  shoppingCart: Ember.inject.service()
```
locks: including a comment "this will inject `service:shopping-cart` might be a good idea, you'd be surprised how much we miss things

source/services/index.md
```diff
@@ -0,0 +1,105 @@
...
+});
+```
+
+### Injecting Services Eagerly
```
locks: I'd like to see some discussion on the inclusion of this section. cc @michaelrkn

source/services/index.md
SaladFork: Definitely agree it is worth having `carts.` used somewhere in the JS. Even if it's a built-in CPM. @elbeezy and I discussed this at length and couldn't come up with an example that was simple and idiomatic (things like `isEmpty` belong in the service, not the component). Suggestions welcome.

source/services/index.md
SaladFork: Done.

SaladFork:
Diff was marked as outdated so repeating here as it warrants discussion:
> I'd like to see some discussion on the inclusion of[the Injecting Services Eagerly] section.
-- @locks

source/services/index.md
```diff
@@ -0,0 +1,109 @@
+An `Ember.Service` is a long-lived `Ember.Object` available application-wide. Services are singletons; only one instance of a given service can exist at a time.
+
+Other example uses of services include:
```
trek: "Other" implies an antecedent example, but none exists.

source/services/index.md
```diff
@@ -0,0 +1,109 @@
+An `Ember.Service` is a long-lived `Ember.Object` available application-wide. Services are singletons; only one instance of a given service can exist at a time.
```
trek: "An `Ember.Service` is a long-lived `Ember.Object` available application-wide." is probably still too jargony. It's _true_ but I don't know that it's helpful for someone who doesn't already know this.

michaelrkn:
I wonder if we should add another page to this section about initializers, since they're an important part of the Services story and are referenced in this page. #336 has already noted that we should cover them somewhere in the Guides, so maybe this is the right time and place.

Fryie:
@SaladFork and I were having a discussion about this on Slack. Sorry if I have not been following the discussion closely enough, but my idea was this: Saying ```Ember.Service``` objects *are* singletons is not technically true, even though you would want to use them that way. And I think saying stuff that isn't true is going to confuse some developers. Maybe not the programming novices, but certainly those who aren't new to programming or even JS, just Ember.

Basically this way, at some point your run-of-the-mill OO developer will think "but this is just a class, how come it's singleton? I don't believe it. I'll just Service.create() it. Wow, that worked. The guides must be wrong!"?. At least I know I would.

At the same time, I understand not wanting to discuss / mention the container. But there needs to be some understanding that there's some lifecycle management that Ember does for you that you want to use, that services are meant to be singletons and that if you use them "correctly", they are. But that you can still break out of it if you really, really need to (maybe in tests or something).

So how about something like:
"Ember will generate singleton instances of any services automatically, so that you can access their state across the whole application.".?
That's not saying anything that's technically incorrect, but neither does it go too deep into implementation details. (I'm just not too happy about that "automatically" part, because you still need to inject them somewhere.)

source/services/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+```app/components/cart-contents.js
+export default Ember.Component.extend({
+  shoppingCart: Ember.inject.service()
+});
+```
+
+This also injects the shopping cart service, as the `shoppingCart` property.
+
+### Injecting Services Eagerly
```
Fryie: I think there are two different concepts mentioned here: 1. availability without the need to declare it on the object, 2. eagerness of initialization (i.e. not on-demand). I think those two should be discussed separately. Like, "when you do this, Ember will do two things: blahblah"

locks:
@trek commit was amended with changes

elbeezy:
It seems like the major sticking point is how to handle discussion (if any) of initializers, and if so, whether it should be in this section or elsewhere. Let's try to get at least an initial consensus so we can get something into 2.0, even if we keep discussing and decide to move things around for 2.1?

michaelrkn:
Thanks for all your hard work on this!

PR #535, opened 2015-08-03
Avoid JSHint unused error when importing helpers

adirhaleli:
Currently the documentation suggests that you need to import your custom test helper in `start-app.js`.
While this is true, the documented import syntax will cause a failing JSHint test on `start-app.js` because the imported test helper is not directly begin used.
I think the documentation should suggest to import using `import "./helper"` instead of `import "helper" from "./helper"`.

### DIFF
--- a/source/testing/test-helpers.md
+++ b/source/testing/test-helpers.md
@@ -198,8 +198,8 @@ import Ember from 'ember';
 import Application from '../../app';
 import Router from '../../router';
 import config from '../../config/environment';
-import shouldHaveElementWithCount from "./should-have-element-with-count";
-import dblclick from "./dblclick";
-import addContact from "./add-contact";
+import "./should-have-element-with-count";
+import "./dblclick";
+import "./add-contact";
 ```
 


locks:
That is not valid ES2015 syntax for modules @adirhaleli, so the guides should be correct. Do your changes work in your application?

rwjblue:
> That is not valid ES2015 syntax for modules @adirhaleli

Yes, I believe it is valid.  

* Using `import "./foo"`: https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&playground=false&code=import%20%22.%2Ffoo%22%3B
* Using `import from "./foo";`: https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=true&spec=false&playground=false&code=import%20from%20%22.%2Ffoo%22%3B
* https://leanpub.com/understandinges6/read#leanpub-auto-importing-without-bindings

locks:
:o

adirhaleli:
@locks In my application everything seems to work as expected and like I've mentioned, using the classic binding import form does leads to a failing JSHint test.

PR #536, opened 2015-08-03
Update examples to 1.13.x standards

typeoneerror:
* No Ember.View
* No Ember.keys

Example more es6-y.

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -79,23 +79,19 @@ Now the same handlebars code above renders the following HTML:
        type="text" data-toggle="tooltip" data-placement="bottom" title="Name">
 ```
 
-You can also automatically bind data attributes on the base view with the
-following:
+You can also automatically bind data and aria attributes on the base component with the following:
 
 ```javascript
-export default Ember.View.reopen({
-  init() {
-    this._super();
-    var self = this;
-
-    // bind attributes beginning with 'data-'
-    Ember.keys(this).forEach(function(key) {
-      if (key.substr(0, 5) === 'data-') {
-        self.get('attributeBindings').pushObject(key);
+export default Ember.Component.reopen({
+  bindAttributes: Ember.on('init', function() {
+    Object.keys(this).forEach((key) => {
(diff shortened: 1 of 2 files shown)


source/templates/binding-element-attributes.md
```diff
@@ -79,23 +79,19 @@ Now the same handlebars code above renders the following HTML:
...
-    this._super();
-    var self = this;
-
-    // bind attributes beginning with 'data-'
-    Ember.keys(this).forEach(function(key) {
-      if (key.substr(0, 5) === 'data-') {
-        self.get('attributeBindings').pushObject(key);
+export default Ember.Component.reopen({
+  bindAttributes: Ember.on('init', function() {
+    Object.keys(this).forEach((key) => {
```
locks: unsure about `Object.keys`

locks:
I could swear a previous PR removed all mentions of `Ember.View`. Maybe because the PR is pointing to the `v1.13.0` branch, not master. Can you update?

source/templates/binding-element-attributes.md
typeoneerror: I am too. Which seems like that might be a problem. I get a deprecation warning if I use Ember.keys with 1.13.x, and others are removing it (https://github.com/switchfly/ember-test-helpers/pull/71), so seemed right? Unsure.

typeoneerror:
Still there in master: 

* https://github.com/emberjs/guides/blob/master/source/object-model/bindings.md.
* https://github.com/emberjs/guides/blob/master/source/templates/binding-element-attributes.md

I'll move the PR to master.

source/templates/binding-element-attributes.md
locks: @typeoneerror according to the documentation "On browsers that support it, this uses the native Object.keys implementation." so I think you are right. I thought maybe it did something with computed properties or the like.

typeoneerror: @locks the open PR has been reverted back to Ember.keys. see: https://twitter.com/trek/status/628305995589152768.

locks: :+1: thank you for the clarification

PR #540, opened 2015-08-04
Remove top part of router.js examples

michaelrkn:
They may cause confusion because they omit the configuration options provided by Ember CLI.

### DIFF
--- a/source/controllers/dependencies-between-controllers.md
+++ b/source/controllers/dependencies-between-controllers.md
@@ -3,15 +3,11 @@ to have some kind of connection between two controllers. Let's take this
 router as an example:
 
 ```app/router.js
-var Router = Ember.Router.extend({});
-
 Router.map(function() {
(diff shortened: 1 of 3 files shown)


locks:
@michaelrkn I say remove the export as well, like the previous snippets.

I'm a bit weary of not adding a note explaining the full file thingie since people complain about partial snippets, but we can discuss that separately.

michaelrkn:
Ooh yes good point. Removed `export default`. I agree that I'm a little hesitant about not having the full file, but I think in this case it will be clear enough, at least for folks using Ember CLI (which we're assuming now in the Guides.)

michaelrkn:
I just updated this after searching through the entire project to make sure it was consistent everywhere.

michaelrkn:
@locks can you re-review?

locks:
:shipit:

PR #541, opened 2015-08-04
Rewrite "Core concepts"

michaelrkn:
I've cut out the long introduction so that the reader can get to the point. I've included many more examples to better illustrate the concepts. And I've reorganized the material to minimize the number of times a concept is referenced before it is introduced.

### DIFF
--- a/source/getting-started/core-concepts.md
+++ b/source/getting-started/core-concepts.md
@@ -1,82 +1,58 @@
 To get started with Ember.js, there are a few core concepts you
 should understand.
 
-Ember.js is designed to help developers build ambitiously large web
-applications that are competitive with native apps. Doing so requires
-both new tools and a new vocabulary of concepts. We've spent a lot of
-time borrowing ideas pioneered by native application frameworks like
-Cocoa and Smalltalk.
+## Templates
 
-However, it's important to remember what makes the web special. Many
-people think that something is a web application because it uses
-technologies like HTML, CSS and JavaScript. In reality, these are just
-implementation details.
+Templates, written in the Handlebars language, describe the user interface of
+your application. In addition to plain HTML, templates can contain expressions,
+like `{{title}}` or `{{author}}`, which take information from a component or
+controller and put it into HTML. They can also contain helpers, like
+`{{if isAdmin}}30 people have viewed your blog today.{{/if}}.` Finally, they can
+contain components, such as if a template listing out all the blog posts in an
+app rendered a component for each post.
 
-Instead, **the web derives its power from the ability to bookmark and
-share URLs.** URLs are the key feature that give web applications
-superior shareability and collaboration. Today, most JavaScript
-frameworks treat the URL as an afterthought, instead of the primary
-reason for the web's success.
+## Components
 
-Ember.js, therefore, marries the tools and concepts of native
-GUI frameworks with support for the feature that makes the web so
-powerful: the URL.
+Components are the primary way user interfaces are organized in Ember. They
+consist of two parts: a template, and a source file written in JavaScript that
+defines the component's behavior. For example, a blog application might have a
+component for displaying a list of blog posts called `all-posts`, and another
+component for displaying an individual post called `view-post`. If users can
+upvote a post, the `view-post` component might define a behavior like _when the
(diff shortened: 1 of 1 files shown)


source/getting-started/core-concepts.md
```diff
@@ -1,82 +1,59 @@
...
-parts of your app.
-
-Now that you understand the roles of these objects, you're equipped to
-dive deep into Ember.js and learn the details of how each of these
-individual pieces work.
+## Templates
+
+Templates, written in the Handlebars language, describe the user interface of
+your application. In addition to plain HTML, templates can contain expressions,
+like `{{title}}` or `{{author}}`, which take information from a component and
```
locks: "from a component"?

source/getting-started/core-concepts.md
```diff
@@ -1,82 +1,59 @@
...
+put it into HTML. They can also contain helpers, like `{{if isAdmin}}30 people
+have viewed your blog today.{{/if}}.` Finally, they can contain components, such
+as if a template listing out all the blog posts in an app rendered a component
+for each post.
+
+## Components
+
+Components are the primary way user interfaces are organized in Ember. They
+consist of two parts: a template, and JavaScript that defines the component's
```
locks: "and a source file written in JavaScript" just to hammer it home, maybe?

source/getting-started/core-concepts.md
```diff
@@ -1,82 +1,59 @@
...
+behavior. For example, a blog application might have a component for displaying
+a list of blog posts called `all-posts`, and another component for displaying an
+individual post called `view-post`. If users can upvote a post, the `view-post`
+component might define a behavior like _when the user clicks the upvote button,
+increase the `vote` property's value by 1`_.
+
+## Controllers
```
locks: I both like and dislike the direction you went with in this paragraph, not sure how to be more helpful with my feedback.

Are you sure this is more enlightening than confusing?

source/getting-started/core-concepts.md
```diff
@@ -1,82 +1,59 @@
...
+
+Controllers are very much like components, so much so that in future versions of
+Ember, controllers will be replaced entirely with components. At the moment,
+components cannot be routed to (see below), but when this changes, it will be
+recommended to replace all controllers with components.
+
+## Models
+
+Models stores _persistent state_. For example, a blog application would want to
```
locks: `Models store`

source/getting-started/core-concepts.md
```diff
@@ -1,82 +1,59 @@
...
+save the content of a blog post when a user publishes it, and so the blog post
+would have a model defining it, perhaps called the `Post` model. A model
```
locks: I suggest a slight rewording in the spirit of "persists information to a server using the JSON API format" with a link to it.

source/getting-started/core-concepts.md
michaelrkn: Maybe "from a component or controller"?

source/getting-started/core-concepts.md
michaelrkn: That sounds good to me. I wish we could just call it a "component controller", but that would be confusing, I guess. There really should be a name for that "source file written in JavaScript".

source/getting-started/core-concepts.md
michaelrkn: I like that wording. I'm not sure about including a link, though, because this is meant to be the 30-second overview. They'll get to the full explanation of all these concepts in the relevant sections.

Fryie: Hmm, I think "Models *represent* persistent state" might be a better wording, because you can have in-memory models which you don't store (because they don't validate or something) and also because the responsibility of actually storing stuff on the server lies with adapters and serializers.

Although I can see that you don't want to throw too many concepts around, initially. But "represents" sounds more accurate to me, and the later sentence that models typically persist to a JSON API backend should be fine in this context without the need to explicitly mention adapters etc.

michaelrkn: @Fryie I like it!

michaelrkn:
@locks can you re-review?

PR #544, opened 2015-08-05
Update test-helpers.md regarding the fillIn() helper

jdhines:
When filling in <select> elements, you need to give the value of the option for the text argument of fillIn(), as giving the display text (innerHTML) won't work.

### DIFF
--- a/source/testing/test-helpers.md
+++ b/source/testing/test-helpers.md
@@ -37,7 +37,7 @@ next one starts.
     is complete.
 * `fillIn(selector, text)`
   - Fills in the selected input with the given text and returns a promise that
-     fulfills when all resulting async behavior is complete. Works with `<select>` elements as well as `<input>` elements.
+     fulfills when all resulting async behavior is complete. Works with `<select>` elements as well as `<input>` elements (with select elements, fill in the value of the option, e.g. 'true' and not the display, e.g. 'Yes').
 * `keyEvent(selector, type, keyCode)`
   - Simulates a key event type, e.g. `keypress`, `keydown`, `keyup` with the
     desired keyCode on element found by the selector.


source/testing/test-helpers.md
```diff
@@ -37,7 +37,7 @@ next one starts.
     is complete.
 * `fillIn(selector, text)`
   - Fills in the selected input with the given text and returns a promise that
-     fulfills when all resulting async behavior is complete. Works with `<select>` elements as well as `<input>` elements.
+     fulfills when all resulting async behavior is complete. Works with `<select>` elements as well as `<input>` elements (with select elements, fill in the value of the option, e.g. 'true' and not the display, e.g. 'Yes').
```
rwjblue: This is good information, but perhaps we can reword it to make select less of a special case.  In both scenarios you are setting the `value` of the element, maybe we can just say that instead of calling out `<select>`?

jdhines: Well with input they're the same, right, so that's just straightforward. I just recently noticed this "gotcha" though with selects since I was trying to fill in with the value the user would see, and couldn't understand why it wasn't working all of a sudden, since my others had worked fine (the selects where the value and display were the same).

michaelrkn: @rwjblue with with `<input>` tags, the value and the display are the same, so it seems necessary to me to state the special case of `<select>`.

rwjblue: @michaelrkn - How is it a special case?  In both case you are setting the value...

michaelrkn: Sorry, poorly said. What I'm trying to say is that for `<input>`, there is no difference between the value and what's displayed on the screen, so it goes without saying that you'd use the value. With `<select>`, there is a difference, and so it is helpful to mention that you should use the value.

If we just said "Works with `<select>` elements as well as `<input>` elements (use the value of the option, e.g. 'true' and not the display, e.g. 'Yes').", the parenthetical comment doesn't make much sense in the context of the `<input>` tag. I'm struggling to come up with a general-case phrasing that doesn't feel confusing for `<input>`. Maybe you can suggest something?

rwjblue: Maybe describing how it works for each element type supported (instead of a parenthetical)?

locks:
@jdhines @rwjblue @michaelrkn what's the status here?

michaelrkn:
Closing in favor of #678.

PR #545, opened 2015-08-05
Fix formatting of `#each key=` special value list

mattmcmanus:
Noticed this was off when trying to make sense of the new #each abilities.

### DIFF
--- a/source/templates/displaying-a-list-of-items.md
+++ b/source/templates/displaying-a-list-of-items.md
@@ -2,7 +2,7 @@ If you need to enumerate over a list of objects, use Handlebars' `{{#each}}` hel
 
 ```handlebars
 <ul>
-  {{#each people key="id" as |person|}}
+  {{#each people as |person|}}
     <li>Hello, {{person.name}}!</li>
   {{/each}}
 </ul>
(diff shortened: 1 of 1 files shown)


source/templates/displaying-a-list-of-items.md
```diff
@@ -44,10 +44,11 @@ and related [Ember Mutable Array methods](http://emberjs.com/api/classes/Ember.M
...
   
   There are a few special values for `key`:
-    * `@index` - The index of the item in the array.
-    * `@item` - The item in the array itself.  This can only be used for arrays of strings
+  
+  * `@index` - The index of the item in the array.
+  * `@item` - The item in the array itself.  This can only be used for arrays of strings
       or numbers.
-    * `@guid` - Generate a unique identifier for each object (uses `Ember.guidFor`).
+  * `@guid` - Generate a unique identifier for each object (uses `Ember.guidFor`).
```
rwjblue: This option is no longer present in 2.0.0-beta's.

source/templates/displaying-a-list-of-items.md
```diff
@@ -44,10 +44,11 @@ and related [Ember Mutable Array methods](http://emberjs.com/api/classes/Ember.M
   property. Make sure the value you pass to `key` is unique!
   
   There are a few special values for `key`:
-    * `@index` - The index of the item in the array.
-    * `@item` - The item in the array itself.  This can only be used for arrays of strings
+  
+  * `@index` - The index of the item in the array.
+  * `@item` - The item in the array itself.  This can only be used for arrays of strings
```
rwjblue: This option is no longer present in 2.0.0-beta's.

rwjblue:
Can you remove `@guid` and `@item`?  Basically, the only special value that works in beta is `@index`, but it is generally unusual in 1.13.6+ to need to specify any key value (an `{{each` without key should "just do the right thing" now).

mattmcmanus:
Oh right, I just noticed the different branches as well. Didn't realize master was for 2.0. I'll update this PR to fix the 1.13 docs and then one for master. 

Is `#each` with index (`{{#each people as |people index|}}`) still in 2.0?

mattmcmanus:
Rereading the section, why not just remove that whole paragraph in favor of explaining the with index format introduced in `1.11`?

locks:
@mattmcmanus yes, I believe index as a block param will still be available.
As for `key=@index` and the index block param, they are different things I believe. The first tells Ember what property to use to trigger state changes, the other is merely the index of the element in the array.

michaelrkn:
@mattmcmanus We are cutting a release today. Will you have time to get to this in the next few hours? If not, no worries - I'll get this finished in a bit myself.

mattmcmanus:
Hey @michaelrkn, I'll take a stab right now.

mattmcmanus:
Alrighty! If you give it a once over for spelling/grammer/correctness. I think it's an improvement.

rwjblue:
Thank you, that definitely reads much better to me. Can you squash commits?

mattmcmanus:
Squashed.

locks:
:boom: :hammer:

PR #546, opened 2015-08-05
Formatting fix in 1.13 docs for #each with key

mattmcmanus:
This one should be to the right branch

### DIFF
--- a/source/templates/displaying-a-list-of-items.md
+++ b/source/templates/displaying-a-list-of-items.md
@@ -43,10 +43,10 @@ and related [Ember Mutable Array methods](http://emberjs.com/api/classes/Ember.M
   property. Make sure the value you pass to `key` is unique!
 
   There are a few special values for `key`:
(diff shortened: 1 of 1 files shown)


locks:
@mattmcmanus We don't really accept PRs to the version branches since we don't currently redeploy them. While 1.13 will be a special case, I'm not sure this helps with out workflow.
@michaelrkn what do we do about this? related: https://github.com/emberjs/guides/pull/545 Should @mattmcmanus make the PR against master instead then we merge it before #545?

Sorry if I'm overcomplicating.

michaelrkn:
If this feature isn't available in 2.0, I don't think we want to encourage people to use it in 1.13.

mattmcmanus:
Wasn't aware of that. Sorry. Just saw an easy fix and jumped on it.

michaelrkn:
@mattmcmanus no worries, thanks for contributing!

PR #547, opened 2015-08-05
Fix broken link to Disabling Prototype Extensions page.

BaseCase:
A link on the object-model/enumerables page was broken.


michaelrkn:
Thanks!

PR #422, opened 2015-07-11
Added a link to a recent video covering emberjs testing strategies at…

grandschemathings:
… Yahoo

PR #549, opened 2015-08-05
Remove jsbin links at the bottom of query params

michaelrkn:
I think in the same spirit of the discussion on #422, we should remove these links. Let's not merge until the discussion about #422 has finished.


locks:
:fire:

PR #551, opened 2015-08-06
Name the input field in the docs 'search' in 'Controllers'

jalcine:
I read this portion of the guide about three times before noticing what looked so weird to me. If I've misinterpreted this, can an explanation on why `value` is the correct attribute to use here?


jalcine:
Wait, no, just re-read that, that's the preset value. My bad.

PR #552, opened 2015-08-06
Corrected sessionService example.

lpaluszk:
There is a small mistake in one of the examples:

http://guides.emberjs.com/v1.13.0/understanding-ember/dependency-injection-and-service-lookup/#toc_lightweight-services-with-code-ember-inject-code

      sessionService: Ember.inject.service('session'),
      isLoggedIn: Ember.computed.alias('serviceService.isAuthenticated'),
      actions: {
        signIn() {
          // There is an alias to the session property, so this change propagates
          // to the session object then the IndexController.
          this.set('isLoggedIn', true);
        }
      }
    });

where serviceService.isAuthenticated is called instead of sessionService.

This PR fixes that.


michaelrkn:
Thanks for the contribution! This page is actually going to be removed entirely in the next Guides release and replaced with an updated Services section, so I'm going to close this out. (We don't update content, layout, or styles for older versions of the Guides. They are considered static and immutable, as it is too difficult to maintain content for every version ever released. Issues will only be fixed for future releases.)

Issue #554, opened 2015-08-08
Give more context and background for The Object Model section


sajt:
In the /object-model/classes-and-instances.md the first example is not clear for me.

If I put this:

```javascript
Person = Ember.Object.extend({
  say(thing) {
    alert(thing);
  }
});
```

To my routes js I get the following error:

    routes/posts.js: line 10, col 1, 'Person' is not defined.

but if I use 
```javascript
Ember.Person = Ember.Object.extend({
```

What is the perfect solution or where can I define classes?

michaelrkn:
You're right, this isn't totally clear. I'm going to work on better explaining this section.

Check out http://guides.emberjs.com/v1.13.0/models/defining-models/ for the answer to your question.

sajt:
thanks

Issue #543, opened 2015-08-05
Make it obvious to use KVO-compliant Array methods


xudaiqing:
With help from the forum i finally find out how to use array in ember. 
Most important, how to use addObject and others. And the information of prototype extension of native array.
Maybe it's just because as i have used ember more than a year ago, i didn't read though the guide carefully.
But i believe it shouldn't that hard.
I think use array is a common case in develop a application and guide should cover that clearly. 
Currently I can't find enough information when i first enconter the problem.

http://guides.emberjs.com/v1.13.0/templates/displaying-a-list-of-items/
have cover it a little but [].pushObject seems come out of nowhere and when follow the link to the http://emberjs.com/api/classes/Ember.MutableArray.html 
It only says concrete implementations of this class include ArrayProxy and ArrayController.

michaelrkn:
I agree. We should make it really obvious in http://guides.emberjs.com/v1.13.0/object-model/enumerables/ that you can't use `push`, that you need to use `pushObject` (and the other KVO-compliant methods).

michaelrkn:
Closed by #555.

PR #555, opened 2015-08-08
Improve Object Model section

michaelrkn:
This PR:

* Adds an introduction to the section (fixing #554)
* Makes it more clear when to use CPs, Observers, and Bindings
* Improves the organization of the Enumerable guide
* Emphasizes the need to use observable Enumerable methods (fixing #543)


source/object-model/bindings.md
```diff
@@ -1,8 +1,12 @@
 A binding creates a link between two properties such that when one changes, the
 other one is updated to the new value automatically. Bindings can connect
-properties on the same object, or across two different objects. Unlike most other
-frameworks that include some sort of binding implementation, bindings in
-Ember.js can be used with any object, not just between views and models.
+properties on the same object, or across two different objects.
+
+Unlike most other frameworks that include some sort of binding implementation,
+bindings in Ember.js can be used with any object, not just between views and
```
locks: mentions `views`

source/object-model/classes-and-instances.md
```diff
@@ -17,8 +21,8 @@ You can also create a _subclass_ from any existing class by calling
 its `extend()` method. For example, you might want to create a subclass
 of Ember's built-in `Ember.Component` class:
 
-```app/views/person.js
-PersonView = Ember.Component.extend({
+```app/components/person.js
```
locks: component needs dasherized name

source/object-model/index.md
```diff
@@ -0,0 +1,16 @@
...
+functionality, as well as bring several features like mixins and initialization
+to make working with them a more pleasant experience. Although these features
+aren't available in standard JavaScript, many of them are designed to align with
+proposed additions to the ECMAScript standard.
+
+In addition to bringing its own object model, Ember also extends the built-in
+JavaScript `Array` prototype with its Enumerable interface to enable it to
+observe changes and provide more features.
+
+Finally, Ember extends the `String` prototype with a few [formatting and
```
locks: do we mention somewhere that prototype extensions are disabled in addons? or that's for the ember-cli addons documentation, or?

locks:
Small remarks aside, this looks good.

source/object-model/index.md
HeroicEric: I think the goal is to move away from talking about prototype extensions in an effort to get people to also move away from using them https://github.com/emberjs/guides/pull/457#discussion_r35277298

locks: That's a related but separate issue @HeroicEric, I'm fine with the PR as it is.

HeroicEric: @locks cool. I just figured that since this is new content it, it would be better to not add more references to prototype extensions

source/object-model/bindings.md
michaelrkn: would it be more appropriate to say "between components and models"? or maybe we should just strike this sentence entirely - i'm not sure we need to say how great Ember is compared to other frameworks. thoughts?

source/object-model/index.md
michaelrkn: My takeaway from #110 was that we are discourage use of Function prototype extensions but were okay with Array prototype extensions, at least for now. If I misunderstood or there's been other discussion since, just let me know!

Re: addons, I think that's a separate story, and should go in the guide for authoring addons when it's written.

source/object-model/classes-and-instances.md
michaelrkn: note to myself for revising:

```
```app/components/todo-item.js
  classNameBindings: ['isUrgent'],
  isUrgent: true
});
```

source/object-model/index.md
HeroicEric: @michaelrkn I know the goal to eventually move away from extending `Array` but it looks like there is a good amount of work left (https://github.com/emberjs/ember.js/pull/10899) before that will happen. And I do think that part about the arrays is still useful to know because you're likely to run into them with current state of things, so I'm not totally against that part.

IMO, the part about `String` prototype extensions is less important to tell them about. I think it would be better to encourage using `Ember.String.foo` whenever possible.

We can always open another issue about this to discuss this further without blocking this PR :smile:

michaelrkn:
Just updated this to remove references to Ember.Set, as it is deprecated. Fixes #558.

michaelrkn:
This should be good to go. @locks @HeroicEric any final thoughts?

locks:
:boom: :hammer:

PR #559, opened 2015-08-10
Add Example Uses of `[]` in Computed Properties

elwayman02:
Given the deprecation of `@each` for leaf node observing in computed properties, adds examples to the guides that show users how to utilize the `[]` key for those scenarios
Updates the title of the page to remove specific mention of `@each` to avoid confusion (per @rwjblue)

### DIFF
--- a/source/object-model/computed-properties-and-aggregate-data.md
+++ b/source/object-model/computed-properties-and-aggregate-data.md
@@ -57,3 +57,65 @@ todosController.get('remaining');
 
 Note that `@each` only works one level deep. You cannot use nested forms like
 `todos.@each.owner.name` or `todos.@each.owner.@each.name`.
+
+Sometimes, you may only care if the items of an array have been added, removed, or replaced, 
+rather than changes to properties on the items themselves. For example, items may have been 
+added to or removed from the array. In this case, you can create a computed property as follows:
+
+
+```app/controllers/todos.js
+export default Ember.Controller.extend({
+  todos: [
+    Ember.Object.create({ isDone: true }),
+    Ember.Object.create({ isDone: false }),
+    Ember.Object.create({ isDone: true })
+  ],
+
+  selectedTodo: null,
+  indexOfSelectedTodo: Ember.computed('selectedTodo', 'todos.[]', function () {
+    return this.get('todos').indexOf(this.get('selectedTodo'));
+  })
+});
+```
+
+Note here that we use the special key `[]` instead of `@each.foo` (which is used for observing 
+a specific dependent key, such as `foo`, on each item). This instructs Ember.js to only care about 
+the array itself changing (such as adding/removing items), rather than changes to the items themselves.
+
+Several of the [Ember.computed](http://emberjs.com/api/classes/Ember.computed.html) macros 
+utilize the `[]` key to implement common use-cases as well. For example, if you wanted to 
+create a computed property that mapped properties from an array, you could use 
+[Ember.computed.map](http://emberjs.com/api/classes/Ember.computed.html#method_map)
+or build the computed yourself:
+
+```javascript
+const Hamster = Ember.Object.extend({
+  excitingChores: Ember.computed('chores.[]', function () {
+    return this.get('chores').map(function (chore, index) {
+      return `CHORE ${index}: ${chore.toUpperCase()}!`;
+    }
+  })
+});
+
+const hamster = Hamster.create({
+  chores: ['clean', 'write more unit tests']
+});
+
+hamster.get('excitingChores'); // ['CHORE 1: CLEAN!', 'CHORE 2: WRITE MORE UNIT TESTS!']
+hamster.get('excitingChores').pushObject('review code');
+hamster.get('excitingChores'); // ['CHORE 1: CLEAN!', 'CHORE 2: WRITE MORE UNIT TESTS!', 'CHORE 3: REVIEW CODE!']
+```
+
+By comparison, using the computed macro abstracts some of this away:
+
+```javascript
+const Hamster = Ember.Object.extend({
+  excitingChores: Ember.computed.map('chores', function (chore, index) {
+    return `CHORE ${index}: ${chore.toUpperCase()}!`;
+  })
+});
+```
+
+The computed macros expect you to use an array, so there is no need to use the '[]' key in these cases. 
+However, building your own custom computed property requires you to tell Ember.js that it is watching 
+for array changes, which is where the `[]` key comes in handy.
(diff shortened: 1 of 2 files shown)


source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
 
 Note that `@each` only works one level deep. You cannot use nested forms like
 `todos.@each.owner.name` or `todos.@each.owner.@each.name`.
+
+Sometimes, you may only care if the items of an array have changed, rather than their specific properties.
+For example, items may have been added to or removed from the array. In this case, you can created a
```
rwjblue: s/created/create/

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
...
+  selectedTodo: null,
+  indexOfSelectedTodo: Ember.computed('selectedTodo', 'todos.[]', function () {
+    return this.get('todos').indexOf(this.get('selectedTodo'));
+  })
+});
+```
+
+Note here that we use the special key `[]` instead of `@each`, which is used for observing 
+a specific dependent key on each item. This instructs Ember.js to only care about the array 
+itself having items added, removed, or replaced, rather than changes to the items themselves.
```
rwjblue: "changes to properties on the items" (not sure of this one)?

rwjblue:
I left a few changes/tweaks but looks good to me.

source/object-model/computed-properties-and-aggregate-data.md
elwayman02: whoops

source/object-model/computed-properties-and-aggregate-data.md
elwayman02: :+1:

elwayman02:
@rwjblue updated

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
 
 Note that `@each` only works one level deep. You cannot use nested forms like
 `todos.@each.owner.name` or `todos.@each.owner.@each.name`.
+
+Sometimes, you may only care if the items of an array have changed, rather than their specific properties.
```
locks: "you may only care if the items of an array have changed" is a bit ambigious IMO

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
...
+For example, items may have been added to or removed from the array. In this case, you can create a
+computed property that relies on the leaf node:
```
locks: what's a "leaf node"?

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
...
+  ],
+
+  selectedTodo: null,
+  indexOfSelectedTodo: Ember.computed('selectedTodo', 'todos.[]', function () {
+    return this.get('todos').indexOf(this.get('selectedTodo'));
+  })
+});
+```
+
+Note here that we use the special key `[]` instead of `@each`, which is used for observing
```
locks: `[]` is used for observing a specific dependent key on each item? (aka, can you touch up the phrasing? ;P)

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
...
+```
+
+Note here that we use the special key `[]` instead of `@each`, which is used for observing 
+a specific dependent key on each item. This instructs Ember.js to only care about the array 
+itself having items added, removed, or replaced, rather than changes to properties on the items themselves.
+
+Several of the [Ember.computed](http://emberjs.com/api/classes/Ember.computed.html) macros 
+utilize the `[]` key to implement common use-cases as well. For example, if you wanted to 
+create a computed property that mapped properties from an array, you could use 
+[Ember.computed.map](http://emberjs.com/api/classes/Ember.computed.html#method_map)
```
locks: I would show a snippet using `Ember.computed.map` for comparison

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -57,3 +57,51 @@ todosController.get('remaining');
...
+    }
+  })
+});
+
+const hamster = Hamster.create({
+  chores: ['clean', 'write more unit tests']
+});
+
+hamster.get('excitingChores'); // ['CHORE 1: CLEAN!', 'CHORE 2: WRITE MORE UNIT TESTS!']
+hamster.get('excitingChores').push('review code');
```
locks: I think this should be `hamster.get('chores').pushObject('review code');`

source/object-model/computed-properties-and-aggregate-data.md
elwayman02: haha good catch

source/object-model/computed-properties-and-aggregate-data.md
elwayman02: :+1: thought about it, didn't follow through

source/object-model/computed-properties-and-aggregate-data.md
elwayman02: Chores is just a native array, though...and I'm not pushing an object. Show me `pushObject` in the API docs here: http://emberjs.com/api/classes/Ember.Array.html

source/object-model/computed-properties-and-aggregate-data.md
elwayman02: Will rephrase

source/object-model/computed-properties-and-aggregate-data.md
rwjblue: We augment native arrays with pushObject for observability.

elwayman02: @rwjblue @locks I know that we augment native arrays, but shouldn't `pushObject` be listed in the API docs if that's a method we should be using? I'm loathe to include example code using methods that aren't even documented.

elwayman02:
Updated and squashed again...just need a resolution on the `pushObject` discussion.

rwjblue:
Push object is public API. If it isn't listed in API docs that is a bug.

elwayman02:
@locks resolved it for me.  Updated now :)

locks:
:cookie: for you :D

elwayman02:
I LOVE COOKIES!

PR #561, opened 2015-08-10
Fixed old link in docs

acorncom:
The repo has moved, but the link still works due to a redirect on Github's end


locks:
Given that the adapter is still broken for 1.13, can you remove the link instead?

michaelrkn:
Thank you! Can you squash the two commits, and then we'll get this merged?

acorncom:
@michaelrkn Done.

Issue #568, opened 2015-08-12
Add configuring options


locks:
Add a list of configuring options as per https://github.com/emberjs/guides/pull/524#issuecomment-128547671.

PR #569, opened 2015-08-12
Document Ember CLI configuration options

jkarsrud:
This documents the configuration options a user can override in `.ember-cli`, as requested in #568.

Not really sure if a table is the way to go or not, but it made more sense to me than using comments in json.


locks:
I'd prefer prose, with a short explanation what each setting is for.

PR #572, opened 2015-08-12
Removes prose related to Web Component standard

locks:


locks:
cc @tomdale

Eric-Guo:
@locks Can you explain motivation of removing? does it means Ember.js is now not target to [Web
-Components specification](http://www.w3.org/TR/components-intro/) or just already implement it so no need mention in guide?

locks:
Ember Components are no longer aligned with the specs, especially after
some of the upcoming RFCs.

PR #580, opened 2015-08-13
Remove imports

locks:
There were some imports that used paths allowed by `ember-cli-shims` which aren't final yet.
While I was fixing it I realized the Guides in general omit the imports, so I combed through and removed the ones I could find.

### DIFF
--- a/source/components/composing-components.md
+++ b/source/components/composing-components.md
@@ -202,8 +202,6 @@ We'll be looking at a `{{user-profile}}` component that implements a save action
 Here's our component's definition:
 
 ```app/components/user-profile/component.js
-import Ember from 'ember';
-
 export default Ember.Component.extend({
   actions: {
     saveUser() {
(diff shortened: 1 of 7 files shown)


rwjblue:
The `import Ember from 'ember';` and `import DS from 'ember-data';` modules definitely work (and have for a long time).  Ember CLI defaults to not allowing globals for Ember and DS (via `.jshintrc` config) so we should keep those imports.

locks:
@rwjblue IIRC we decided not to include imports at all, hence the removal to match the rest of the Guides. The expectation is not to rely on globals, but that `ember generate`'s blueprints already include the proper imports.

This might change in the Big Review coming up because there are some issues with the partial code snippets, but we'll update accordingly then.

michaelrkn:
I'm on board with this. Want to give it some time for others to weigh in.

locks:
ping @wycats @tomdale @trek

tomdale:
I'm okay with removing `import`s that require `ember-cli-shims` because folks can just assume `import Ember from "ember"`. However, once the shims are finalized and shipping with Ember CLI by default I strongly want to do a pass of the docs and update all examples to include the imports, otherwise people are going to be very confused.

locks:
That is my own as well

PR #583, opened 2015-08-13
Move version for version switching to publishing step

### LINKED ISSUE #581: Version picker doesn't work from v2.0.0 pages
Steps to reproduce: 

1. Visit http://guides.emberjs.com/v2.0.0
2. Click on any of the pages on the right, let's say: http://guides.emberjs.com/v2.0.0/getting-started/first-app/
3. Try to change the version in the top left, to any version (even v2.0.0). It'll try to visit http://guides.emberjs.com/v2.0.0/getting-started/v1.13.0 (with the last digits being whatever you clicked) and that gives you a 404.

This only happens from v2.0.0 though.  (Note that it *doesn't* happen from just http://guides.emberjs.com/v2.0.0, but only pages further down in the URL than that)
I tried taking a look at this locally, but I can't figure out how to get `versions.json` to work in dev.

michaelrkn:
This will be just like we do for the version number for search. In the publishing step, we will replace `'CURRENT_REVISION_WILL_APPEAR_HERE_WHEN_DEPLOYED'` with the current version. Corresponding PR coming soon.

### DIFF
--- a/source/javascripts/app/guides/version-select.js.erb
+++ b/source/javascripts/app/guides/version-select.js.erb
@@ -4,8 +4,7 @@ $.ajax('/versions.json')
     versionSelectBox.select2({
       data: data
     });
-    var slugs = window.location.href.split('/');
-    var currentRevision = slugs[slugs.length - 2];
+    var currentRevision = 'CURRENT_REVISION_WILL_APPEAR_HERE_WHEN_DEPLOYED';
     versionSelectBox.val(currentRevision).change();
 
     versionSelectBox.on('change', function() {


swr:
Well done. Beat me to it.

swr:
I think this may cause problems when the other versions of the guides don't have the same sections. Hence why the older versions just switched to the root level of the guides.

For instance go here: http://guides.emberjs.com/v2.0.0/object-model/ and then switch to v1.10.0

michaelrkn:
@swr see https://github.com/emberjs/guides/issues/206#issuecomment-90096118. The 404 page is not working properly, unfortunately.

PR #588, opened 2015-08-14
Minor nitpick - callout empty arrays are not falsy

dmin:
I thought it would be helpful (especially to those new to JS) to be (slightly) more explicit about Ember treating ```[]``` differently here than "vanilla JS". A minor point I realize - do with it what you will.

As a new Ember user I'm impressed with the guides - great work! Thanks for all the work that went into these!

### DIFF
--- a/source/templates/conditionals.md
+++ b/source/templates/conditionals.md
@@ -10,7 +10,7 @@ We can use the `{{#if}}` helper to conditionally render a block:
 ```
 
 Handlebars will not render the block if the argument passed evaluates to
-`false`, `undefined`, `null` or `[]` (i.e., any "falsy" value).
+`false`, `undefined`, `null` or `[]` (i.e., any "falsy" value or an empty array).
 
 If the expression evaluates to falsy, we can also display an alternate template
 using `{{else}}`:


locks:
@wycats why are empty arrays falsey in Handlebars? It doesn't follow Mustache's "host language defines truthiness" spec?

source/templates/conditionals.md
```diff
@@ -10,7 +10,7 @@ We can use the `{{#if}}` helper to conditionally render a block:
 ```
 
 Handlebars will not render the block if the argument passed evaluates to
-`false`, `undefined`, `null` or `[]` (i.e., any "falsy" value).
+`false`, `undefined`, `null` or `[]` (i.e., any "falsy" value or an empty array).
```
mixonic: > any "falsy" value including an empty array

And I would merge :+1:

dmin: > any "falsy" value including an empty array

That wording, to me anyway, seems to imply that an empty array is just another falsy value in JS, which it is not.  My intention with the PR was to point out that even though an empty array in JS is truthy Handlebars will treat it as though it was falsy.

I can understand if you don't think that needs to be explained here - if that's the case I'd say its good as is.  I know its minor so I don't mean to belabor the point :smile:

mixonic:
I think my main determination here is that I loathe "falsy" in scare quotes. I'd like to rework it a little more than this, but IMO you are on the right track.

PR #591, opened 2015-08-14
Create style guide

michaelrkn:


STYLE_GUIDE.md
```diff
@@ -0,0 +1,30 @@
...
+
+The Guides are primarily meant to cover the "Ember happy path", and are not intended to be comprehensive. Leave edge cases and rarely-used features to the API documentation. Start each guide and each section with the simplest, most commonly-used features, and progress to the more advanced and less commonly-used features.
+
+Each guide should thoroughly explain the feature it documents, and include links to the API documentation. Links to authoritative sources of information on background concepts are also encouraged: for example, the _Handlebars Basics_ guide appropriately links to http://handlebarsjs.com/. Do not link to other outside content like blog posts or meetup slides, as reviewing and updating this content is better suited for content aggregators.
+
+Liberally use examples in your writing. For example, the sentence "Templates can contain expressions in double curly braces." should have added to it "Templates can contain expressions in double curly braces, like \`&lt;h1&gt;Welcome {{user.name}}&lt;/h1&gt;\`." A short example is often more clear than a long explanation. In fact, it is often to give two examples to make things even clearer.
```
locks: you accidentally a word in "it is often to give two examples"?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,30 @@
...
+
+Write code examples using ES2015; for example, use `let` instead of `var`. Omit the boilerplate that Ember CLI generates, especially the `import Ember from 'ember'` at the top of every file. Always use double quotes for strings. In fenced code blocks, include the filename or language after the triple-backticks, like <code>&#96;&#96;&#96;routes/kittens.js</code> or <code>&#96;&#96;&#96;hbs</code>. Paths should be relative to the `app` folder, except when talking about any other folder (like `config`), in which case they should be relative to the project root.
```
locks: I'd make this paragraph more than a list, and maybe put it at the end with a sub-heading

locks:
Great work :)

STYLE_GUIDE.md
locks: so, always use `let` instead of `var` correct? and when to use `const`?

michaelrkn:
Thanks for the feedback! Updated.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,35 @@
...
+
+Each guide should thoroughly explain the feature it documents, and include links to the API documentation. Links to authoritative sources of information on background concepts are also encouraged: for example, the _Handlebars Basics_ guide appropriately links to http://handlebarsjs.com/. Do not link to other outside content like blog posts or meetup slides, as reviewing and updating this content is better suited for content aggregators.
+
+Liberally use examples in your writing. For example, the sentence "Templates can contain expressions in double curly braces." should have added to it "Templates can contain expressions in double curly braces, like \`&lt;h1&gt;Welcome {{user.name}}&lt;/h1&gt;\`." A short example is often more clear than a long explanation. In fact, it is often helpful to give two examples to make things even clearer.
+
+In code samples:
+
+* Follow the [Ember Style Guide](https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md).
```
michaelrkn: I think this will be way simpler than trying to define our own standards. We can leave the discussion of `let`, `const`, shorthand method definition, and any other features here.

locks:
LGTM, we can iterate later.

STYLE_GUIDE.md
elyobo: That sounds good in theory @michaelrkn but the style guide doesn't appear to hold any sort of opinion on let vs const and so the examples in these guides seem to use `var`, `let` and `const` haphazardly (including frequent use of `let` in cases where it's clearly not modified, which is a good example of what not to do).

Is the correct thing to do to agitate for improvements to the style guide, or to have specific rules here for some consistency of style within the guide where the style guide otherwise falls short?

PR #594, opened 2015-08-14
Removes get helper guide

locks:
While talking with @mixonic an issue was raised about whether the get helper does even belong to the guides, or if it should only have API documentation.

If it belongs, I don't think it should have a whole guide for it, we should instead have a guide for helpers, one of which is get.

(guide as defined in https://github.com/emberjs/guides/pull/591/files#diff-1550192a99adeea1f5b641e7576116b7R6)

PR open for discussion


mixonic:
Agreed!

PR #595, opened 2015-08-15
Cleanup ahead of other work

mixonic:
Small cleanup work as I read.

### DIFF
--- a/source/object-model/index.md
+++ b/source/object-model/index.md
@@ -1,16 +1,18 @@
-You'll soon notice that standard JavaScript objects aren't used widely in Ember,
-except as key-value pairs (often referred to as _hashes_). This is because
-JavaScript objects don't support the ability to observe when an object changes,
-which Ember needs to update data throughout an application.
-Ember's object model builds on standard JavaScript objects to enable this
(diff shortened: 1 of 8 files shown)


source/object-model/index.md
```diff
@@ -1,16 +1,18 @@
...
+JavaScript objects don't support the observation of property value changes.
+Consequently, if an object is going to participate in Ember's binding
+system you may see an `Ember.Object` instead of a plan object.
+
+`Ember.Object` also provides a class system, supporting features like mixins
+and constructor methods. Some features in Ember's object model are not present in
+JavaScript classes or common patterns, but all are aligned as much as possible
+with the language and proposed additions.
+
+Ember also extends the JavaScript `Array` prototype with its
```
locks: "the native JavaScript `Array`" just to really hammer it home?

mixonic: JavaScript `Array` means only one thing. `native` doesn't really add anything. It isn't like Ember (or any other libs beside maybe immutable data structure libs) provide alternative array implementations. A javascript array is a javascript array.

PR #607, opened 2015-08-17
Update language for enabling support for data attributes

dmin:
It appears that there was originally two ways to enable support for data attributes, now there is only one.  This updates the language on the  "templates / binding element attributes" page to reflect that change.

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -53,12 +53,13 @@ renders the following HTML:
 ```html
 <a id="ember239" class="ember-view" href="#/photos">Photos</a>
 
-<input id="ember257" class="ember-view ember-text-field" type="text" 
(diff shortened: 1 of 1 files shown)


source/templates/binding-element-attributes.md
```diff
@@ -57,8 +57,9 @@ renders the following HTML:
        title="Name">
 ```
 
-There are two ways to enable support for data attributes. One way would be to add an
-attribute binding on the view, e.g. `Ember.LinkComponent` or `Ember.TextField` for the specific attribute:
+To enable support for data attributes an attribute binding must be
+added to the view, e.g. `Ember.LinkComponent` or `Ember.TextField`
```
rwjblue: Can you tweak this to say "component" instead of "view"?

dmin: Done.

locks:
:+1:

PR #608, opened 2015-08-17
Workaround for ember cli still using v1.13

balupton:
Reference https://github.com/ember-cli/ember-cli/issues/4671

### DIFF
--- a/source/getting-started/first-app.md
+++ b/source/getting-started/first-app.md
@@ -5,12 +5,21 @@ running:
 ember new my-new-app
 ```
 
-This will create a new `my-new-app` directory with your new Ember app inside.
-
-Once the generation process finishes, launch the Ember CLI development server:
+Which will generate a new `my-new-app` directory with your new Ember app inside. Change into it using:
 
 ```shell
 cd my-new-app
+```
+
+Currently, Ember CLI still installs Ember v1.13, this will change in the future. In the meantime, to install Ember v2.0, you must use bower to grab the latest version (if asked which version you want, press the number for ember 2):
+
+```shell
+bower install ember
+```
+
+To get started with your development, launch the Ember CLI development server using:
+
+```shell
 ember server
 ```
 
--- a/source/getting-started/first-app.md
+++ b/source/getting-started/first-app.md
@@ -11,13 +11,13 @@ Which will generate a new `my-new-app` directory with your new Ember app inside.
 cd my-new-app
 ```
 
-Currently, Ember CLI still installs Ember v1.13, this will change in the future. In the meantime, to install Ember v2.0, you must use bower to grab the latest version (if asked which version you want, press the number for ember 2):
+Currently, [Ember CLI still installs Ember v1.13](https://github.com/ember-cli/ember-cli/issues/4671), this will change in the future. In the meantime to install Ember v2.0, you can use Bower to grab the latest version (if asked which version you want, press the number for ember 2), do this by:
 
 ```shell
 bower install ember
 ```
 
-To get started with your development, launch the Ember CLI development server using:
+To start with your development, launch the Ember CLI development server using:
 
 ```shell
 ember server


locks:
Thanks for the PR, but I am not fond of this solution. I'd rather deal with the support questions and not have to change it back again and redeploy the guides when CLI 2.0 does get released. I understand how confusing this can be to newcomers, but hopefully it won't be for long :)

I also believe it is incomplete, as you have to also update Ember Data or your application will break due to Ember Data using deprecated code (that is removed in 2.0).

balupton:
Can we at least get a warning in there? As otherwise it is misleading to have the Ember 2.0 instructions not actually install Ember 2.0, when they are the default instructions. It's hard to be appealed by Ember as a newcomer, when the docs are not correct, and when efforts aren't made to correct them.

Happy to submit a new pull request when the cli changes.

greyhwndz:
I guess clarity over these matters would come a long way in supporting the pathways taken both by experienced and those who are just starting to look into being part of the community. Assuming that the guides is supposed to serve as a 'guide' :)

locks:
The installation documentation isn't wrong, as it doesn't mention a specific version on purpose. In an ideal world every 2.0 would have released in the same day, but alas :\
Like I said I understand that it is frustrating, but it won't be a problem very shortly.

jimbeaudoin:
@bravedick rm -rf bower_components && bower install ember#2­.0.0 --save && bower install ember-data#2.0.0-beta.2 --save

Which version of ember-cli you have?

bravedick:
`version: 1.13.8`

yes, 2.0.0-beta.2 fix it, but it was not said anywhere above. thx.

devinrhode2:
Hey @balupton, anybody that sees this version mismatch will probably end up reading this thread, like me, and as of 3 months later we still have a mis-match.

How should I go about using the current stable version of Ember and ember-data? How about using the beta?

michaelrkn:
Yup, totally not ideal. You still need to do as described in https://github.com/emberjs/guides/pull/608#issuecomment-132271843. Ember CLI 2.2 is supposed to be released any day now...

devinrhode2:
With a pretty new and pretty small app, I successfully used the dependencies listed: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/bower.json and here: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/package.json Note: this meant removing ember-resolver from bower and adding it package.json, although different versions.. i.e. if it's not your package, stick to the dependency version combinations listed in those two files.

devinrhode2:
..Although I'm getting this deprecation..
```
ember.debug.js:5679DEPRECATION: Usage of `ember/resolver` module is deprecated, please update to `ember-resolver`. [deprecation id: ember-resolver.legacy-shims]
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16340:7)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:16250:12)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:16340:7)
        at invoke (http://localhost:4200/assets/vendor.js:16356:7)
        at deprecate (http://localhost:4200/assets/vendor.js:16309:32)
        at Object.deprecate (http://localhost:4200/assets/vendor.js:26205:37)
        at http://localhost:4200/assets/vendor.js:80350:22
        at mod.state (http://localhost:4200/assets/vendor.js:150:29)
        at tryFinally (http://localhost:4200/assets/vendor.js:31:14)
```

devinrhode2:
I searched and searched and could not find any information regarding this deprecation. How do you lookup a deprecation by the id??

balupton:
@devinrhode2 I ended up giving up on Ember.

I hope Ember realises the extent of which things like this has a negative impact on people's perceptions and takeaways of the framework: https://github.com/balupton/is-ember-ready-yet

PR #618, opened 2015-08-19
Move routing configuration into Configuring Ember

michaelrkn:
Keeps the section focused on the happy path and consolidates the configuration options in one place.


source/configuring-ember/specifying-url-type.md
```diff
@@ -1,4 +1,12 @@
-By default the Router uses the browser's hash to load the starting state of your
+The Ember router has four options to manage your application's URL: `auto`
+(default), `hash`, `history`, and `none`. By default, Ember CLI configures the
```
mixonic: I see there is a list of details below but a few small pieces of context may be helpful here:

> The Ember router has four options to manage your application's URL: `hash` which are anchor-based URLs, `history` which uses the HTML5 History API, `auto` and `none`.

The default behavior is the next sentence, seems needless to repeat it above. For a newb, this quick context on what hash and history are helps to explain why there is an auto behavior that is so complex.

source/configuring-ember/specifying-url-type.md
```diff
@@ -1,4 +1,12 @@
...
+router to use `auto`, which will use `history` if supported by the user's
+browser and fall back to `hash` otherwise. You can change this option in
+`config/environment.js` under `ENV.locationType`.
+
+## hash
+
+The `hash` option uses the browser's hash to load the starting state of your
```
mixonic: > uses the browser's hash

Not part of the browser. A URL's anchor, or a URL's hash segment.

source/configuring-ember/specifying-url-type.md
```diff
@@ -13,30 +21,22 @@ Router.map(function() {
 });
 ```
 
+## history
```
mixonic: Maybe slightly out of scope so feel free to punt: History API is supported in everything except IE9 and JSBins, and thus I think it should be the first item in this list. `hash` should be portrayed as a fallback, not as the most important (first mentioned) kind of URL for an Ember app.

Good direction for a followup perhaps.

source/routing/index.md
```diff
@@ -30,24 +30,3 @@ or more of the following:
 2. Update a controller so that it represents a particular model.
 3. Change the template on screen, or place a new template into an
    existing outlet.
-
-###Logging Route Changes
```
mixonic: Debugging seems relevant to share here. Maybe something like mentioning the logging option and reminder the reader that the Ember CLI inspector can display routes. A screenshot with the pane open and an arrow to the tab would go a long way.

source/configuring-ember/specifying-url-type.md
locks: What about something like the following?

> `auto` which gracefully degrades, and `none`.

I'm otherwise :+1: with this idea.

source/configuring-ember/specifying-url-type.md
locks: is there an MDN-like reference on this we can link to?

source/configuring-ember/specifying-url-type.md
locks: :+1: `history`, `hash`, `auto` (because we already both the previous two), `none`?

source/routing/index.md
locks: link to http://guides.emberjs.com/v2.0.0/ember-inspector/routes/

michaelrkn: This is the first page that people learn about routes, and I think it's purpose is to introduce the subject at a high level and get them ready to dive into all of the details in the following pages. When you're learning about a new subject, debugging is only useful to know about after you've learned about the major important concepts, not before. Terrible analogy: It's like teaching somebody how to escape from a crashing airplane before they've learned how to fly it. 

In the Templates section, one of the last guides is Development Helpers, which is essentially the debugging tools for templates. Similarly, we could add a section called Debugging Routes or something to the end of the Routing section that included http://guides.emberjs.com/v2.0.0/configuring-ember/debugging/#toc_log-router-transitions and info from Ember Inspector... But currently those logging config options are all grouped together, as is Ember Inspector, and if we want to break them out into their relevant sections I suggest we open another issue for doing that so we can discuss it for all of the Guides, instead of just for Routing.

michaelrkn:
Thanks for all the great feedback! I've updated the guide and it's ready for another review.

source/routing/index.md
mixonic: :+1:

source/configuring-ember/embedding-applications.md
```diff
@@ -33,3 +34,17 @@ var ENV = {
   locationType: 'none'
 };
 ```
+
+### Specifying a Root URL
+
+If your Ember application is one of multiple web applications served from the same domain, it may be necessary to indicate to the router what the root URL for your Ember application is. By default, Ember will assume it is served from the root of your domain.
```
mixonic: (newlines would be sweeet)

mixonic:
:confetti_ball:

PR #619, opened 2015-08-19
Remove "setting up controller" page from routing

michaelrkn:
Since we don't want people thinking controllers are a good idea.


locks:
This makes me a bit icky. Have we decided to completely kill controllers? Routable components are a long ways away.

MarkMT:
+1
Good goal for the future but perhaps best on a branch until controllers are actually fully deprecated. The guides don't just help people know how to build a new app; they also help them understand existing code examples they run into - something I've found useful as a newbie.

mixonic:
:-1: IMO the goal of the guides for controllers should be demonstrating how to use them in ways that are future-safe and easy to convert forward to components. I do not favor dropping them from the guides until we have a real routable component solution. They simply serve a very narrow case today: backing routed templates.

michaelrkn:
What should be in this guide, then? Lines 33-39 seem like the only things that might be useful to know about. Should I just remove everything but those lines?

locks:
@acorncom poke

michaelrkn:
@mixonic can you give some guidance re: https://github.com/emberjs/guides/pull/619#issuecomment-133044923?

greyhwndz:
QueryParameters is still dependent on Controllers to fully harness its functionality of making URLs a first-class thing in Ember

locks:
@greyhwndz query params are unrelated to the section removed.

greyhwndz:
@locks: ok. Will query params be found in the guides?

PR #620, opened 2015-08-20
Improve routing/templates

michaelrkn:
Improved explanation of default behavior. Added more examples. Used headings to split up different parts.

### DIFF
--- a/source/routing/rendering-a-template.md
+++ b/source/routing/rendering-a-template.md
@@ -25,17 +25,6 @@ export default Ember.Route.extend({
 });
 ```
 
-If you want to use a different controller than the route handler's
-controller, pass the controller's name in the `controller` option:
-
-```app/routes/post.js
-export default Ember.Route.extend({
-  renderTemplate() {
-    this.render({ controller: 'favoritePost' });
-  }
-});
-```
-
 Ember allows you to name your outlets. For instance, this code allows
 you to specify two outlets with distinct names:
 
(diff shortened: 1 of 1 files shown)


source/routing/rendering-a-template.md
```diff
@@ -1,21 +1,19 @@
...
+By default, a route handler will render the template with the same name as the
+route. For example, the `post` route will render the `post.hbs` template, and
+the `posts.new` route will render `posts/new.hbs`.
 
-```app/router.js
-Router.map(function() {
-  this.route('post');
-});
-```
+Each template will be rendered into the `{{outlet}}` of its parent route's
```
locks: this is tricky because you can name the route `posts.new` without it being the child, kinda. would it benefit from a small router example?

source/routing/rendering-a-template.md
```diff
@@ -25,16 +23,7 @@ export default Ember.Route.extend({
...
-controller, pass the controller's name in the `controller` option:
-
-```app/routes/post.js
-export default Ember.Route.extend({
-  renderTemplate() {
-    this.render({ controller: 'favoritePost' });
-  }
-});
-```
+## Named Outlets
```
locks: We should confirm if components replace all use cases of named outlets and if this section should be removed from the guide. cc @rwjblue

source/routing/rendering-a-template.md
mixonic: :+1:

source/routing/rendering-a-template.md
mixonic: @ef4 is also bullish on that story.

source/routing/rendering-a-template.md
```diff
@@ -1,21 +1,19 @@
-One of the most important jobs of a route handler is rendering the
+One job of a route handler is rendering the
```
mixonic: :clap:

source/routing/rendering-a-template.md
ef4: I am fairly convinced that we don't want to encourage named outlets for anything. As an alternative, we need to write examples like a basic sidebar and modal using component + service.

source/routing/rendering-a-template.md
trek: :clap: :clap:

michaelrkn:
Thanks for the feedback! I've updated the PR. Re "we need to write examples like a basic sidebar and modal using component + service", it sounds like that's not appropriate for the Routing section, so I've just removed the stuff on named outlets without replacement. Maybe somebody with a better understanding of how the above example might work can open an issue to write a guide for it?

PR #566, opened 2015-08-12
Routing cleanup

michaelrkn:
* Remove large tables in favor of shorter explanations
* Provide more examples
* Remove redundant information
* Remove many (but not all) references to controllers
* Remove the page on "The Application Template" since it is more about routing

This is just a start - there is plenty more work to be done on this section, especially on removing stuff about controllers.

PR #629, opened 2015-08-20
Templates cleanup, tweaks

mixonic:
* Update "Handlebars Basics" to use more modern terminology (binding-aware is not a thing we ever say now)
* "Handlebars Basics" linked to "application template", removed in https://github.com/emberjs/guides/pull/566. This leaves us with no examples of basic `{{outlet}}` usage, only the text explanations on http://guides.emberjs.com/v2.0.0/routing/defining-your-routes/#toc_the-application-route. I think it would be nice to have some simple example of `{{outlet}}` usage in templates. Good followup issue.
* I propose removing the "specifying keys" section of `{{#each}}` guides http://guides.emberjs.com/v2.0.0/templates/displaying-a-list-of-items/#toc_specifying-keys
  * "The {{#each}} helper does provide the ability to override the key to use the array's @index for situations where you would need this." is not actually helpful. What situations?
  * The description of default behavior is inaccurate, we currently do a combo of identity + binning
  * `@index` should really never be needed in 1.13.8 or 2.0+. Possibly you want a custom key, but it would be a property
  * Sufficient documentation exists in the [API docs](http://emberjs.com/api/classes/Ember.Templates.helpers.html#toc_specifying-keys).
  * Really, we do not want new users to need to reach for this functionality. Seems more confusing than helpful.
* Introduce helper invocation styles of inline, nested, and block during the conditionals section. That there are three ways to call a helper is cross-cutting knowledge useful across the rest of the templates section. For example it is much easier to teach here than to teach while going over actions. Also, it flows nicely that we can talk about properties (display a value), inline helpers (display a value), nested helper (return a value), then get to block helpers (manage DOM, yield to blocks).

### DIFF
--- a/source/templates/conditionals.md
+++ b/source/templates/conditionals.md
@@ -1,7 +1,49 @@
-Sometimes you may only want to display part of your template if a property
-exists.
+Conditional statements bring a minimum of logic into Ember templating.
+Statements like `if` and `unless` are implemented as built-in helpers. Helpers
+can be invoked three ways.
 
-We can use the `{{#if}}` helper to conditionally render a block:
+The first style of invocation is **inline invocation**. This looks similar to
+displaying a property, but helpers accept arguments. For example:
+
+```handlebars
+<div>
+  {{if isFast "zoooom" "putt-putt-putt"}}
+</div>
+```
+
+`{{if}}` in this case returns `"zoooom"` when `isFast` is true and
+`"putt-putt-putt"` when `isFast` is false. Helpers invoked as inline expressions
+render a single value, the same way that properties are a single value.
+
+Inline helpers don't need to be used inside HTML tags. They can also be used
+inside attribute values:
+
+```handlebars
+<div class="is-car {{if isFast "zoooom" "putt-putt-putt"}}">
+</div>
+```
+
+**Nested invocation** is another way to use a helper. Just like inline helpers,
+nested helpers generate and return a single value. For example, this template
+only renders `"zoooom"` if both `isFast` and `isFueled` are true:
+
+```handlebars
+<div>
+  {{if isFast (if isFueled "zoooom")}}
+</div>
+```
+
+The nested helper is called first returning `"zoooom"` only if `isFueled` is
+true. Then the inline expression is called, rendering the nested helper's
+value (`"zoooom"`) only if `isFast` is true.
+
+The third form of helper usage is **block invocation**. Use block helpers
+to render only part of a template. Block invocation of a helper can be
+recognized by the `#` before the helper name, and the closing `{{/` double
+curly brace at the end of the invocation.
+
+For example, this template conditionally shows
+properties on `person` only if that it is present:
 
 ```handlebars
 {{#if person}}
@@ -9,11 +51,12 @@ We can use the `{{#if}}` helper to conditionally render a block:
 {{/if}}
 ```
 
-Handlebars will not render the block if the argument passed evaluates to
-`false`, `undefined`, `null` or `[]` (i.e., any "falsy" value or an empty array).
+`{{if}}` checks for truthiness, which means all values except `false`,
+`undefined`, `null`, `''`  or `[]` (i.e., any JavaScript falsy value or an
+empty array).
 
-If the expression evaluates to falsy, we can also display an alternate template
-using `{{else}}`:
+If a value passed to `{{#if}}` evaluates to falsy, the `{{else}}` block
+of that invocation is rendered:
 
 ```handlebars
 {{#if person}}
@@ -23,7 +66,8 @@ using `{{else}}`:
 {{/if}}
 ```
 
-Handlebars also supports chained else helpers, the most common use being else if. An example:
+`{{else}}` can chain helper invocation, the most common usecase for this being
+`{{else if}}`:
 
 ```handlebars
 {{#if isAtWork}}
@@ -33,26 +77,12 @@ Handlebars also supports chained else helpers, the most common use being else if
 {{/if}}
 ```
 
-To only render a block if a value is falsy, use `{{#unless}}`:
+The inverse of `{{if}}` is `{{unless}}`, which can be used in the same three
+styles of invocation. For example, this template only shows an amount due when the
+user has not paid:
 
 ```handlebars
 {{#unless hasPaid}}
   You owe: ${{total}}
 {{/unless}}
 ```
-
-`{{#if}}` and `{{#unless}}` are examples of block expressions. These allow you
-to invoke a helper with a portion of your template. Block expressions look like
-normal expressions except that they contain a hash (#) before the helper name,
-and require a closing expression.
-
-You can also use the inline `{{if}}` helper:
-
-```handlebars
-<span class={{if isEnabled "enabled" "disabled"}}>Warning!</span>
-```
-
-In this case, if the `isEnabled` property is `true`, the `enabled` class will be
-added. The second argument is optional and only necessary if you want to handle
-the `false` case. Here, if the property is `false`, the class `disabled` will be
-added.
--- a/source/templates/displaying-a-list-of-items.md
+++ b/source/templates/displaying-a-list-of-items.md
@@ -1,4 +1,10 @@
-If you need to enumerate over a list of objects, use Handlebars' `{{#each}}` helper:
+To iterate over a list of items, use the `{{#each}}` helper. The first
+argument to this helper is the array to be iterated, and the value being
+iterated is yielded as a block param. Block params are only available inside
+the block of their helper.
+
+For example, this template iterates an array named `people` that contains
+objects. Each item in the array is provided as the block param `person`.
 
 ```handlebars
 <ul>
@@ -8,10 +14,22 @@ If you need to enumerate over a list of objects, use Handlebars' `{{#each}}` hel
 </ul>
 ```
 
-The template inside of the `{{#each}}` block will be repeated once for
-each item in the array, with the each item set to the `person` keyword.
+Block params, like function arguments in JavaScript, are positional. `person`
+is what each item is named in the above template, but `human` would work just
+as well.
 
-The above example will print a list like this:
+The template inside of the `{{#each}}` block will be repeated once for
+each item in the array, with the each item set to the `person` block param.
+
+Given an input array like:
+
+```js
+[ {name: 'Yehuda'},
+  {name: 'Tom'   },
+  {name: 'Trek'  } ]
+```
+
+The above template will render HTML like this:
 
 ```html
 <ul>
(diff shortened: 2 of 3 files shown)


source/templates/conditionals.md
```diff
@@ -1,7 +1,49 @@
-Sometimes you may only want to display part of your template if a property
-exists.
+Conditional statements bring a minimum of logic into Ember templating.
+Statements like `if` and `unless` are implemented as built-in helpers. Helpers
```
michaelrkn: At this point in the Templates section, the term "helpers" hasn't been defined yet.

source/templates/conditionals.md
```diff
@@ -1,7 +1,49 @@
...
+can be invoked three ways.
```
michaelrkn: Oh, I'm starting to understand what you've done with this guide - it now about both the specific case of the conditionals helpers, as well as the general case of all helpers. I think this works, but I think it's a little confusing without stating at the beginning somewhere that the purpose of the page is both to introduce conditionals (which is the title) **and** to give an overview of how helpers work in general.

source/templates/conditionals.md
```diff
@@ -1,7 +1,49 @@
...
+</div>
+```
+
+The nested helper is called first returning `"zoooom"` only if `isFueled` is
+true. Then the inline expression is called, rendering the nested helper's
+value (`"zoooom"`) only if `isFast` is true.
+
+The third form of helper usage is **block invocation**. Use block helpers
+to render only part of a template. Block invocation of a helper can be
+recognized by the `#` before the helper name, and the closing `{{/` mustache
```
michaelrkn: I think remove the word "mustache" - it doesn't add anything, and not everybody knows to turn their head sideways to the left ;)

source/templates/displaying-a-list-of-items.md
```diff
@@ -1,4 +1,11 @@
-If you need to enumerate over a list of objects, use Handlebars' `{{#each}}` helper:
+To iterate over a list of items, use the `{{#each}}` helper. The first
+argument to this helper is the array to be iterated, and the value being
+iterated is yielded as a block param. Block params are only available inside
+the block of their helper.
+
+For example, this template iterates an array named `people` that contains
+objects. Each item in the array
+is yielded as a block param named `person` by the template.
```
michaelrkn: "yielded as a block param" - is there a less jargony way to say this?

michaelrkn: I like how you said it below - "with the each item set to the `person` block param".

source/templates/displaying-a-list-of-items.md
```diff
@@ -21,15 +36,19 @@ The above example will print a list like this:
...
-application adds a new item to the array, or removes an item, the DOM
-will be updated without having to write any code. Note that a `[].push()`
-will not update the helper. Adding items need to be done with `[].pushObject`,
-and related [Ember Mutable Array methods](http://emberjs.com/api/classes/Ember.MutableArray.html) so that Ember can observe the change.
+Like other helpers, the `{{#each}}` helper is bound.  If a new item is added to
+or removed from the iterated array, the DOM will be updated without having to
+write any additional code.
 
-### Accessing the list item's `index`
+Ember requires that you use special methods to update bound arrays, for example
```
michaelrkn: This is covered in https://github.com/emberjs/guides/blob/master/source/object-model/enumerables.md#use-of-observable-methods.

source/templates/conditionals.md
michaelrkn: Oh I see your addition defining the term below.

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
 Ember.js uses the [Handlebars templating library](http://www.handlebarsjs.com)
-to power your app's user interface. Handlebars templates are just like
-regular HTML, but also give you the ability to embed expressions that
-change what is displayed.
+to power your app's user interface. Handlebars templates contain static HTML and
+dynamic content inside what we call mustaches: `{{}}`.
```
michaelrkn: Is "mustache" really the term we want to use? It's not currently used anywhere in the Guides. Maybe "double curly braces"? That term is used elsewhere already.

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
...
+automatically updated to the latest value.
 
 ### Defining Templates
 
-The first thing you should change is your [application template](../the-application-template) that is created
-automatically for you and is displayed when your app loads.
-
-Next, you can define templates in the `app/templates` folder. By default,
-a route will render a template with the same name as the route.
+Templates in an Ember CLI app are usually stored in the `app/templates` folder.
```
michaelrkn: I'd say omit "usually", since we're trying to focus on a single happy path in the Guides.

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
...
-```handlebars
+```app/templates/application.hbs
 Hello, <strong>{{firstName}} {{lastName}}</strong>!
 ```
 
-This would look up the `firstName` and `lastName` properties from the
-controller, insert them into the HTML described in the template, then
-put them into the DOM.
+The `firstName` and `lastName` properties are read from the
+context (the application controller in this case), and rendered inside the
```
michaelrkn: backticks around "application"?

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
...
+`<strong>` HTML tag.
 
-By default, your top-most application template is bound to your application controller. Note that this file is not shown by default because it is created behind the scenes by Ember CLI. To customize the controller, create the following file:
+To provide a `firstName` and `lastName` to the above template, properties
+must be added to the application controller. If you are following along with
+and Ember CLI application, you may need to create this file:
```
michaelrkn: and -> an

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
...
 
 ```app/controllers/application.js
+import Ember from 'ember';
+
 export default Ember.Controller.extend({
-  firstName: "Trek",
-  lastName: "Glowacki"
+  firstName: 'Trek',
+  lastName: 'Glowack'
```
michaelrkn: I don't think we have agreement about single vs double quotes. I'd love for that to get added to https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md and then we change it everywhere in the Guides with a single PR.

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
...
+The `app/templates/application.hbs` file is the main template for your
+application. If you have a new Ember app, you can change the contents of that
+file and that content should appear in the browser. As a new user, you
+may want to experiment with content in the application template before diving
+in to routing.
+
+If you want to learn more about how routes and templates work together in
+Ember, jump to
+[The Application Route](../../routing/defining-your-routes/#toc_the-application-route)
+section of the guides.
```
michaelrkn: I don't really like the idea of encouraging people to jump around in the Guides. I'd like to get them to a point where you can sanely read them from top to bottom. If people want to jump to another section, they should start at the beginning of that section.

source/templates/handlebars-basics.md
```diff
@@ -1,64 +1,72 @@
...
 <h1>Kittens</h1>
 <p>Kittens are the cutest!</p>
 ```
 
-If you would like to create a template that is shared across many areas of your site, you should investigate [components](../../components/defining-a-component/).
+The `app/templates/application.hbs` file is the main template for your
+application. If you have a new Ember app, you can change the contents of that
+file and that content should appear in the browser. As a new user, you
+may want to experiment with content in the application template before diving
+in to routing.
```
michaelrkn: I think that the tutorial should be the place where we provide specific instructions on how to play around with creating a specific Ember app, and the rest of the Guides should be explanations with code examples, not instructions on how to try out different features.

michaelrkn:
Re: {{outlet}} helper, I've opened https://github.com/emberjs/guides/pull/630.

source/templates/handlebars-basics.md
rwjblue: The Ember codebase enforces single quote usage for JS, this repo should as well.

michaelrkn: https://github.com/emberjs/ember.js/pull/12159

source/templates/conditionals.md
mixonic: The first paragraph reads:

> Conditional statements bring a minimum of logic into Ember templating.
> Statements like `if` and `unless` are implemented as built-in helpers. Helpers
> can be invoked three ways.

Conditional helpers provide a nice contextual way to introduce the different usages, because they are so universally useful. I'm not sure what to change here. Adding a section "using Ember helpers" could precede conditionals, but would probably still want to cover conditionals as the example use-case. That or talk about adding custom helpers but, that move seems aggressive to me.

Perhaps this section could be re-named "Using Helpers"? It defaults in the other direction away from conditionals, but it perhaps more of a useful title.

source/templates/displaying-a-list-of-items.md
mixonic: Adjusted this.

source/templates/displaying-a-list-of-items.md
mixonic: It seems worth re-iterating? It is a common stumbling block for new users, and I fear not everyone will read the guides front-to-back.

source/templates/handlebars-basics.md
mixonic: "double curly braces" does not seem an optimal replacement for all uses in this document. Having a term is useful.

I really dislike "Handlebars expression", though that is what http://handlebarsjs.com/ introduces them as. In conversation we never use that term though, and mustache is a lot more common. I'd hoped to start unifying that common parlance with the language we use in the guides. Historically even Handlebars ties back to Mustache.js, from whence the name arises.

Regardless I've fallen back to "Handlebars expression", and audited the PR to remove usages of mustache.

source/templates/handlebars-basics.md
mixonic: significantly, they are not with pods. But the usually provides little help on that front, removed.

source/templates/handlebars-basics.md
mixonic: `ApplicationController`, yes, but we don't use that terminology in Ember CLI. :-/

"The controller in `app/controllers/application.js`" is far too verbose. Quoting `application` here would not really refer to anything specific code-wise though, hence I prefer to not backtick.

source/templates/handlebars-basics.md
mixonic: :+1: thanks!

source/templates/handlebars-basics.md
mixonic: I agree with the sentiment. With more time a linear flow might be perfected, but this seems reasonable.

We've just spent a paragraph talking about the application template, and before that introduced a `kitty.hbs` template they basically have no idea how to use. And we about to suggest they try using the `if` helper without knowing how to provide properties to their templates.

IMO this is an excellent point to bridge them, and to hint at where they can find answers if the are confused by how to provide a `firstName` in the following sections.

mixonic:
Ace, rebased with several comments addressed.

source/templates/handlebars-basics.md
michaelrkn: Maybe what we should do is move the Routing section before Templates and eliminate lines 21-29. I think that flows better - then readers already know how templates (and, down the line, components) get loaded and we can avoid that discussion in the specific sections.

source/templates/handlebars-basics.md
michaelrkn: :+1:

source/templates/conditionals.md
michaelrkn: Cool, I'm on board. I think I just approached it with a certain mindframe and the flow didn't fit into my mindframe, but taking a fresh look I think it works the way you've done it here.

PR #636, opened 2015-08-21
Focus the routing/models guide on the ember data happy path

michaelrkn:
* Move Templates before Models since the explanation of models references templates
* Make all examples use Ember Data
* Remove info on refreshing model, since it is not a common use case


source/routing/specifying-a-routes-model.md
```diff
@@ -1,120 +1,39 @@
...
-until that promise resolves to render the `pullRequests` template.
-
-(For more information on jQuery's XHR functionality, see
-[jQuery.ajax](http://api.jquery.com/jQuery.ajax/) in the jQuery
-documentation.)
-
-Because Ember supports promises, it can work with any persistence
-library that uses them as part of its public API. You can also use many
-of the conveniences built in to promises to make your code even nicer.
+Typically, the `model` hook should return an Ember Data record.
```
locks: Hhhmm, won't this lead even more people into thinking they _have_ to send a `DS.Model` instance from the model hook? Could we maybe have some sub-headers for like POJO (short one), and then Ember Data (which then links to `/models`)?

source/routing/specifying-a-routes-model.md
```diff
@@ -156,61 +75,11 @@ export default Ember.Route.extend({
...
+photo's ID (`params.photo_id`) as an argument to Ember Data's `findRecord`
+method.
 
 Note: A route with a dynamic segment will only have its `model` hook called
 when it is entered via the URL. If the route is entered through a transition
 (e.g. when using the [link-to](../../templates/links) Handlebars helper), then a model context is
 already provided and the hook is not executed. Routes without dynamic segments
 will always execute the model hook.
-
-### Refreshing your model
```
mixonic: Refresh still seems relevant? Or too far afield of normal usage?

mixonic:
Would be happy to see these poor `getJSON` suggestions go. Some of the language in "Asynchronously Loading Models" specifically seems unfortunate to dump entirely.

source/routing/specifying-a-routes-model.md
michaelrkn: My intuition says too far afield - my informal gut check is if it's not something I think will benefit more than 20% of users, it shouldn't go in the Guides. @locks what do you think?

source/routing/specifying-a-routes-model.md
michaelrkn: How about:

> Typically, the `model` hook should return an [Ember Data](../../models/) record, but it can also return any [promise](https://www.promisejs.org/) object, or a plain JavaScript object or array.

michaelrkn:
@mixonic I think it's only useful if you aren't using Ember Data, which we've decided not to address in the Guides.

source/routing/specifying-a-routes-model.md
locks: I'd say nix it. We can see about bringing it back if needed.

locks:
:boom: :hammer:

PR #640, opened 2015-08-21
Always use paths relative to the project root

michaelrkn:


mixonic:
:+1: will wait a moment for any further feedback

locks:
3's a quorum.

PR #638, opened 2015-08-21
Use consistent quotations

HeroicEric:
Fixes https://github.com/emberjs/guides/issues/634

Issue #634, opened 2015-08-20
Change all double quotes to single quotes


michaelrkn:
Per https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md#strings.

HeroicEric:
@michaelrkn in templates too? I usually use double-quotes in templates because HTML so wanted to clarify

michaelrkn:
@rwjblue can you give some guidance here?

rwjblue:
JS === single quotes
HTML / HBS === double quotes

Ember styleguide doesn't really address templates because there are very very few templates in the Ember repo.

HeroicEric:
working on this now

josenberg:
Why the use of quotes was defined this way?

PR #641, opened 2015-08-21
Add double-quotes in templates to styleguide

HeroicEric:
Adds a double-quotes in templates rule to the style guide in response to https://github.com/emberjs/guides/issues/634 and https://github.com/emberjs/guides/pull/638


locks:
:fire:

PR #642, opened 2015-08-21
Prefixes paths with `app/` after #640

locks:


michaelrkn:
Thank you for cleaning up my mess ;)

PR #647, opened 2015-08-22
cleanup enumerables page

kagemusha:
A few changes to make the language more consistent. `some` is a javascript method, whereas the Ember enumerables method is `any`.


rwjblue:
:+1:

PR #651, opened 2015-08-23
[WIP] Models section rewrite

locks:
- [ ] New TOC
- [ ] New top-level adapters section
- [ ] New content

**sidenote** I might split this into separate PRs depending on how I go about it. For now I'm trying to make the commits atomic, so if you're curious and want to review, go commit-a-commit.


source/models/defining-models.md
```diff
@@ -3,32 +3,24 @@ data that you present to the user. Anything that the user expects to see
...
-
-```html
-<script type="text/javascript" src="ember.js"></script>
-<script type="text/javascript" src="ember-data.js"></script>
-```
-
-For every model in your application, create a subclass of `DS.Model`:
+When you want a new model for your application you need to create a new file
+under the models folder and extend from `DS.Model`. This is more conveniently
+done by using one of Ember CLI's generator commands. For instant, let's create
```
greyhwndz: @locks: perhaps you mean "instance"? :)

source/models/defining-models.md
```diff
@@ -180,6 +174,18 @@ export default DS.Model.extend({
 });
 ```
 
+#### Synchronous relationships
+
+Starting with Ember Data 2.0 all relationships are asynchronous by default. If
+you want to specify a synchronous relationship you need to pass an async option
+set to false:
```
HeroicEric: "false" -> `false`

PR #656, opened 2015-08-23
tighten up language on computed properties and aggregate data

kagemusha:
removes a little repitition

### DIFF
--- a/source/object-model/computed-properties-and-aggregate-data.md
+++ b/source/object-model/computed-properties-and-aggregate-data.md
@@ -58,10 +58,10 @@ todosController.get('remaining');
 Note that `@each` only works one level deep. You cannot use nested forms like
 `todos.@each.owner.name` or `todos.@each.owner.@each.name`.
 
-Sometimes, you may only care if the items of an array have been added, removed, or replaced, 
-rather than changes to properties on the items themselves. For example, items may have been 
-added to or removed from the array. In this case, you can create a computed property as follows:
-
+Sometimes you don't care if properties of individual array items change. In this
+case use the `[]` key instead of `@each`. Computed properties dependent on an array
+using the `[] key will only update if items are added to or removed from the array,
+or if the array property is set to a different array. For example:
 
 ```app/controllers/todos.js
 export default Ember.Controller.extend({
@@ -78,12 +78,11 @@ export default Ember.Controller.extend({
 });
 ```
 
-Note here that we use the special key `[]` instead of `@each.foo` (which is used for observing 
-a specific dependent key, such as `foo`, on each item). This instructs Ember.js to only care about 
-the array itself changing (such as adding/removing items), rather than changes to the items themselves.
+Here, `indexOfSelectedTodo` depends on `todos.[]`, so it will update if we add an item
+to `todos`, but won't update if the value of `isDone` on a `todo` changes.
 
 Several of the [Ember.computed](http://emberjs.com/api/classes/Ember.computed.html) macros 
-utilize the `[]` key to implement common use-cases as well. For example, if you wanted to 
+utilize the `[]` key to implement common use-cases. For instance, to
 create a computed property that mapped properties from an array, you could use 
 [Ember.computed.map](http://emberjs.com/api/classes/Ember.computed.html#method_map)
 or build the computed yourself:


source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -58,10 +58,9 @@ todosController.get('remaining');
 Note that `@each` only works one level deep. You cannot use nested forms like
 `todos.@each.owner.name` or `todos.@each.owner.@each.name`.
 
-Sometimes, you may only care if the items of an array have been added, removed, or replaced, 
-rather than changes to properties on the items themselves. For example, items may have been 
-added to or removed from the array. In this case, you can create a computed property as follows:
-
+Sometimes you care about changes to the composition of the array (cases 2, 3,
```
locks: I don't feel like the callback to the list is beneficial to remove the duplication. If it were right after the list I wouldn't have minded it as much. Also, the working "composition of the array" even with the callbacks is too dense.

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -78,12 +77,11 @@ export default Ember.Controller.extend({
 });
 ```
 
-Note here that we use the special key `[]` instead of `@each.foo` (which is used for observing 
-a specific dependent key, such as `foo`, on each item). This instructs Ember.js to only care about 
-the array itself changing (such as adding/removing items), rather than changes to the items themselves.
+Here, `indexOfSelectedTodo` depends on `todos.[]`, so will update if we add an item
```
locks: `so it will update when we add`

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -78,12 +77,11 @@ export default Ember.Controller.extend({
...
+to `todos`, but won't update if the value of`isDone` on a `todo` changes.
 
 Several of the [Ember.computed](http://emberjs.com/api/classes/Ember.computed.html) macros 
-utilize the `[]` key to implement common use-cases as well. For example, if you wanted to 
+utilize the `[]` key to implement common use-cases. For instance,
```
locks: `For instance, to`

locks:
Hi @kagemusha, thanks for contributing. I left some notes on the diff, can you address them?

source/object-model/computed-properties-and-aggregate-data.md
kagemusha: i think `when` here creates expectation that something must happen, whereas it may not, so i feel `if` is a little more neutral and appropriate.  but a very subtle diff, so if you have strong feeling for `when` we can change.

kagemusha:
@locks thanks for looking at.  let me know if this is better.

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -58,10 +58,9 @@ todosController.get('remaining');
 Note that `@each` only works one level deep. You cannot use nested forms like
 `todos.@each.owner.name` or `todos.@each.owner.@each.name`.
 
-Sometimes, you may only care if the items of an array have been added, removed, or replaced, 
-rather than changes to properties on the items themselves. For example, items may have been 
-added to or removed from the array. In this case, you can create a computed property as follows:
-
+Sometimes you care about changes to the contents of the array (i.e. items are added or
+removed, or the property is set to a new array), but not changes to properties of
```
mixonic: > or the property is set to a new array

This is explicitly *not* a change to the contents of an array. It is replacing a property with a new value.

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -78,12 +77,11 @@ export default Ember.Controller.extend({
 });
 ```
 
-Note here that we use the special key `[]` instead of `@each.foo` (which is used for observing 
-a specific dependent key, such as `foo`, on each item). This instructs Ember.js to only care about 
-the array itself changing (such as adding/removing items), rather than changes to the items themselves.
+Here, `indexOfSelectedTodo` depends on `todos.[]`, so it will update if we add an item
+to `todos`, but won't update if the value of`isDone` on a `todo` changes.
```
mixonic: Missing a space after "of"

source/object-model/computed-properties-and-aggregate-data.md
kagemusha: @mixonic edited.  pls take a look.

locks:
Thank you!

PR #659, opened 2015-08-24
component templates are in the templates directory

mcfiredrill:
It looks like this was a small typo. I believe component templates are supposed to be in app/templates/components/, as mentioned further on down the page.


locks:
Thanks!

PR #661, opened 2015-08-24
Link folks to services from controllers page

acorncom:


source/controllers/dependencies-between-controllers.md
```diff
@@ -46,4 +46,4 @@ export default Ember.Controller.extend({
 ```
 
 For more information about aliases, see the API docs for
-[aliased properties](http://emberjs.com/api/#method_computed_alias).
+[aliased properties](http://emberjs.com/api/#method_computed_alias).  If you need have more extensive "data sharing" needs across your app, see the [services page](../../services/), which largely replaces injected controllers.
```
locks: why two spaces after comma?

locks: and I'm not sure about the phrasing. paging dr @tomdale.

acorncom: Re: two spaces after the period.  Long habit ingrained by my English teacher in high-school ;-)

PR #663, opened 2015-08-24
Tweaking multiple models page to be cleaner

acorncom:


source/controllers/representing-multiple-models.md
```diff
@@ -24,10 +24,7 @@ each song:
 </ul>
 ```
 
-You can use the controller to collect aggregate information about
-the model it hosts. For example, imagine we want to display the
-number of songs that are over 30 seconds long. We can add a new computed
-property called `longSongCount` to the controller:
+You can use the controller to collect aggregate information about its `model`. Imagine we want to display the number of songs that are over 30 seconds long. We can add a new computed property called `longSongCount` to the controller:
```
locks: You forgot to manually introduce line breaks like the rest of the guides, can you add them?

source/controllers/representing-multiple-models.md
```diff
@@ -53,7 +50,7 @@ Now we can use this property in our template:
 
 ### Representing Several Models
 
-Multiple models can be returned through `Ember.RSVP.hash`. We can return a hash from the `model` hook, but `Ember.RSVP.hash` takes parameters that return promises and when all parameter promises resolve then the `Ember.RSVP.hash` promise resolves. To continue with the above example, if we also wanted to return the `model:album` records then our `model` hook would look like:
+Multiple models can be returned through [Ember.RSVP.hash](http://emberjs.com/api/classes/RSVP.html#method_hash). We can return a hash from the `model` hook, but `Ember.RSVP.hash` takes parameters that return promises and when all parameter promises resolve then the `Ember.RSVP.hash` promise resolves. To continue with the above example, if we also wanted to return the `model:album` records then our `model` hook would look like:
```
locks: line breaks, plz.
can you change `above example` to `example above`? I think it's still correct and less confusing.

acorncom:
Those changes are made (including line breaks).  I didn't see anything in the `CONTRIBUTING.md` file about line breaks, is that documented somewhere that I missed?

mixonic:
@acorncom it is just the style used through much of the rest of the guides, breaking at 80 characters to help code review and interaction in an editor. Thanks much!

PR #665, opened 2015-08-25
Replaced Ember.create with Object.create as per deprecation

rogersillito:
Changed code sample to bring in line with http://emberjs.com/deprecations/v1.x/#toc_ember-create


locks:
Thank you!

PR #556, opened 2015-08-09
Tweak on the Getting Started section

mixonic:
I'm aware this may be going away. I felt it good to give it a wash anyway as I troll for modern helper/action usage.

PR #668, opened 2015-08-26
Edits to Core Concepts

mixonic:
Changes to Core Concepts.

* Focus prose, edit for visual pacing (intersperse examples and explanation instead of long paragraphs, use lists)
* Introduce Helpers as a first-class part of the framework. Since they are a simple concept (a function resulting in a value), use them as a stepping stone to components (a more complex concept).
* Provide more a more idiomatic description of routes, and their usage
* Defer Model and Routes, the two less likely to be used facets, to the bottom
* Bump all commentary on controllers to the controller section

TODO:

* [ ] ~~Services should likely be introduced~~ Nope, deferring to keep things shallow here.
* [x] Routes section inaccurately tells the user about default behavior, and basically repeats the end of the models section.

Split from https://github.com/emberjs/guides/pull/556

### DIFF
--- a/source/getting-started/core-concepts.md
+++ b/source/getting-started/core-concepts.md
@@ -1,58 +1,95 @@
-To get started with Ember.js, there are a few core concepts you
-should understand.
+Ember.js is a big framework with a large API surface. Learning a few core
+concepts will help you be successful understanding more detailed sections of
+the guides.
 
 ## Templates
 
-Templates, written in the Handlebars language, describe the user interface of
-your application. In addition to plain HTML, templates can contain expressions,
-like `{{title}}` or `{{author}}`, which take information from a component or
-controller and put it into HTML. They can also contain helpers, such as
-`{{#if isAdmin}}30 people have viewed your blog today.{{/if}}.` Finally, they
-can contain components such as a template listing blog posts rendering a
-component for each post.
+Ember.js uses templates to organize the layout of HTML in an application. Most
+templates in an Ember codebase are instantly familiar, and look like any
+fragment of HTML. For example:
+
+```handlebars
+<div>Hi, this is a valid Ember template!</div>
+```
+
+Ember templates use the syntax of [Handlebars](http://handlebarsjs.com)
+templates. Anything that is valid Handlebars syntax is valid Ember syntax.
+
+Each template has a context, and properties you access are read off that context.
+For example this template shows a welcome message with a name:
+
+```handlebars
+<div>Hi {{name}}, this is a valid Ember template!</div>
+```
+
+The Handlebars expression `{{name}}` refers to a bound property.
+Besides properties, double curly braces (`{{}}`) may also contain
+helpers and components.
 
 ## Components
 
 Components are the primary way user interfaces are organized in Ember. They
-consist of two parts: a template, and a source file written in JavaScript that
-defines the component's behavior. For example, a blog application might have a
-component for displaying a list of blog posts called `all-posts`, and another
-component for displaying an individual post called `view-post`. If users can
-upvote a post, the `view-post` component might define a behavior like _when the
-user clicks the upvote button, increase the `vote` property's value by 1_.
+consist of two parts:
 
-## Controllers
+* A template, for example `app/templates/components/welcome-message.hbs`
+* Optionally a source file written in JavaScript, for example
+  `app/components/welcome-message.js`.
+* **Components must contain a - in their name**. More about that later.
 
-Controllers are very much like components, so much so that in future versions of
-Ember, controllers will be replaced entirely with components. At the moment,
-components cannot be routed to (see below), but when this changes, it will be
-recommended to replace all controllers with components.
+Components differ from helpers in that they can output not only a value, but
+also manage DOM. An important property of components is that they are **isolated**.
+For example, a component's DOM is called "shadow" DOM.
+
+In this example, a component's template adds a div around some other DOM:
+
+```app/templates/components/my-component.hbs
+<div class="inner">
+  {{yield}}
+</div>
+```
+
+```handlebars
+<span>Hello</span>
+<div class="outer">
+  {{#my-component}}
+    Hello there!
+  {{/my-component}}
+</div>
+```
+
+Because components deal with DOM, they are also responsible for maintaining
+the bulk of an application's state. Additionally they handle events like
+clicks, keyUp, and other more semantic actions.
+
+## The Router and Routes
+
+The router maps a URL to a template. For example, when a user visits the `/posts`
+URL, a configured router may load the `posts` route and template. The router can also load
+nested routes. For example, if a blogging app had a list of blog posts on the
+left of the screen and showed the current blog post on the right, we'd say
+that the `posts/show` route was nested inside the `posts` route.
+
+Perhaps the most important thing to remember about Ember is that **the URL drives
+the application state**. URLs determine what template and route to load, and
+that in turn determines what model must be fetched.
+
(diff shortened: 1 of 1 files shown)


source/controllers/index.md
```diff
@@ -1,14 +1,23 @@
 ## Controllers
```
michaelrkn: The changes to this file are great - if you break this into its own PR, I'd love to get it merged in straight away.

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
-To get started with Ember.js, there are a few core concepts you
-should understand.
+Before moving to more focused topics, it is important to review several
```
michaelrkn: "review" - this isn't a review, it's the first time they've seen these concepts.

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
...
+templates. Anything that is valid Handlebars syntax is valid Ember syntax.
+
+Each template has a context, and properties you access are read off that context.
+For example this template shows a welcome message with a name:
+
+```handlebars
+<div>Hi {{name}}, this is a valid Ember template!</div>
+```
+
+The mustache of this template (the `{{}}`) refers to some dynamic content.
```
michaelrkn: "mustache of" --> "curly braces in"

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
...
+fragment of HTML. For example:
+
+```handlebars
+<div>Hi, this is a valid Ember template!</div>
+```
+
+Ember templates use the syntax of [Handlebars](http://handlebarsjs.com)
+templates. Anything that is valid Handlebars syntax is valid Ember syntax.
+
+Each template has a context, and properties you access are read off that context.
```
michaelrkn: > Each template has a context, and properties you access are read off that context.

I think this is too much detail for the Getting Started page. This page should be as high-level as possible - ideas like "context" and "properties" belong in the Templates section. I'd like to see this page introduce the _concept_ of templates as concisely as possible. Otherwise, it's just too much information to absorb for somebody who is seeing these concepts for the first time.

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
...
+For example this template shows a welcome message with a name:
+
+```handlebars
+<div>Hi {{name}}, this is a valid Ember template!</div>
+```
+
+The mustache of this template (the `{{}}`) refers to some dynamic content.
+Besides properties, you can also use helpers and components inside a mustache.
+
+## Helpers
```
michaelrkn: I think this section is too much for _Core Concepts_. Helpers are not a _core_ concept in my opinion - they're a detail of templates that the reader will learn about when they dive deeper into the Templates section. Again, I really want to avoid information overload on this page, and keep it as high-level as possible.

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
...
-component for displaying an individual post called `view-post`. If users can
-upvote a post, the `view-post` component might define a behavior like _when the
-user clicks the upvote button, increase the `vote` property's value by 1_.
+consist of two parts:
 
-## Controllers
+* A template, for example `app/templates/components/welcome-message.hbs`
+* Optionally a source file written in JavaScript, for example
+  `app/components/welcome-message.js`.
+* **Components must contain a - in their name**. More about that later.
```
michaelrkn: Too much detail.

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
...
 
-Controllers are very much like components, so much so that in future versions of
-Ember, controllers will be replaced entirely with components. At the moment,
-components cannot be routed to (see below), but when this changes, it will be
-recommended to replace all controllers with components.
+Components differ from helpers in that they can output not only a value, but
+also manage DOM. An important property of components is that they are **isolated**.
+For example, a component's DOM is called "shadow" DOM.
```
michaelrkn: 59-61 - too much detail.

source/getting-started/core-concepts.md
```diff
@@ -1,58 +1,120 @@
...
+## Models
+
+Models represent persistent state. For example, a blog application would
+persist the content of a blog post when a user publishes it. A blog post model
+would represent the state of that persisted content (such as the post itself,
+the title, and the author).
+
+A model
+typically persists information to a server, although models can be configured to
+save to anywhere. For example to the localStorage API, or IndexedDB.
```
michaelrkn: > For example to the localStorage API, or IndexedDB.

This is not a complete sentence.

source/getting-started/core-concepts.md
greyhwndz: I think @michaelrkn is making the initial dive easily 'digestible' and that is a good thing. I agree Helpers can be in another 'level' of journey. A well understood journey on the core concepts would provide a good foundation in tackling further into Helpers and then Components later

locks:
For reference: http://emberigniter.com/5-essential-ember-2.0-concepts/

michaelrkn:
@mixonic can you break out the changes to controllers/index into its own PR so we can get that merged?

mixonic:
I hope I can just make the needed changes to the rest of this and get it 
mergable.

michaelrkn:
That would be nice too, but 1) they are discrete changes and so don't really belong in the same commit, and 2) I'd like to do some work on the Controllers introduction and having your changes already in there will make it easier for me and help us avoid merge conflicts.

source/controllers/index.md
mixonic: Split into https://github.com/emberjs/guides/pull/717.

source/getting-started/core-concepts.md
mixonic: "Handlebars expression" is the phrase we agreed on in the templates section. Updating to this.

mixonic: > The Handlebars expression `{{name}}` refers to a bound property. Besides properties, double curly braces (`{{}}`) may also contain helpers and components.

source/getting-started/core-concepts.md
mixonic: Components are where we spent 90% of our time developing in most Ember applications. Already experienced users do not leverage helpers because the learned to avoid a volatile and poorly documented part of the API.

I strongly disagree that these concepts should be kicked off core concept entirely.

In many ways, helpers are simpler to understand that components. Like a bound property, the result in a single value being passed. They also provide a great context to begin understanding nested helper invocation, which is not useful this for core concepts but is very useful later on.

mixonic:
I've dropped the section on helpers. Headers are:

* Templates
* Components
* The Router and Routes
* Models

mixonic:
I've rebased again.

This (heavily edited) version of what I first proposed for core concepts I hope represents a good compromise between my intent and the feedback of several people. There are still some sections that you think are too detailed @michaelrkn, however I maintain that when you view them as a rendered page instead of reading the diff they are far, far more simple to understand than the current wall of text.

There are only four major sections:

* Templates
* Components
* The Router and Routes
* Models

Seeing as there are now other rewrites coming in https://github.com/emberjs/guides/pull/743 I'd love to see this rejected or accepted.

michaelrkn:
@locks and I discussed the tutorial a bit today. Since @kellywerks is owning the first half of the tutorial, which includes the Core Concepts guide, we're going to defer to her to figure out how best to incorporate both of your ideas. I know she's pretty swamped this week, so apologies for continuing the holding pattern.

PR #670, opened 2015-08-28
Add section about the `value` option of the action helper

balinterdi:
I took a stab of explaining what the `value` option of the action helper is and what it is good for. Let me know if any corrections need to be made.

### DIFF
--- a/source/templates/actions.md
+++ b/source/templates/actions.md
@@ -184,7 +184,7 @@ This way the `{{action}}` will fire when clicking with the alt key
 pressed down.
 
 ### Default Event
-By default, `event.preventDefault()` is called on all events handled by `{{action}}` helpers. To avoid this you can add `preventDefault=false` as a parameter. 
+By default, `event.preventDefault()` is called on all events handled by `{{action}}` helpers. To avoid this you can add `preventDefault=false` as a parameter.
 
 ### Stopping Event Propagation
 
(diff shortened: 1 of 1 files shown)


source/templates/actions.md
```diff
@@ -255,3 +255,41 @@ will trigger the action, but the user will remain on the current page.
...
+### Modifying the action's first parameter
+
+If a `value` option for the `{{action}}` helper is specified, its value will be
+considered a property path that will be read off of the first parameter of the
+action. This comes very handy with event listeners and enables to work with
+one-way bindings.
+
+```handlebars
+<label>What's your favorite band?</label>
+<input type="text" value={{favoriteBand}} onblur={{action "bandDidChange"}}/>
```
locks: are you sure you're supposed to do `value={{favoriteBand}}`?

balinterdi: Not totally, but I think it enables various things as the input value and the property (`favoriteBand`) are no longer two-way bound and thus you could do some filtering/transformation before you "write back" the value. This is silly but hopefully demonstrates my point:

```handlebars
<input type="text" value={{rating}} onblur={{action "clickRating" value="target.value"}}
```

```js
  actions: {
    clickRating(value) {
      const newValue = window.parseInt(value) + 1;
      this.set('rating', newValue);
    }
  }
```

When you focus out of the text field, the number you inputted will be incremented by one.

locks: What do you mean they're no longer two-way bound? I think I'm missing something here

rwjblue: This is `<input>` with attribute bindings **NOT** `{{input}}`.  Attribute bindings are one way (`{{input}}` gets around this by using the `change` event and updating the internal value).

balinterdi: If you type something in the input box, `rating` does not get updated right away to the inputted value, only when you the input loses focus (in other words, when the action is fired). It is somewhat hard to get one's head around but maybe because we're too used to two-way bindings.

balinterdi: @rwjblue So is the code example I gave demonstrative enough? I originally used `oninput`, not `onblur` but I guess both could work.

locks:
I'm convinced :P

balinterdi:
Thank you for merging, Ricardo.

PR #674, opened 2015-08-29
Update fullName property with guide's standard

zacharywelch:
All other examples of fullName follow this pattern

### DIFF
--- a/source/object-model/observers.md
+++ b/source/object-model/observers.md
@@ -17,10 +17,7 @@ Person = Ember.Object.extend({
   lastName: null,
 
   fullName: Ember.computed('firstName', 'lastName', function() {
-    var firstName = this.get('firstName');
-    var lastName = this.get('lastName');
-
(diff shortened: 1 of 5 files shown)


source/object-model/observers.md
```diff
@@ -17,10 +17,7 @@ Person = Ember.Object.extend({
   lastName: null,
 
   fullName: Ember.computed('firstName', 'lastName', function() {
-    var firstName = this.get('firstName');
-    var lastName = this.get('lastName');
-
-    return firstName + ' ' + lastName;
+    return this.get('firstName') + ' ' + this.get('lastName');
```
locks: Could you make it ```return `${firstName} ${lastName}`;``` instead, and update other examples to that?

zacharywelch:
:smile:

locks:
We're pushing ES2015 in the guides, so even if it's trickier at first it's where we're headed :)

PR #676, opened 2015-08-29
[Components] Updated code examples to be explicit in importing ember

greyhwndz:


locks:
There has been a discussion around this already, and we are eliding `import` statements until the new modules land.

PR #683, opened 2015-08-31
Clean up templates introduction

michaelrkn:
* Improves introduction of helpers by moving the content from Conditionals to Handlebars Basics.
* Removes redundant information on how templates get loaded, since it is covered in Routing.
* Moves Routing before Templates, so that readers know how templates are loaded and we don't have to cover it twice.

### DIFF
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -35,6 +35,28 @@
     - title: "Enumerables"
       url: 'enumerables'
 
+- title: "Routing"
+  url: 'routing'
+  pages:
+    - title: "Introduction"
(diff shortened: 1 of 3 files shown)


locks:
Squash please, looks good.

michaelrkn:
Not sure squashing is appropriate, since they are logically separate changes.

locks:
Gotcha, merging :)

locks:
Grr, three whole commits. :trollface:

michaelrkn:
mwahahaha

PR #684, opened 2015-08-31
Fix broken JS TOC expand/collapse

michaelrkn:
Removing the back to top library caused re-expanding the TOC sections to break. This removes the references to the back to top library and fixes re-expanding.


locks:
Thank you for contributing to the guides, Michael.

michaelrkn:
You are most welcome, Mr. Locks.

Issue #681, opened 2015-08-31
Service eager injection section is incorrect


foxnewsnetwork:
To content at:
http://guides.emberjs.com/v2.0.0/services/#toc_injecting-services-eagerly

Suggests that:
`app/instance-initializers/whatever.js`
```js
application.inject('route', 'cart', 'service:shopping-cart');
```

Is the proper way of eagerly doing a service injection.

This is incorrect because the instance of app passed to an instance-initializer doesn't have the `inject/3` method.

The correct way of doing this is with the regular initializer like so:

`app/initializers/whatever.js`
```js
(container, application) => application.inject("route", "cart", "service:shopping-cart");
```

If no one else is already working on it and EmberJS isn't planning changing up the initializer stack so that instance-initializers have inject, I would like make the correction to the guides and submit a PR.

rwjblue:
@foxnewsnetwork - The goal of initializers is to "set things up so that when they are looked up, they can be instantiated properly", so `inject` is definitely supposed to be done by initializers (not instance initializers).

foxnewsnetwork:
got it, I will submit pr momentarily.

PR #685, opened 2015-08-31
Instance-initializer is used in place of initializer in service injection

foxnewsnetwork:
In reference this issue:

https://github.com/emberjs/guides/issues/681#issuecomment-136225799

I changed changed "instance-initializer" to "initializer", made it a little more explicitly exactly what got passed to an initializer function, and added an example on when an instance-initializer is used.


locks:
@rwjblue requesting review

rwjblue:
@locks - Technical content LGTM.

locks:
:boom: :hammer:

PR #687, opened 2015-08-31
Tweaks "Defining Models"

locks:
Remove reference to script tags.
It is one of the default dependencies of an ember-cli project, so this
is unnecessary.

### DIFF
--- a/source/models/defining-models.md
+++ b/source/models/defining-models.md
@@ -3,55 +3,30 @@ data that you present to the user. Anything that the user expects to see
 if they leave your app and come back later (or if they refresh the page)
 should be represented by a model.
 
-Make sure to include `ember-data.js` after `ember.js`
-
-```html
-<script type="text/javascript" src="ember.js"></script>
-<script type="text/javascript" src="ember-data.js"></script>
-```
-
-For every model in your application, create a subclass of `DS.Model`:
+When you want a new model for your application you need to create a new file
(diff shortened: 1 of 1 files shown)


source/models/defining-models.md
```diff
@@ -180,6 +174,18 @@ export default DS.Model.extend({
 });
 ```
 
+#### Synchronous relationships
+
+Starting with Ember Data 2.0 all relationships are asynchronous by default. If
```
michaelrkn: "asynchronous" isn't defined. Maybe something like "Calling a relationship, like `myPost.get('comments')`, returns a promise. If you want the value to be returned instead, you need to pass an `async` option set to `false`:".

Or maybe we should just omit, this, since isn't it generally a bad idea and not something most people would want to do?

michaelrkn:
What do you think about taking out the bit that says "The table below shows how model names map to model file paths." and the following table?

source/models/defining-models.md
locks: my thing here is that it's a breaking update, and one that might go unnoticed if not mentioned explicitly. I think I'd rather add the explanation then.

michaelrkn: Isn't that better suited for a blog post or deprecation guide?

locks: hhmmm… ok

PR #696, opened 2015-09-02
Move multiple model info from controllers to routes

michaelrkn:
The rest of the page is covered by object-model/computed-properties-and-aggregate-data and not really controller-specific, so I deleted the whole thing.


josemarluedke:
:+1:

PR #697, opened 2015-09-02
Remove redundant before/aftermodel info from async routing

michaelrkn:


greyhwndz:
@michaelrkn: is this being transferred to somewhere or totally being scrapped out?

michaelrkn:
the most common use cases are covered in routing/redirection

greyhwndz:
Thanks @michaelrkn

PR #699, opened 2015-09-02
Simplify loading and errors page

michaelrkn:
I moved it after "Preventing and Retrying Transitions" since that guide makes sense to be right after "Redirecting".


greyhwndz:
@michael: I dunno if it would help but if there are examples that fits the cookbook then maybe encouraging them to come up with one and submit it as a PR to the cookbooks. Perhaps they can better come up with a more generalized example for the guides and point to the cookbook for some detailed examples.

PR #707, opened 2015-09-06
Improve wording of paragraph introducing CP setters

eoinkelly:
I wanted the paragraph to provide a more succinct summary of how you would enable _setting_ of CP.

I realise this is a minor change but it bothered me as I was reading the guide so I thought it was better to PR than not. No big deal if it is not enough of an improvement to merge.


michaelrkn:
Hey, thanks for contributing! "Enable getting and setting" - getting is always enabled, so this isn't really true. I personally find the wording less clear in the PR, and I don't think it's much of a difference either way, so I'm going to close this. But feel free to keep PRs coming if there are other things you think could be clearer!

PR #713, opened 2015-09-09
Update model guide page to ES6 style.

rengokantai:
like this: http://guides.emberjs.com/v2.0.0/models/#toc_models


michaelrkn:
I don't think this is a good idea. `ember generate model` creates:

```js
import DS from 'ember-data';

});
```

We want our code samples to reflect what Ember CLI generates.

If I've missed something, feel free to comment and/or re-open.

PR #717, opened 2015-09-11
Edits to the Controllers intro

mixonic:
@michaelrkn edits to the controllers alone as requested.


michaelrkn:
Awesome! Thank you!

PR #710, opened 2015-09-07
Inspecting routes guide review

mattw59:

PR #704, opened 2015-09-03
Inspector library info guide review

mattw59:
addresses #602

PR #693, opened 2015-09-01
Object inspector guide review

mattw59:
Addresses part of #602

PR #718, opened 2015-09-11
style guide review of Object Inspector, Routes, Libraries

mattw59:
Replaces #693, #704, #710


michaelrkn:
Thanks for your good work here!

PR #720, opened 2015-09-11
Confusing to me to say a sample component would look like a generate command

SERVANT14:


locks:
I'm pretty sure `ember generate` is introduced before, so that part is unnecessary.

SERVANT14:
Created a new pull request removing the generate command.

PR #723, opened 2015-09-13
Fix broken link to dependency injection guide

MaximAbramchuck:
While info about Dependency Injection in new Ember > 2.0 is not ready yet, I think that we need to route users to the oldest one just to understand concepts.


michaelrkn:
@locks how is the DI guide coming along?

michaelrkn:
#729 brings back the info, but I don't know if there's a good way to link to it from here. Since controllers are going away shortly, I would say either find a good place to link to once #729 gets merged, or delete this paragraph entirely.

MaximAbramchuck:
@michaelrkn yea, we can wait while #729, but it will be not so soon, so let's redirect user's to old guide for now. it is better than 404.

michaelrkn:
It doesn't matter if we change it here, because it won't get updated until the guides are de-deployed, and they won't be re-deployed before #729 is merged.

michaelrkn:
Oh, this is on an outdated branch, which we be replaced by master shortly.

PR #726, opened 2015-09-13
Removing unnecessary generate command in example

SERVANT14:
Having the command there was confusing to me because we're supposed to be seeing a sample component, not learning how to generate one.


locks:
It might have been an oversight from someone's edit, because it looks confusing to me as well. Thanks!

PR #728, opened 2015-09-13
Remove errorenus commas from each-in helpers

alexanderchr:
http://emberjs.com/api/classes/Ember.Templates.helpers.html#method_each-in


locks:
Thanks!

PR #729, opened 2015-09-14
Improve Services section

### LINKED ISSUE #599: Add info on dependency injection and service lookup
http://guides.emberjs.com/v1.13.0/understanding-ember/dependency-injection-and-service-lookup/ was removed as part of adding the Services section in 2.0.0. Some of the content didn't make it over. Let's make sure that all of the content is added back.

michaelrkn:
* Add info about initializers
* Restore content about DI and service lookup

Hat tip to @johnmcdowall's nice blog post on initializers for highlighting the important areas to cover on initializers and for a couple examples I used in here.

Fixes #599.

### DIFF
--- a/source/services-and-initializers/container.md
+++ b/source/services-and-initializers/container.md
@@ -1,0 +1,95 @@
+When an Ember application starts running, it will create and use a single
+instance of the `Ember.Container` object. This container object is responsible
+for managing factories and the dependencies between them. At the level of the
+container, a factory can be any part of the framework, like _route_ or
+_template_. For example, the index template is a factory with the name
+`template:index`, and the application route is a factory with the name
+`route:application`. The container understands how to use these factories
+(Are they singleton? Should they be instantiated?) and manages their
+dependencies.
+
+Factory names have two parts segmented by a `:`. The first segment is the
+framework factory type, and the second is the name of the factory requested.
+Hence, the `index` template would be named `template:index`. Ember has several
+built-in factory types, and you can also create your own by simply naming your
+factory appropriately. For example, to create a `user` type, you'd simply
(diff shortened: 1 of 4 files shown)


source/services-and-initializers/container.md
```diff
@@ -0,0 +1,95 @@
...
+for managing factories and the dependencies between them. At the level of the
+container, a factory can be any part of the framework, like _route_ or
+_template_. For example, the index template is a factory with the name
+`template:index`, and the application route is a factory with the name
+`route:application`. The container understands how to use these factories
+(Are they singleton? Should they be instantiated?) and manages their
+dependencies.
+
+Factory names have two parts segmented by a `:`. The first segment is the
+framework factory type, and the second is the name of the factory requested.
```
michaelrkn: "framework factory type" - is this the correct phrase?

source/services-and-initializers/container.md
```diff
@@ -0,0 +1,95 @@
...
+export function initialize(container, application) {
+  var Logger = Ember.Object.extend({
+    log(m) {
+      console.log(m);
+    }
+  });
+
+  application.register('logger:main', Logger);
+}
+```
```
michaelrkn: I worry this is a bad example, because `Logger` should be a service, shouldn't it? Is there any example of something an initializer would register that shouldn't be a service?

knownasilya: Maybe if you want to inject third party libs, without using them as globals, like piwik's tracking object, i.e. `_paq` by default, but by injecting you can easily rename it and see where it's injected. I wouldn't say this is a strong case for not using a service, since you could us the service as a wrapper.

Issue #600, opened 2015-08-17
Flip version ordering in the version drop down 


acorncom:
Over time it seems to make more sense to have older versions of the guides at the bottom of the versions menu instead of at the top ...

locks:
I don't have much of an opinion on this. @michaelrkn ?

michaelrkn:
Sounds good to me. I can take care of this before long.

locks:
@michaelrkn I volunteer @acorncom for the task :trollface:

michaelrkn:
@acorncom if you want to, the relevant code is in https://github.com/emberjs/guides.emberjs.com/blob/master/tasks/update-versions.js. Keep in mind that the change will only be for future versions - versions already released will have the old order.

locks:
Yup, that's my one opinion on it (I have one after all). It will be confusing for someone changing between versions.

acorncom:
I don't mind making the changes if we want to make them :-)

Re: versions already released having the old order ...  That's my one concern with making the change.  Is there no way to go back and change old versions?  I'm happy to do the leg work on the various pull requests if it's doable

acorncom:
> I volunteer @acorncom for the task :troll face:

And agreed, @michaelrkn, you've got bigger fish to fry, leave this low-hanging fruit for the rest of us ;-)

locks:
> Is there no way to go back and change old versions? I'm happy to do the leg work on the various pull requests if it's doable

Not currently feasible.

acorncom:
Think we should skip it then?  Seems like bad UX to flip things on folks.  On the other hand, better to do it now than when we're on 2.6+

locks:
I dream a dream that by `2.6` we'd have a different architecture.

acorncom:
:-)  So then should I send in a pull request?

locks:
@acorncom since this is non-critical and is a breaking ui change, I'd live to let it simmer a bit more.

acorncom:
:+1:

josemarluedke:
:+1

locks:
@josemarluedke are you agreeing that we should change or that we shouldn't change? and why?

josemarluedke:
Sorry, yeah, I think we should flip the order. For me makes more sense to have the newer versions on the top.

greyhwndz:
Having the recent version on top would be :+1:

michaelrkn:
Closed by #600.

PR #709, opened 2015-09-06
Implement warning when viewing old guides version

SaladFork:
_Addresses #506. A rebase of @acorncom's #563 which rebased @wagenet's #114. cc @michaelrkn_

##### Todo:
 - ~~How does `'CURRENT_REVISION_WILL_APPEAR_HERE_WHEN_DEPLOYED'` (in `version-select.js.erb`) work and how/why does it differ from `'<%= data.search.revision %>'` (in `versions.js.erb`)?~~
 - ~~How does `versions.json` get set when deploying?~~
 - ~~Would be nice to have documentation for the above~~
 - [x] Remove changes to `versions.json` and `version-select.js.erb` as we won't be supporting beta versions ([todo source](https://github.com/emberjs/guides/pull/563#issuecomment-129709603))
 - [x] Update code to handle `versions.json` format of `["v1.10.0", "v1.11.0", "v1.12.0", "v1.13.0"]`. ([todo source](https://github.com/emberjs/guides/pull/563#issuecomment-129709603))
 - [x] Clean up duplicate AJAX calls to `version.json` in `version-select.js.erb` and `outdated-warnings.js` ([todo source](https://github.com/emberjs/guides/pull/563#issuecomment-129709603))
 - [x] Add mechanism to allow user to jump straight to latest guide
   - Do we want to do this with an actionable button ("Jump to latest guides"), an informative link ("The latest guides are for vx.x.x"), or otherwise?
 - [x] Squash rebase prior to merge

##### Issues (advice welcome):
 - [x] Currently `outdated-warnings.js` and `version-select.js.erb` get concat'd before `versions.js.erb` (at least when running locally). This causes an issue as `window.GUIDE_VERSIONS` hasn't been defined yet.

PR #730, opened 2015-09-14
Flip version ordering in the version drop down

MaximAbramchuck:
Fixes https://github.com/emberjs/guides/issues/600


michaelrkn:
This won't work because #709 changes things dramatically with how this part of the code works. #709 also includes sorting, although I'm not sure which direction it sorts things.

michaelrkn:
@MaximAbramchuck FYI, issues without the "help wanted" label are likely not good to work on - there is usually a reason they aren't labelled that way. I hope this doesn't discourage you from getting involved - we'd love to have your help with anything labelled [help wanted](https://github.com/emberjs/guides/labels/help%20wanted)!

Issue #445, opened 2015-07-21
Add tutorial for building your first Ember app


michaelrkn:
Once upon a time, the guides opened with a tutorial that showed brand new Ember developers how to build an app from start to finish. As that tutorial became increasingly out of date, it was removed from the guides with the intent of replacing it with an up-to-date tutorial. Much time has passed, but at long last @poteto has spearheaded the creation of this tutorial. I (@michaelrkn) am helping her out, and we'll hopefully have something to show before too long!

ygrenzinger:
Meanwhile is there an up to date tutorial somewhere on the net ?

I have to create an internal app with EmberJS and most of the resources except the official docs seems outdated.

michaelrkn:
I'm not sure, but I'd start with something like [this search](https://www.google.com/search?q=ember+cli+tutorial&espv=2&biw=1360&bih=765&source=lnt&tbs=cdr%3A1%2Ccd_min%3A5%2F1%2F2015%2Ccd_max%3A&tbm=) (the date is arbitrary).

ygrenzinger:
Thanks ! Yes it gives better answer :)

locks:
@ygrenzinger check http://emberwatch.com/books.html for more suggestions, of the first four I'm not unfamiliar with Puri's but the others should be relatively up to date.

wldcordeiro:
Oops, didn't see this issue, #565 is a duplicate. Closing that and subscribing to this. Also willing to help out with writing a basic tutorial.

michaelrkn:
Thanks for the offer @wldcordeiro! Let me get back to you within the next few days.

wldcordeiro:
@michaelrkn no problem. Here's an idea of a TOC for the tutorial, I'm not super familiar with Ember myself but I think that could be useful when writing beginner tutorials.

* Installation
* Creating app and tour of folder structure
* Generating models and interacting with them in console
* Creation of routes and controllers
* Writing templates
* Writing Tests
* Basic styling

locks:
@wldcordeiro feels backwards to me to talk about models ahead of Routes, which is where everything falls from in Ember. I also don't like the idea of interacting with them in the console that much.

MarkMT:
Agree with that. Routes are more fundamental to the overall architecture than models.

greyhwndz:
[...]

PR #737, opened 2015-09-14
add application setup file to getting started

kellywerks:
Part of #445.


michaelrkn:
I think as part of this, you should delete `getting-started/first-app` since it covers the same content.

michaelrkn:
It looks like some other commits got folded into your PR - can you rebase so that this is just your commit?

kellywerks:
Done and done.

Do the titles need to make clear that they are part of a tutorial?
Titles like "Routes and Templates" are under "Getting Started" but don't
explicitly indicate that the content of these routes and templates is in
the context of a demo/tutorial ie distinct from more thorough Routes and
Templates content in other sections.  A thought...

Issue #328, opened 2015-05-27
Improve Loading / Error Substates page


Dremora:
[This page](http://guides.emberjs.com/v1.11.0/routing/loading-and-error-substates/) talks about `loading` event and has a call to `displayLoadingSpinner()` included in two of its snippets. This is exactly what I'm trying to do — however, there's no mention on how this function could be implemented (not to mention is should probably be a method). I can't figure out how to access controller from within `loading` event handler.

The documentation should be improved to provide relevant information. If it's impossible to change controller's state from the event handler, the mention of `displayLoadingSpinner()` should be removed.

michaelrkn:
Agreed. @Dremora did you figure out how to do this? If so, it would be awesome if you could open a PR to help explain it.

locks:
Getting a hold of the controller should be as simple as `this.controller`. The `displayLoadingSpinner` function is left as an exercise to the reader because it could be anything. It being a "global" function is an artefact from the pre-global days, I'd take suggestions as how to improve it.

Would something like:

```
export default Ember.Route.extend({
  displayLoadingSpinner() {
    this.controller.set('isLoading', true);
  },

  actions: {
    loading: function(transition, originRoute) {
      this.displayLoadingSpinner();

      // substate implementation when returning `true`
      return true;
    }
  }
});
```

solve your problem?

JKGisMe:
It also is not complete. According to this issue: https://github.com/emberjs/ember.js/issues/11888 the examples in the guides would never trigger the loading substate
```javascript
Router.map(function() {
  this.route('foo', function() {
    this.route('bar', function() {
      this.route('baz');
    });
  });
});
```

It would have to be:

```javascript
Router.map(function() {
  this.route('foo', function() {
    this.route('bar', function() {
      this.route('baz', function(){});
    });
  });
});
```

And according to this issue: https://github.com/emberjs/ember.js/issues/10645 when using pods the loading/error templates have to actually be put in separate directories which is never mentioned in the guides.

```
baz/controller.js
baz/route.js
baz/template.js
baz-loading/template.hbs
baz-error/template.hbs
```

As more people start to use pods this can be really hard to figure out. Not that I spent an inordinate amount of time trying to figure it out or anything.

And for the comment before mine... `this.controller` is private.

michaelrkn:
Fixed by #699.

PR #742, opened 2015-09-15
add pods loading substate

JKGisMe:
I have been thinking about this a lot and I don't think it's a good idea to ignore pods for this topic. I spent quite a lot of trying to figure out how to get the substates to work with pods and from talking about it in the Slack, I'm not the only one. Even some of the less nooby people had trouble getting this completely undocumented functionality to work. 

The current version also doesn't explain per-route custom loading routes which this addition does.

refs #328

### DIFF
--- a/source/routing/loading-and-error-substates.md
+++ b/source/routing/loading-and-error-substates.md
@@ -56,14 +56,18 @@ Router.map(function() {
   });
 });
 ```
+For projects using pods, a custom `loading` substate can be defined
+for each route by creating a directory, `foo/bar/slow-model-loading`, 
+with a template (and optional route) inside. 
 
-Ember will try to find a `loading` route in the hierarchy
-above `foo.bar.slow-model` that it can transition into, starting with
-`foo.bar.slow-models`'s sibling:
+Ember will automatically try to find a `loading` route that it can
+transition into in the following order:
 
-1. `foo.bar.loading`
-2. `foo.loading`
-3. `loading`
+1. `foo.bar.slow-model-loading`
+2. `foo.bar.loading`
+3. `foo.loading`
+4. `loading`
+
 
 ### The `loading` event
 


michaelrkn:
My understanding is that the pods structure isn't finalized (see http://emberjs.com/blog/2015/08/13/ember-2-0-released.html) and so isn't something we want to document in the Guides.

@locks is that your understanding?

JKGisMe:
Right now, for 2.0 and 2.1, the first place it looks is `foo.bar.slow-model-loading`... If that is changed in the future than shouldn't those guides reflect that new way while these guides reflect the current way?

michaelrkn:
From that blog post:

> Pods. We are also actively working on finalizing the pods directory layout for applications, which makes it easier to group related parts of your application together. This change will be completely opt-in, and existing layouts will continue to work.

I don't think changing little bits of the Guides to use pods is a good idea until pods are more finalized, at which point we will switch the whole Guides to use the pods structure.

elwayman02:
@michaelrkn generally I agree with you as well, but I think @JKGisMe has a point here. The released 2.0.1 version of Ember (to pick an example) isn't going to be changed...future versions may finalize pods more and update the behavior, but right now there is current behavior that is completely undocumented and has been causing problems for a number of people (it's come up on slack several times in the last couple weeks). Shouldn't that current behavior be documented since these are, in fact, versioned docs?

michaelrkn:
> Shouldn't that current behavior be documented

The purpose of the Guides is not to be a source of extensive documentation. Their purpose is to document the 20% of features that comprise 80% of Ember development (I'm making those numbers up for the purpose of expressing the principle). IMHO most Ember developers at this point will not find a single, narrow example of the pods structure useful. If you want to use pods at this point, that's fine, but the Ember Guides are not going to be the place for you to find out how. Once the pods format is finalized (I expect after routable components lands) and pods become the standard way to organize Ember apps, then we will switch the entire Guides to use that structure.

elwayman02:
:+1: I can get behind that philosophy. However, perhaps this behavior should be better documented in the actual Ember code so that it shows up in the API docs, if it is not already.

michaelrkn:
Yeah, I personally have no idea where that documentation is/should be.

JKGisMe:
Today, while talking to maybe the 3rd person of about 5 in the last 2 days who came into slack trying to get their loading substates to work I discovered this is how it works for _all_ 2.0+ (likely 1.13) projects whether they are using pods or not. I noted there were inconsistencies in Issue #328 I just have been using pods for so long(months) I thought it was pod specific. 

Here is a non-pods repo showing the behavior. 
https://github.com/JKGisMe/loadingExperiments

michaelrkn:
So you're saying that we just need to add [line 66](https://github.com/emberjs/guides/pull/742/files#diff-220ac094033fa3524d3f59a48fd10f2bR66), right? Can you open a PR for that?

PR #744, opened 2015-09-15
getting started tutorial models section

kellywerks:
Part of #445 

Note: Image in Github appears broken.  Raw code and Middleman view are correct.


source/getting-started/models.md
```diff
@@ -0,0 +1,66 @@
...
+}];
+
+export default Ember.Route.extend({
+  model() {
+    return rentals;
+  },
+});
+```
+
+If you haven't used the ES6 shorthand method definition syntax before, `model()` is the same as writing `model: function()`.
```
locks: link to explanation

source/getting-started/models.md
```diff
@@ -0,0 +1,66 @@
...
+
+When we add a method to a class as part of using the Ember framework, we often call it a `hook`. So here, we've added the model hook to our `index` route handler.
```
locks: is this definition sufficient? I'm uncertain it's clear that `model` is a method that Ember calls when you go to a route

PR #749, opened 2015-09-16
add ember-data with firebase to getting started

kellywerks:
Part of #445


data/pages.yml
```diff
@@ -10,8 +10,8 @@
   pages:
     - title: "Installing Ember"
       url: "index"
-    - title: "Your First Ember App"
-      url: "first-app"
```
michaelrkn: Why is this deleted in this commit? I think this was deleted in another PR, too.

locks:
Can you rebase?

PR #751, opened 2015-09-17
Added gotcha for homebrew users.

andrewstarks:
When using homebrew to install node, npm's modules were being loaded into `/usr/local/Cellar/node/0.12.7_1/libexec/npm
`.

This meant that ember could not be found. Searching the inter webs lead me to discover the incantations required to find the prefix, change it and then make it work as described by this text. Other, better solutions may be out there.

### DIFF
--- a/source/getting-started/index.md
+++ b/source/getting-started/index.md
@@ -55,6 +55,8 @@ Install Ember using npm:
 npm install -g ember-cli
 ```
 
+Mac users that used Homebrew to install node may wish to run `npm config get prefix`, prior installing ember-cli. If the prefix that this command returns is not in your path, then you can add it to your path or change the prefix with something like `npm config set prefix /usr/local`.
+ 
 While you're at it we recommend you also install PhantomJS to run tests from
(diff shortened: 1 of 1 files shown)


locks:
I don't think this should go in the guides, maybe later when we have the asides design? cc @michaelrkn

michaelrkn:
I agree. I think there could be dozens or hundreds of caveats about the installation process. I think that's outside the scope of the Guides.

andrewstarks:
Sg. I think, generally, if I go to a getting started page and I follow the friendly and pretty instructions and then it doesn't work, it's putting a bad light on the product at the moment I'm most likely to be judging it. 

I don't know if my experience generalizes to every homebrew/yosemite user, but if it does, perhaps removing the blurb about homebrew would be a better fix. If my experience is normal, then IMHO suggesting homebrew is ill-advised.

michaelrkn:
I've installed Ember on dozens of machines with Homebrew and haven't come across this problem.

PR #764, opened 2015-09-18
[WIP] Glimmer Components guides

chancancode:
## The Goal

Introduce the Glimmer Component feature to new and existing users.

The actual motivations of the feature itself should already be documented in the first few sections, but the high order bid is "component as an abstraction".

Once the feature has shipped (looking like past 2.2, so maybe 2.3? see below), my current thinking is that we can somehow keep the "legacy" components guides accessible but hidden by default. For example, on the top of the "Introduction" section, there will be a highly visible "Note: this is for new stuff, for old stuff, see here", and once you navigated into the old section, it will be available in the navigation.

## Philosophy

* Prioritize linear reading/initial learning
* Introduce concepts in small increments

For example, I chose to *not* mention yielding values to blocks when the concept of "blocks" was first introduced. (As opposed to mentioning "everything you need to know about blocks" in one section, which is better for someone who already learned the concepts and want to look something up again.)

## Table of Content

- [x] Introduction (motivation)
- [x] Refactoring Templates into Components
- [x] The Anatomy of a Component (syntax for defining and invoking components)
- [x] Passing Data to Components (`attrs` and blocks)
- [x] Adding Behavior and Keeping States (introducing the `GlimmerComponent` JS class and keeping internal states)
- [x] Handling User Interaction (internal to the component)
- [ ] Communicating With The Outside World (communicating with the parent by passing callbacks/closure actions, aka DDAU, `{{mut}}` helper)
- [ ] Reacting to Changes (computed properties, and lifecycle callbacks)
- [ ] Composing Components (a more fleshed out example with more than two levels of composition, probably based on the current examples, and "advanced" features like yielding values to blocks that are deliberately left-out from earlier chapters)
- [ ] Migrating From Legacy Components (a table of differences, enumerate benefits and steps to migrate)

## Current Status

The feature itself is pretty close to completion in the broad strokes, but there are plenty of nuances that needs to be fleshed out. Writing these guides was quite helpful in terms of uncovering holes in the model, and we have identified the right solutions for most of these things (but not implemented at the moment).

We also identified some performance issues, which we decided to prioritize over this feature, so we will likely not ship this until ~2.3.

In particular, I am currently semi-blocked on the "Communicating With The Outside World" section by one of these issues. (I already written some of the text, but haven't committed yet. Here is [the example](http://ember-twiddle.com/6537a168c78cd55083ab) I am planning to use.)

Since we are prioritizing performance for 2.2, this will probably be put back on the back burner for a bit until we pick up the GC feature again in a few weeks.

Grammar/spell-checks are not very high-priority at the moment (because entire sections could still be rewritten or scraped), but if you noticed anything, feel free to leave a line comment and I'll eventually deal with them.

High-level feedback (I don't like the flow here, I don't think you can assume knowledge about X here, I tried to learn it following the guides but I was confused about Y) would be quite helpful.

## Misc TODOs

- Fix known issues (top-level partials, recursive invocation, etc)
- Finalize the actual `import` syntax/path
- Update the Twiddle examples as things change
- Properties vs Attributes
- Overridable `update` in `{{mut}}`
- API docs

### DIFF
--- a/source/glimmer-components/adding-behavior-and-keeping-states.md
+++ b/source/glimmer-components/adding-behavior-and-keeping-states.md
@@ -1,0 +1,222 @@
+Most HTML elements fufills one or more of these purposes: describing the
+semantic significance of its content, altering their presentation and adding
+behavior.
+
+Take the `<article>` tag for instance: it has no default style rules, but it
+provides a semantic description its content. The `<b>` tag on the other hand,
+changes the presentation of its content (bolding the text) without conveying
+any semantic meanings. Of course, there are also many tags that serves both of
+these purposes at the same time, such as the `<ol>` and `<li>` pair. All the
+Components we have written so far fall into this category.
+
+Besides semantic and presentational elements, there are also HTML elements
+that add *behavior* for its content. For example, clicking on a `<label>` tag
+would move the focus to the corresponding input field; likewise, clicking on an
+`<a>` tag would navigate the browser to the URL specified in the `href`
+attribute.
+
+Ember allows us to add behavior to our Components using JavaScript. In
+additional to a Handlebars template, each Component can also have a
+corresponding JavaScript file in the `app/components/` directory.
+
+As an example, let's implement a `<random-color>` Component gives its content
+a solid fill or a "glow" effect in a different color everytime you click on it:
+
+```app/templates/application.hbs
+<random-color fill={{true}}>So</random-color>
+<random-color glow={{true}}>many</random-color>
+<random-color fill={{true}} glow={{true}}>colors</random-color>!!!
+```
+
+```app/templates/components/random-color.hbs
+<span>{{yield}}</span>
+```
+
+```app/components/random-color.js
+import Ember from 'ember';
+
+// Returns a random integer from 0 up to (and excluding) max
+function randomInteger(max) {
+  return Math.floor( Math.random() * max );
+}
+
+// Returns a random color in rgb(x,y,z) format
+function randomColor() {
+  let r = randomInteger(256),
+      g = randomInteger(256),
+      b = randomInteger(256);
+
+  return `rgb(${r},${g},${b})`;
+}
+
+// Returns a text-shadow for the "glow" effect
+function randomGlow() {
+  return `0px 0px 4px ${ randomColor() }`;
+}
+
+export default Ember.GlimmerComponent.extend({
+
+  setColor() {
+    if (this.attrs.fill) {
+      this.element.style.color = randomColor();
+    }
+
+    if (this.attrs.glow) {
+      this.element.style.textShadow = randomGlow();
+    }
+  },
+
+  didInsertElement() {
+    this.setColor();
+  },
+
+  click() {
+    this.setColor();
+  }
+
+});
+```
+
+You can see the result [at this Ember twiddle](http://ember-twiddle.com/e55043ce86b0a7fd2117).
+
+**TODO**: remove hax from twiddle
+
+**TODO**: fix ember-cli-shims and use `import Component from 'blah'` here
+
+There are quite a few things going on in this example, so let's break it down.
+Like usual, we created a template for our Component in `app/templates/components`,
+using a `<span>` as our top-level element and yielding any passed content. As
+promised, we also created a JavaScript file for our Component in `app/components`
+with a matching filename, `random-color.js`.
+
+In the JavaScript file, we exported a class that inherits `Ember.GlimmerComponent`,
+sometimes referred to as the *custom Component class*. We are required to
+export one such class whenever we define a Component JavaScript file.
+
+Inside our custom Component class, we defined a custom `setColor` method which
+assigns a random color and/or glow to the Component's element (in our case, the
+`<span>` tag) using standard [DOM APIs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style).
+Finally, we defined a `click` handler and implmented the `didInsertElement`
+hook to call our `setColor` method.
+
+While this is a rather silly example, it illustrated a few important things
+about a Component's JavaScript class:
+
+*   Each Component has a corresponding HTML element (the top-level element)
+    which can be accessed through its `element` property. (Technically, the
+    element is only available after the Component has been rendered; we will
+    get to that bit later.)
+
+    On a related note, if you prefer jQuery, Ember also provides the `$` method
+    as a jQuery proxy to the Component's element (it is roughly equivilant to
+    calling `$(this.element)`). For example, we can rewrite `setColor` with the
+    equivilant jQuery APIs, such as `this.$().css('color', randomColor())`.
+
+*   Components have access to the attributes via the `attrs` hash.
+
+*   Components can implement *event handlers* like `click`, `mouseEnter` and
+    `change` to handle UI events.
+
+*   Components can also implement *lifecycle hooks* like `didInsertElement`
+    and `willDestroyElement` to get notified at specific points during the
+    rendering process.
+
+We will get into the details of each of these point in subsequent sections, but
+if you are extra-curious, you can take a peek at [GlimmerComponent's API Docs](http://emberjs.com/api/classes/Ember.GlimmerComponent.html).
+
+**TODO**: write said API Docs ;)
+
+## Maintaining States
+
+Admittedly, our `<random-color>` Component isn't particularly useful or
+interesting. That's because we are missing a key ingredient to writing more
+complex components: states.
+
+Let's say we want to build a `<expandable-content>` Component. By default, the
+Component limits its content to a fixed height and renders a "Show more..."
+prompt. When clicked, the Component expands its content to their natural
+length. Clicking on it toggles the Component between the two modes.
+
+To do this, we would have to keep track of whether the Component is currently
+in the expanded mode or not. Since `GlimmerComponent` is a subclass of
+`Ember.Object`, we can just track this state as a property on the Component
+instance like usual:
+
+```app/templates/application.hbs
+<expandable-content>
+  <p>Lorem ipsum dolor sit amet, eu reque aperiam vix. Vel cu albucius
+  cotidieque, ut mel admodum vivendo. Paulo populo vivendo vix ...</p>
+
+  <p>Novum efficiendi eu ius...</p>
+
+  <p>Ut eos eloquentiam...</p>
+
+  <p>Dicta saperet perfecto...</p>
+
+  <p>Ne utroque fabellas...</p>
+</expandable-content>
+```
+
+```app/components/expandable-content.js
+import Ember from 'ember';
+
+export default Ember.GlimmerComponent.extend({
+  isExpanded: false,
+
+  click() {
+    this.toggleProperty('isExpanded');
+  }
+});
+```
+
+```app/templates/components/expandable-content.hbs
+<expandable-content class="{{if isExpanded 'expanded'}}">
+  {{yield}}
+
+  <button class="prompt">
+    {{#if isExpanded}}
+      Show Less...
+    {{else}}
+      Show More...
+    {{/if}}
+  </button>
+</expandable-content>
+```
+
+```css
+expandable-content {
+  display: block;
+  height: 200px;
+  overflow: hidden;
+}
+
+expandable-content.expanded {
+  height: auto;
+}
+
+/* More CSS to position the button, etc (see Ember Twiddle) */
+```
+
+You can see the complete example [at this Ember twiddle](http://ember-twiddle.com/42d98522232b59cb2e9c).
+
+There isn't anything too interesting on the invoking side: we simply wrap the
+content we want to display with our `<expandable-content>` Component.
+
+On the Component's JavaScript class, we added an `isExpanded` property and set
+its initial value to `false` (so our Componented starts off in the "folded"
+mode).
+
+Since we mostly relying on CSS to do the heavy lifting (showing and hiding the
+extra content), all we need to do is to add and remove the "expanded" CSS class
+on the top-level element depending on the `isExpanded` state and change the
+button's label accordingly.
+
+You might have observed that we were able to access the `isExpanded` property
+directly from the template. This is because the context of the template is
+the Component instance. As a result, all of its properties (including computed
+properties) are "in-scope" for the template.
+
+In fact, this is how we were able to access the `attrs` hash from the template
+all along – if we don't explicitly define a JavaScript class for our Component,
+Ember will automatically create an instance of the generic `GlimmerComponent`
+class and use it as the context to run our templates.
--- a/source/glimmer-components/handling-user-interaction.md
+++ b/source/glimmer-components/handling-user-interaction.md
@@ -1,0 +1,121 @@
+Components allow you to define UI controls that can be reused throughout
+your application. If they're generic enough, they can even be shared across
+multiple applications.
+
+In order to make a useful control, we would need to allow our users to
+interact with our Components and handle these interaction somehow. Let's
+look at two ways to handle UI events in Ember Components.
+
+## Event Handlers
+
+The first way to handle UI events in Components is to implement event handler
+methods on the Component's JavaScript class. We have already seen this in
+action in the previous section – the `click` method we implemented on both of
+our Components are an example of this.
+
+More generally, when an UI event fires inside your Component (i.e. on the top-
+level element itself or any of its children, including any yielded content),
+Ember would try to invoke the corresponding event handler method on the
+Component instance if one exists.
+
+You can refer to [GlimmerComponent's API Docs](http://emberjs.com/api/classes/Ember.GlimmerComponent.html)
+for the full list of available events.
+
+**TODO**: write said API Docs T___T
+
+Your event handler methods are called with the jQuery Event object as an
+argument, which allows you to supress the browser's default behavior for
+the event or inspect its target.
+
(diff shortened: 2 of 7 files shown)


locks:
Why are they a different section from the regular components?

michaelrkn:
Thanks for all your work on this, @chancancode! 

> my current thinking is that we can somehow keep the "legacy" components guides accessible but hidden by default.

I'm not really into that idea. The goal of the Guides is to show users the happy path to using Ember. When Glimmer components are released, they will be the happy path. I think we should just update (or replace) the current Components section with your content.

> Migrating From Legacy Components (a table of differences, enumerate benefits and steps to migrate)

I think this would make a very useful blog post, but I don't think it should go in the Guides, just like we don't include deprecation info in the Guides.

source/glimmer-components/passing-data-to-components.md
```diff
@@ -0,0 +1,231 @@
...
+You might be wondering: what's the point of going through the `attrs` hash?
+Couldn't Ember just expose the attributes as `{{type}}` instead of the nosier
+`{{attrs.type}}`? Hold your thoughts! We will come back to that soon.
+
+In addition to string attributes, sometimes you might need to pass other kind
+of data such as numbers and booleans. You can do this using the curlies syntax
+instead of quotation marks:
+
+```app/templates/step-two.hbs
+<progress-bar value={{2}} max={{10}} animated={{true}} /> Step 2 / 10
```
chadhietala: Why are curlies needed for primitive types?

source/glimmer-components/index.md
```diff
@@ -0,0 +1,37 @@
+## HTML elements as an abstraction for text documents
+
```
chadhietala: I think this introduction needs to talk about the programming model instead of components as high level abstractions for html.

source/glimmer-components/passing-data-to-components.md
wycats: Because `<progress-bar value=2></progress-bar>` is already defined in HTML as meaning:

```html
<progress-bar value="2"></progress-bar>
```

We can get away with some divergence, like adding support for `<progress-bar />`, in areas where the thing we're adding is very, very unused in practice. In the case of `<progress-bar />`, it is actually far more likely that the HTML definition ("ignore the `/`") is not what the user intended.

However, when someone writes `<progress-bar value=2>` in an HTML-like context, there is a very strong likelihood that they really do mean the HTML semantics, so diverging here is not acceptable.

wycats: The only other divergence from HTML is:

```hbs
<input checked={{isActive}}>
```

In HTML, attributes without quotes are still attributes, but in Ember, the above syntax assigns a property, not an attribute.

I think we can get away with this divergence because we are introducing new syntax that allows you to add dynamic content to HTML (`{{}}` is pretty much the consensus syntax for this extension). Because assigning both attributes and properties to HTML (especially with web components) is common, having a terse syntax for describing each is important.

Again, this restricts the divergence from "normal" HTML to:

* `/>` means what it means in XHTML
* `foo={{}}` without quotation marks around the curlies means assigning a property, not an attribute.

wycats: Note that the DOM spec mirrors almost all properties to attributes, so leaving off quotation marks will mostly do what you expect but allow you to pass values (booleans, dates, etc.) instead of strings. This is especially important for things like `value={{name}}`, whose attribute assigns a "default value", but the property updates the `.value`.

Because of the fact that attributes and properties are melded together in the user's mental model already, the DOM has done work to make attribute strings and property values mostly mirror. What this means is if we hew closely to the same fuzzy mental model that the DOM already worked hard to support, things will work with the same caveats as DOM.

source/glimmer-components/index.md
wycats: Can you say more what you mean?

There are certainly three audiences here:

* people familiar with HTML/CSS/JS
* people familiar with other web frameworks that have a component API
* people familiar with the current curly components in Ember

This starts off with a common denominator: all three groups know HTML.

wycats: Anything more belongs in dedicated guides: transition from Angular, transition from React, transition from Ember curlies.

wycats:
> Thanks for all your work on this, @chancancode!

>> my current thinking is that we can somehow keep the "legacy" components guides accessible but hidden by default.

> I'm not really into that idea. The goal of the Guides is to show users the happy path to using Ember. When Glimmer components are released, they will be the happy path. I think we should just update (or replace) the current Components section with your content.

The transition to Glimmer components will take a while, and immediately removing the guides out from under existing apps seems bad.

Up until the first release where we release Glimmer Components, curlies are the first-class way to do components in Ember. If they're working for you, they'll keep working, and there's no reason to make people feel like they have done something wrong for doing exactly what we told them to do.

I think we should use the "Classic" moniker for existing components ("Classic components") rather than "legacy", which indicates that there is a new, better way, but that you shouldn't feel embarrassed about not using the new way yet.

>> Migrating From Legacy Components (a table of differences, enumerate benefits and steps to migrate)

> I think this would make a very useful blog post, but I don't think it should go in the Guides, just like we don't include deprecation info in the Guides.

I disagree. We need a new top-level transitioning section to help people transition concepts from other frameworks like Angular and React, and a guide for transitioning from curly components to angle-bracket components would fit well there.

The blog is for content that, if you came back a year later, you would understand that the content is out of date, and that wouldn't be a big deal. For example, when we write a blog post talking about a release, or the progress towards a feature, nobody is going to get upset if a year later, the information in that blog post isn't precisely the current state of Ember.

However, content about transitioning to angle bracket components (either from curlies or from another framework) is meant to last a long time. As we change Ember's idioms, it is important to keep content like that up to date, so it does not belong on the blog.

wycats:
> Why are they a different section from the regular components?

This PR  is not meant to reflect the final structure, which is an orthogonal question that I have been discussing with @michaelrkn. @chancancode shipped it so that people can review the "new world" before we're ready to make calls about the exact rollout.

eccegordo:
Classic is a good way to describe this. I think for busy people there should be a high level summary table somewhere that shows the important differences side by side. Also might be helpful to show pre and post glimmer components example code side by side whenever possible. I think that will drive home some of the finer points.

source/glimmer-components/index.md
chadhietala: While I understand that this is supposed ease the way in to the components, but throughout these guides there are only touches on "why components" or "why is this model good". I feel like it might be worth stating why components and how they fit into the rest of the system. Maybe it doesn't need to be in this document, but I think should follow instead of going right into refactoring.

wycats: I think it makes sense for there to be a "Philosophy of Ember Components" guide that talks about the programming model at a high level in a non-tutorial style. Is that what you have in mind?

chadhietala: c

michaelrkn:
I feel pretty strongly about is not having 2 different sections for classic components and glimmer components. I think the documentation should primarily assume the user will use glimmer components, and then note differences between the two where appropriate. Otherwise, it will be confusing for people who are new to Ember as they try to parse out the differences between the two, and it will be difficult to maintain the information, as there is significant overlap between the features of classic and glimmer components.

> We need a new top-level transitioning section to help people transition concepts from other frameworks like Angular and React

Hm, that's an interesting idea.

> The blog is for content that, if you came back a year later, you would understand that the content is out of date, and that wouldn't be a big deal.

Transitioning from classic to glimmer components seems like a good candidate. I could be wrong, but my guess is that in a year from glimmer components' release, there will be a relatively small number of people who haven't switched, for whom the content is still relevant.

sandstrom:
> I think we should use the "Classic" moniker for existing components ("Classic components") rather than "legacy", which indicates that there is a new, better way, but that you shouldn't feel embarrassed about not using the new way yet.

I agree with this. Given the high rate of progress lately (which is overall good) it's wise to not tire people with deprecations.

chancancode:
:+1: on "Classic"

I think what might not be immediately obvious here is that the two styles of components end up diverging quite a bit, so I understand the reaction here. The original plan was to call this "angle bracket components", which is the same components we have to day except blah, blah and blah.

As the feature was being worked on, it ultimately become quite clear that there are a lot of important differences that basically end up requiring a pretty different mental model (we assert that it is simpler) and it also enables many scenarios that are not possible (or very annoying) before.

One thing to call out specifically that it has a different base class than classic components (`Ember.Component` vs `Ember.GlimmerComponent`) and over time. Part of it is for performance reasons, but there are already some divergence in their APIs and over time they will probably diverge more.

It might be helpful to think of the time when Components were originally introduced in Ember. It is certainly possible to explain and describe them as "same thing as Views, except blah and blah and blah",  overtime it has become quite obvious that those differences puts you in a vert different mental model and allows for very different use case, so it is more useful to talk about them as a separate feature that happens to share primitives with Views.

> Otherwise, it will be confusing for people who are new to Ember as they try to parse out the differences between the two

Which is why I suggested to keep it hidden (from navs) by default and only link to them from the banner in the introduction section. ("This guide is describing the new Glimmer Component feature introduced in 2.x; go [here](...) if you are looking for information about ..." – we can work on making this obvious that if you are a new user you won't have to care.) I think it is fine for the section to be named simply "Components". The current naming is simply because I couldn't move/hide the existing guides yet.

> and it will be difficult to maintain the information, as there is significant overlap between the features of classic and glimmer components.

It will probably be the contrary – the classic components API will be supported and maintained, but they probably won't get any new features. So it is more likely that they will be quite stable but the overlap will *shrink* significantly overtime.

> in a year from glimmer components' release, there will be a relatively small number of people who haven't switched, for whom the content is still relevant.

I think it is once again useful to think of the Views situation in 1.x. Since Components were introduced, people have slowly transitioned away from views, and if you are working on new ares in your apps, you probably shouldn't use views. But it wasn't an overnight transition, and it was a fully-supported feature for the 1.x series, so it is useful to have the docs still accessible.

I believe it would probably be true here as well, but perhaps to a slightly lesser extent. Especially if you consider add-ons that ships components and large apps, this transition will probably take a while. Also, as Yehuda have pointed out, there is really no immediate need to switch if what you have is working well today.

michaelrkn:
Great points, @chancancode. Thanks for the thorough explanation.

chancancode:
:heart: we difinitely share the same goals (optimize for not confusing new users), so we can keep toying around with different approaches on that front while the feature/content progress :)

michaelrkn:
Sounds good!

chadhietala:
I would like to assume that at 3.x `GlimmerComponent` → `Component`. Prefixing with `Glimmer` just seems like a bridge and is explicit during the transition. There should be no need for curly components once released.

A lot of people that I talk to refer to Ember as an MVC but when they start working with it they are like "Where are the Views?" and hopefully soon "Where are the Controllers?". The "big picture" ideas are outlined in blogs, but seem to have a hard time getting into the guides. This is sort of why I mentioned adding something about the programming model and methodology up front.

locks:
This should be less of an issue with the new module imports.

wycats:
> I would like to assume that at 3.x GlimmerComponent → Component. Prefixing with Glimmer just seems like a bridge and is explicit during the transition.

Indeed.

> There should be no need for curly components once released and would be deprecated at some point in 2.X.

Like views in 1.x, curly components handle a large number of use-cases simply by virtue of the fact that they have existed for a while and have accumulated subtle emergent properties that are the "canonical" solution to certain problems.

For example, `childViews` and `parentView` are used for self-registering components, but Glimmer Components will eventually have a better story for this.

We will probably migrate some of these features over to GlimmerComponent to aid in the transition, but it's likely that certain use-cases that subtly depend on particular combinations of features won't work immediately. As a result, I do not intend to "deprecate" classic components until we observe that people stop asking for features in GlimmerComponent to help in the migration.

> This should be less of an issue with the new module imports.

Confirm.

```js
import Component from 'ember/glimmer-component';

// now you can refer to just "Component" in the rest of the file
export default Component.extend({

});
```

source/glimmer-components/index.md
chancancode: I like that idea, but I have trouble fitting it into the flow. Since we are primarily targeting new users here, front-loading it at the beginning ("you might not even know what a component is yet, but let me tell you why you would love them and get you to buy in on our philosophy!") seems counterproductive.

Without getting into the specifics of the feature, the "abstraction" part *is* the "why components" and "why this model feels good" at the highest level – this is meant to more than just "ease into" the topic, it is meant to introduce the motivation. DDAU, isolation, etc – all boils down to "abstraction" and "refactorability" (encapsulation) at the end of the day (they tell you what your app is doing without immediately exposing the details, and as a result, you can also refactor the implementation all you want without affect the outside), which is what this section is trying to get at without diving into the specifics.

So I think there are two options:

1. Once the programming model has been fully introduced, it makes sense to call them out explicitly again, either as part of "Composing Components" or breaking it up into two sections ("Composing Components" -> "Thinking in Components" aka Thinking in React).

2. Have a top-level "Thinking in Ember" section at the end of the guides that groups all of these topics (again, once they have finished the tutorials).

Thoughts?

rwjblue:
> I would like to assume that at 3.x GlimmerComponent → Component. Prefixing with Glimmer just seems like a bridge and is explicit during the transition. There should be no need for curly components once released and would be deprecated at some point in 2.X.

I am not certain of this.

* Once a given path/name is used it is extremely hard to "take it back" to reference a completely different thing.
* There are things that angle bracket components will not attempt to do (`positionalParams` is one good example), there may still be a need for "classic components".
* It is not obvious to me that 3.x will remove "classic components".
* I am not aware of **any** plans to deprecate `Ember.Component` (aka "classic components").

chancancode:
We should probably save that bikeshed for the feature PR ;) I promise I will update the docs with whatever people eventually agreed on :P

rwjblue:
Agreed.

source/glimmer-components/adding-behavior-and-keeping-states.md
```diff
@@ -0,0 +1,222 @@
+Most HTML elements fufills one or more of these purposes: describing the
```
locks: `fulfil`

source/glimmer-components/adding-behavior-and-keeping-states.md
```diff
@@ -0,0 +1,222 @@
...
+that add *behavior* for its content. For example, clicking on a `<label>` tag
+would move the focus to the corresponding input field; likewise, clicking on an
+`<a>` tag would navigate the browser to the URL specified in the `href`
+attribute.
+
+Ember allows us to add behavior to our Components using JavaScript. In
+additional to a Handlebars template, each Component can also have a
+corresponding JavaScript file in the `app/components/` directory.
+
+As an example, let's implement a `<random-color>` Component gives its content
```
oskarrough: … component _that_ gives?

source/glimmer-components/adding-behavior-and-keeping-states.md
acorncom: `fulfill` ;-)

source/glimmer-components/handling-user-interaction.md
```diff
@@ -0,0 +1,121 @@
+Components allow you to define UI controls that can be reused throughout
+your application. If they're generic enough, they can even be shared across
+multiple applications.
+
+In order to make a useful control, we would need to allow our users to
+interact with our Components and handle these interaction somehow. Let's
```
acorncom: This may be nit-picky too early, but seems like `Component` is capitalized throughout these docs even when not at the beginning of a sentence.  That seems different from the way we use it in the current docs ...

source/glimmer-components/handling-user-interaction.md
```diff
@@ -0,0 +1,121 @@
...
+    if (href && pattern.test(href)) {
+      alert(this.attrs.message || "Link blocked!");
+      event.preventDefault();
+    }
+  }
+
+});
+```
+
+You can try this Component in action over [at this Ember twiddle](https://gist.github.com/chancancode/a92d7a68d438d7acdfa2).
```
acorncom: Twiddle link instead of Gist?

source/glimmer-components/adding-behavior-and-keeping-states.md
```diff
@@ -0,0 +1,222 @@
+Most HTML elements fufills one or more of these purposes: describing the
+semantic significance of its content, altering their presentation and adding
+behavior.
+
+Take the `<article>` tag for instance: it has no default style rules, but it
+provides a semantic description its content. The `<b>` tag on the other hand,
```
knownasilya: `semantic description of it's`?

source/glimmer-components/adding-behavior-and-keeping-states.md
```diff
@@ -0,0 +1,222 @@
...
+  },
+
+  click() {
+    this.setColor();
+  }
+
+});
+```
+
+You can see the result [at this Ember twiddle](http://ember-twiddle.com/e55043ce86b0a7fd2117).
```
locks: > Mixed Content: The page at 'about:srcdoc' was loaded over HTTPS, but requested an insecure script 'http://builds.emberjs.com/canary/ember.debug.js'. This request has been blocked; the content must be served over HTTPS.
VM385 about:srcdoc:1 Mixed Content: The page at 'about:srcdoc' was loaded over HTTPS, but requested an insecure script 'http://builds.emberjs.com/canary/ember-template-compiler.js'. This request has been blocked; the content must be served over HTTPS.

acorncom: Update the link to use `https://` and this all goes away ...

PR #765, opened 2015-09-18
Add missing backtick to component-hide-show-image.md

jyunderwood:
So code snippet for  `app/templates/components/rental-tile.hbs` shows as code.


michaelrkn:
Reverting in order to revert #750.

PR #767, opened 2015-09-19
Remove reference to DS.ActiveModelAdapter ref #2353

davidpett:
The issue https://github.com/emberjs/website/issues/2353 states the link goes nowhere, and the reason is that it was removed from ember-data here: https://github.com/emberjs/data/pull/3457. This removed the reference to `DS.ActiveModelAdapter` as a "buitin adapter"


davidpett:
wrong branch, needs to be on 1.12 as it is not present past that point

PR #768, opened 2015-09-19
removed reference to DS.ActiveModelAdapter

davidpett:
The issue emberjs/website#2353 states the link goes nowhere, and the reason is that it was removed from ember-data here: emberjs/data#3457. This removed the reference to DS.ActiveModelAdapter as a "buitin adapter"


michaelrkn:
Thanks for noting this problem! However, old versions of the Guides are considered static and immutable, and we don't have the bandwidth to maintain them.

davidpett:
no problem, can someone close https://github.com/emberjs/website/issues/2353 then, since it is not valid?

PR #772, opened 2015-09-19
Remove info on pausing boot from initializers

### LINKED ISSUE #771: Pausing the boot process example doesn't work
http://guides.emberjs.com/v2.0.0/services-and-initializers/initializers/#toc_pausing-the-boot-process

`container.lookup('store:main')` is not only private but doesn't work in 2.0. 

Can someone come up with another example to use here?

locks:
As mentioned in #771, I don't think it's possible anymore to pause during booting, so this example should be removed.

cc @wycats

### DIFF
--- a/source/services-and-initializers/initializers.md
+++ b/source/services-and-initializers/initializers.md
@@ -53,28 +53,3 @@ export default {
   }
 };
 ```
-
-## Pausing the Boot Process
-
-If you need to make sure that the rest of your application doesn't load until
-an initializer has finished loading, you can use the `deferReadiness` and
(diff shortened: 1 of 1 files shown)


michaelrkn:
@locks can you look at https://github.com/emberjs/guides/issues/771#issuecomment-141706212 and decide if that is more appropriate?

locks:
Should this be closed or merged meanwhile? @dgeb @michaelrkn

michaelrkn:
Based on https://github.com/emberjs/guides/issues/771#issuecomment-141706212 I would say closed, but I am no expert.

Issue #501, opened 2015-07-30
Update Test Components - Integration Test instead of Unit Test


jimbeaudoin:
Hi,

The section in the guides about Testing Components is using Unit Tests. Since 1.13 the correct way to test components is with an Integration Test right? Also, Ember CLI generate an integration test with a component. My english is not good enough to make a PR but I can create this issue. Sorry! :)

Have a nice day and thank you! :)

rwjblue:
Yes, using the integration test style described in the following articles is the right way forward:

* alisdair.mcdiarmid.org/2015/06/20/ember-component-integration-tests.html
* emberup.co/integration-tests-for-components/
* https://github.com/rwjblue/ember-qunit#component-integration-tests

We should definitely update the guide to support, though that *might* belong in a different issue (I can't recall if there is one already for this).

jimbeaudoin:
Ok, thanks for the links. I appreciate! :) I searched the issue tracker and the PRs but found nothing about this. This is why I decide to open this issue. But I think ther is a ember 2.0 guide in the making.

eccegordo:
As an aside it would be helpful to clearly discuss the side effects and statefulness of integration testing and explain the difference from unit testing now that there is that `integration: true` mode for testing components.

After trying to refactor several hundred component tests in my own project I am seeing a lot of of 

```
Uncaught Error: Assertion Failed: calling set on destroyed object
```

In other tests and through out my app. It seems that integration mode has exposed some underlying issues (I have a lot of data binding). And my hunch is that the integration mode is exposing some problems with the setup and tear down of state.

trek:
Will try to get a PR addressing this in the coming week

bantic:
@trek I'd be happy to write up a PR for this tonight if there's not one already in progress

lukesargeant:
Why are we refactoring component unit tests? Don't component integration tests perform a different role?

toddjordan:
At my day job, we are starting to incorporate integration-component tests as we develop new components.  Some things we are working through are how to deal with things like service dependencies, as well as convert/adapt our existing component unit tests to things like closure actions.  I've got time today and this weekend, so I'll start writing up a guide page.  I'll PR that into here if help is still wanted, but if anything it will help out the team I work on.

toddjordan:
A couple of notes on how I'm going about it... 
[...]

PR #774, opened 2015-09-19
Fix for Issue #501 Update examples in "Testing Components" guide to use Integration Tests

toddjordan:
See issue https://github.com/emberjs/guides/issues/501

I redid the same examples using integration tests.  I also got the examples working in a sample app to verify them.  I also included some other newer practices such as pods and some more es6 mechanisms.

There's enough changes here where you might want to look at the split diff

### DIFF
--- a/source/testing/testing-components.md
+++ b/source/testing/testing-components.md
@@ -1,7 +1,6 @@
-_Unit testing methods and computed properties follows previous patterns shown
-in [Unit Testing Basics] because Ember.Component extends Ember.Object._
 
-Components can be tested using the `moduleForComponent` helper.
+
+Components can be tested with integration tests using the `moduleForComponent` helper.
 
 Let's assume we have a component with a `style` property that is updated
 whenever the value for its `name` property changes. The `style` attribute of the
@@ -15,7 +14,8 @@ export default Ember.Component.extend({
   attributeBindings: ['style'],
 
   style: Ember.computed('name', function() {
-    return 'color: ' + this.get('name') + ';';
+    const name = this.get('name');
+    return `color: ${name}`;
   })
 });
 ```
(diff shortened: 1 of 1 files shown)


source/testing/testing-components.md
```diff
@@ -129,182 +111,69 @@ jQuery triggers can be used to simulate user interaction and test that the title
...
-
-  // Renders the component to the page
-  this.render();
-
-  // Submits the form
-  this.$().find('input[type="submit"]').click();
-});
-```
-
-<!---
```
toddjordan: Since this was already commented I went ahead and deleted.  I can work on bringing it back if we feel its important.  Now that we use integration tests I don't really think there's going to be much different about nested components that cause this section.  you'd just declare the components in the render template and assert the DOM

michaelrkn:
Thanks a bunch for picking this up, @toddjordan!

source/testing/testing-components.md
```diff
@@ -1,16 +1,15 @@
...
+
+Components can be tested with integration tests using the `moduleForComponent` helper.
 
 Let's assume we have a component with a `style` property that is updated
 whenever the value for its `name` property changes. The `style` attribute of the
 component is bound to its `style` property.
 
 > You can follow along by generating your own component with `ember generate
-> component pretty-color`.
+> component pretty-color --pod`.
```
michaelrkn: We're not going to encourage people to use the pod directory structure until it's format has settled down. Can you re-do the examples with the "regular" structure for now?

source/testing/testing-components.md
```diff
@@ -128,183 +110,70 @@ export default Ember.Component.extend({
...
-  assert.equal(finalTitle, 'Hello Ember World');
+  assert.equal(this.$('h2').text(), 'This is Magic', 'title changes after click');
 });
 ```
 
 ### Testing Actions
 
-Components often utilize the `sendAction()` method to send actions to other
-objects in your application.
+Components starting in Ember 2 utilize closure actions, which allow the component to directly invoke functions provided by components higher up in then tree.
```
michaelrkn: Can we simplify this to something like "If a component is nested within another component, the inner component can directly invoke functions provided by the outer component."?

source/testing/testing-components.md
```diff
@@ -128,183 +110,70 @@ export default Ember.Component.extend({
...
-  this.$();
-
-  // perform assertions
-  assert.equal(this.$().find('h3:contains("Cats")').length, 1);
-  assert.equal(this.$().find('img').length, 3);
-});
-```
--->
-
 [Unit Testing Basics]: ../unit-testing-basics
```
michaelrkn: This can be removed since the link is removed.

michaelrkn:
@locks can you review as well?

source/testing/testing-components.md
toddjordan: Absolutely.  I'll go back and update

toddjordan:
@michaelrkn I also have a new section on stubbing services using the example in my blog post here: 

http://presentationtier.com/stubbing-services-in-emberjs-integration-tests/

I'd like to add a new section to "Testing Components" based on the example, with language consistent with the others.  Should I add the section to this PR?

michaelrkn:
@toddjordan I'd say open a separate PR for that, after this is merged. It's easier and faster to review things individually. 

@locks Can you review this so that @toddjordan can move forward?

Thanks!

toddjordan:
will do, thx

mixonic:
@toddjordan I'm concerned with `this.container.lookup` in your test. I don't believe the container is considered public API. The rest seems reasonable, and there are ways to set values after lookup without using the container.

* Stash the instance during `init` of the stub and set against that
* Register a singleton object instead of a class for the service- may require some mucking about with the options for registering.

Nice writeup though!

toddjordan:
Thx for the tip.  I'll look into modifying the example to a lookup alternative.

mixonic:
@toddjordan Also note that Ember 2.1 has tweaked some of these APIs. You can find links to data about those changes in [this issue](https://github.com/emberjs/ember.js/issues/12336).

toddjordan:
I noticed the PR where you can do the same thing in a test with `this.registry.register(...)` as you can with `this.container.register(...)` If that's the preferred way I can update that.

michaelrkn:
@toddjordan is this ready to merge? I don't see `this.container.lookup`, `this.registry.register`, or `this.container.register` in there, so I want to make sure it's in the right place, since that business is a bit above my head :)

toddjordan:
@michaelrkn yeah that discussion was around another PR for the new section on stubbing services for component integration tests.  I haven't PR'd that to guides yet.  This PR is just converting this existing examples, so yes, its good to go when you guys are ok with it.

PR #779, opened 2015-09-19
add substate details

JKGisMe:
Per the discussion in PR #742

I changed more than the 1 line to keep it consistent and to add the analogous bits to the `error` section.

I also changed references to things like  `route:articles/overview` to "the `articles.overview` route" because routes are more commonly referred to in this manner and it's confusing to refer to routes that way but not templates, for example. 

¯\\\_(ツ)_/¯

### DIFF
--- a/source/routing/loading-and-error-substates.md
+++ b/source/routing/loading-and-error-substates.md
@@ -57,13 +57,18 @@ Router.map(function() {
 });
 ```
 
-Ember will try to find a `loading` route in the hierarchy
-above `foo.bar.slow-model` that it can transition into, starting with
-`foo.bar.slow-models`'s sibling:
+Ember will alternate trying to find a `routeName-loading` or `loading` template
+in the hierarchy starting with `foo.bar.slow-model-loading`:
 
-1. `foo.bar.loading`
-2. `foo.loading`
-3. `loading`
+1. `slow-model-loading`
+2. `foo.bar.loading` or `foo.bar-loading`
+3. `foo.loading` or `foo-loading`
+4. `loading` or `application-loading`
+
+It's important to note that for `slow-model` itself Ember will not try to find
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thanks for clarifying this!

KoKuToru:
This doesn't look to be correct (in Ember 2.0.2 with pods),

even if a `loading` route exists it still will try to load a non existing `slow-model-loading` route and display a empty page.

JKGisMe:
Well, first, pods,  the very popular organizational technique that has been around for a couple versions now is wholly unsupported by the guides. (see pr #742)

Second, as this was original a pods focused pr I tried everything in pods first. Can you make a twiddle recreating this behavior?

KoKuToru:
Sadly http://ember-twiddle.com/ wont save.. but i copied the files here (https://gist.github.com/KoKuToru/9f17dbf9d1e8a8fca068)

It might also be a miss understanding of the loading route on my side..
When you have a "test/subtest" router and a "loading" route with the following:

visiting `application-> test/subtest` loading is displayed but
visiting `application-> test -> test/subtest` loading is not displayed

subtest has a slow model

JKGisMe:
There is a subtlety here. Using the guides, `foo`, `foo.bar`, `foo.bar.slow-model`. If you load `foo.bar` it's loaded. So it doesn't need a loading substate. So then you click and open `slow-model` from there and it _does_ need a loading substate. So it looks at `foo.bar.slow-model-loading`, then `foo.bar.loading` but when it tries to look at `foo.loading` it stops because it thinks "hey `foo.bar` is loaded. I don't need any loading substates down here!" 

I guess technically instead of just plain looking for loading routes up the whole hierarchy it really means 

"up the heirarchy to the first already loaded route encountered or the base loading route if no routes have been loaded"

http://ember-twiddle.com/d84a8aa0c84a1fbba80f?numColumns=1

Issue #776, opened 2015-09-19
run-loop guide has mistakes


stefanpenner:
`once` and `scheduleOnce` require stable functions, anonymous ones are new on each invocation and as such are run multiple times

https://github.com/emberjs/website/commit/ddbc9bc047cba3f4648d708b3b8d45e9b08480f2

My recommendation is:

* to remove the current incorrect usage of `once`, the [API docs](http://emberjs.com/api/classes/Ember.run.html#method_scheduleOnce) explain it correctly.

michaelrkn:
Sounds good to me. Can you open a PR?

PR #781, opened 2015-09-20
Ember.run.scheduleOnce does not work with anonymous function arg.

eoinkelly:
`Ember.run.scheduleOnce` cannot ensure that the provided function is called only once if an anonymous function is passed in. This commit adjusts the examples to show passing a named function.

### DIFF
--- a/source/understanding-ember/run-loop.md
+++ b/source/understanding-ember/run-loop.md
@@ -190,15 +190,18 @@ As mentioned above, you should wrap any non-Ember async callbacks in
 Consider the following callback:
 
 ```javascript
+function sayHi() {
+  console.log('hi');
+}
+
 $('a').click(function(){
   console.log('Doing things...');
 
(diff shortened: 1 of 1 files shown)


eoinkelly:
This change should also be applied to the v1.10.0 and v1.11.0 branches. I presume the way to go is to PR it against those branches after it gets reviewed here?

mixonic:
kicked the build

eoinkelly:
Build seems broken in v1.12.0 too. Don't have time right now but will look into it at end of day.

eoinkelly:
Also this addresses issue #776 which I just noticed. I'm happy to amend the example to remove the usage of `scheduleOnce` if docs team would prefer that (it is tangential to the point being explained).

locks:
@eoinkelly please PR against master, old versions are at the moment immutable and won't be updated.

eoinkelly:
@locks will do.

PR #783, opened 2015-09-21
Fix link to "services"

cyjia:
On page `routing/redirection`, "services" should link to `services-and-initializers/services`.


locks:
Appreciated!

PR #784, opened 2015-09-22
Update defining-a-component.md

timrosenblatt:
I was looking at the guide and realized this could be written in a more straightforward way. I added a bit of clarity with respect to path names vs file names

### DIFF
--- a/source/components/defining-a-component.md
+++ b/source/components/defining-a-component.md
@@ -1,5 +1,5 @@
-To define a component, create a template whose name starts with
-`components/`. To define a new component, `{{blog-post}}` for example,
+To define a component, create a template in the `components/` directory.
+For example, to define a new component called `{{blog-post}}`,
 create a `components/blog-post` template.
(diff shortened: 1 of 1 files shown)


source/components/defining-a-component.md
```diff
@@ -1,5 +1,5 @@
-To define a component, create a template whose name starts with
-`components/`. To define a new component, `{{blog-post}}` for example,
+To define a component, create a template in the `components/` directory.
```
michaelrkn: I think it would be even more clear to say "in the `app/templates/components/` directory".

michaelrkn:
Thanks for the suggestion! You're right, this isn't very clear currently.

locks:
The whole start of this guide is confusing, I'll try reviewing it shortly

michaelrkn:
@locks https://github.com/emberjs/guides/pull/784#discussion_r40170578

PR #792, opened 2015-09-24
Add closing backtick that was missing

misfo:


michaelrkn:
Thanks for catching this!

PR #794, opened 2015-09-24
Clarify where component templates go

michaelrkn:


benoittgt:
Thanks. Read this part yesterday and thought of that. Seems better that way.

PR #796, opened 2015-09-24
Ember.run.scheduleOnce does not work with anonymous function arg.

eoinkelly:
`Ember.run.scheduleOnce` cannot ensure that the provided function is
called only once if an anonymous function is passed in.

Replaces PR #781


eoinkelly:
Happy to adjust as docs team sees fit e.g. use `() =>` syntax

eoinkelly:
Just re-read @stefanpenner suggestion in issue #776 and have adjusted the PR to remove the `scheduleOnce` examples because they are quite tangential to the run loop material.

PR #797, opened 2015-09-24
Update links.md

BigGillyStyle:
I believe in the example given there is only a 'photos.edit' route and not a 'photos.photo.edit' route.  The previous mentions of the route on this page just use 'photos.edit'.


locks:
Thanks for spotting it!

PR #807, opened 2015-09-25
Updated rollback() reference

benjaminokeefe:
I updated the rollback() reference to read rollbackAttributes(), as the function name seems to have changed in 2.0.


bmac:
Good catch. Thanks @benjaminokeefe.

PR #818, opened 2015-09-27
Add 'Working with Relationships' guide to Models

### LINKED ISSUE #785: Add a "Working With Relationships" guide to Models
Right now, there's scant documentation of how to work with related Ember Data models in the Guides. We should add a page called Working With Relationships that documents:

* how to create a new instance of a model that `belongsTo` another model
* how to update a model so that it `belongsTo` another model
* how to update a model that `hasMany` of another model
* how to work with relationships, especially that they return promises
* how to remove a relationship

kevinkucharczyk:
Closes #785 

I've added a new guide which aims to clarify how to create/update/remove relationships.

### DIFF
--- a/source/models/working-with-relationships.md
+++ b/source/models/working-with-relationships.md
@@ -1,0 +1,116 @@
+## Creating Records
+
+Let's assume that we have a `post` and a `comment` model, which are related to each other as follows:
+
+```app/models/post.js
+export default DS.Model.extend({
+  comments: DS.hasMany('comment')
+});
+```
+
+```app/models/comment.js
+export default DS.Model.extend({
+  post: DS.belongsTo('post')
+});
+```
+
+When a user comments on a post, we need to create a relationship between the two records. We can simply set the `belongsTo` relationship in our new comment:
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.createRecord('comment', {
+  post: post
+});
+comment.save();
+```
+
+This will create a new `comment` record and save it to the server. Ember Data will also update the post to include our newly created comment in its `comments` relationship.
+
+We could have also linked the two records together by updating the post's `hasMany` relationship:
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.createRecord('comment', {
+});
+post.get('comments').pushObject(comment);
+comment.save();
+```
+
+In this case the new comment's `belongsTo` relationship will be set to the parent post.
+
+## Updating Existing Records
+
+Sometimes we want to set relationships on already existing records. We can simply set a `belongsTo` relationship:
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.peekRecord('comment', 1);
+comment.set('post', post);
+comment.save();
+```
+
+Alternatively, we could update the `hasMany` relationship by pushing a record into the relationship:
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.peekRecord('comment', 1);
+post.get('comments').pushObject(comment);
+post.save();
+```
+
+## Removing Relationships
+
+To remove a `belongsTo` relationship, we can just set it to `null`, which will also remove it from the `hasMany` side:
+
+```javascript
+let comment = this.store.peekRecord('comment', 1);
+comment.set('post', null);
+comment.save();
+```
+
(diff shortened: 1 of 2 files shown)


michaelrkn:
Thanks for the quick turnaround on this!

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
...
+  post: DS.belongsTo('post')
+});
+```
+
+When a user comments on a post, we need to create a relationship between the two records. We can simply set the `belongsTo` relationship in our new comment:
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.createRecord('comment', {
+  // comment data
```
michaelrkn: I think this is self-explanatory and doesn't require a comment.

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
...
+  post: post
+});
+comment.save();
+```
+
+This will create a new `comment` record and save it to the server. Ember Data will also update the post to include our newly created comment in its `comments` relationship. We could have also linked the two records together by updating the post's `hasMany` relationship:
```
michaelrkn: I think make "We could have also..." into a new paragraph.

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
...
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.createRecord('comment', {
+  // comment data
```
michaelrkn: I think this is self-explanatory and doesn't require a comment.

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
...
+Sometimes we want to set relationships on already existing records. We can simply set a `belongsTo` relationship:
+
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.peekRecord('comment', 1);
+comment.set('post', post);
+comment.save();
+```
+
+We can also update the `hasMany` relationship by pushing a record into the relationship:
```
michaelrkn: I suggest "Alternatively, we could update the..."

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
...
+```javascript
+let post = this.store.peekRecord('post', 1);
+let comment = this.store.peekRecord('comment', 1);
+post.get('comments').pushObject(comment);
+post.save();
+```
+
+### Removing Relationships
+
+To remove a `belongsTo` relationship, we can just set it to `null`, which will also remove it from the inverse:
```
michaelrkn: I'd suggest "inverse" -> "`hasMany` side"

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
...
+let comment = this.store.peekRecord('comment', 1);
+post.get('comments').removeObject(comment);
+post.save();
+```
+
+As in the earlier examples, the comment's `belongsTo` relationship will also be cleared by Ember Data.
+
+### Relationships as Promises
+
+While working with relationships it is important to remember that they return promises. In the examples above we have assumed that the records we were working on were already loaded into the store, but this may not always be the case.
```
michaelrkn: In the above examples, we aren't actually working with the returned data, so that they returned promises isn't really relevant. I think removing the last sentence would avoid confusion about when the user needs to be concerned about the asynchronous nature of the relationship.

michaelrkn:
This looks really good! I've added a couple comments.

source/models/working-with-relationships.md
```diff
@@ -0,0 +1,116 @@
+### Creating Records
```
michaelrkn: These should be ##, not ### (and all below ones, too).

michaelrkn:
Looks good! One last comment from me. Then can you squash your commits?

Also, did you sort out your [questions](https://github.com/emberjs/guides/issues/785#issuecomment-143548037) on your own, or should we wait for a final review by an ED expert?

kevinkucharczyk:
Okay, I've added updates according to your comments and squashed my commits.

I figured out the questions on my own, since I didn't receive any replies here nor on Slack (which is understandable, since it was the weekend). The methods I propose do work, though the question still remains whether what I did is, in fact, best practice.

I believe it would be good to have someone from ED review my approach.

wecc:
LGTM

Issue #613, opened 2015-08-18
Lint code snippets


locks:
Lint code snippets and fail the build if they don't pass.

michaelrkn:
@rwjblue is there something in the Ember.js codebase that could be re-used here?

mtthgn:
I'm interested in this issue. Here's my initial thoughts after doing a little digging

1. It doesn't look like there's a good (and by "good", I mean maintained) jshint wrapper for ruby. I don't like the idea of having a dependency outside of the `Gemfile`, tho.

2. Every jshint I looked at operated on files and directories. The ideal situation for this issue would be to operate against a string. We might have to extract every snippet, make a temp directory, and then run the linter against those temporary snippet files.

3. Is there a canonical "Ember's .jshintrc" somewhere? [Ember CLI has one](https://github.com/ember-cli/ember-cli/blob/master/.jshintrc), would that be the best one to use?

michaelrkn:
@mtthgn Thanks for your interest in helping out with this! I don't know that I can add anything to points 1 and 2, but for 3, I'd suggest checking out https://github.com/emberjs/ember.js. My understanding is that PRs against ember.js are linted against their [style guide](https://github.com/emberjs/ember.js/blob/master/STYLEGUIDE.md), but I don't know where the style guide is tested against and if that code could be re-used - @rwjblue may know.

michaelrkn:
Actually, regarding 1, what about https://github.com/damian/jshint?

mtthgn:
@michaelrkn dunno how I didn't see that one! I'll see what I can get done on this over the weekend.

michaelrkn:
:) Also, maybe instead of extracting the code snippets from our Markdown files, saving them into files, and then having the linter check them, you could fork the linter and add a way to create new input sources. Something like creating an API that can be passed a string of JavaScript to be linted, and then separating out the actually getting of the JS into another class, so that anybody can provide an arbitrary source of JS. Then turn the existing file scanning logic into one source, and create your own input source that looks through our Markdown files for code fences that end in `js`. Maybe that's more trouble than it's worth, though. Just another thought.

mtthgn:
@michaelrkn I've got a PR opened [here](https://github.com/damian/jshint/pull/34) to add the ability to lint JS strings as opposed to only files. In the meantime, I'll reference my fork of the gem and get started on linting our snippets.

[...]

PR #825, opened 2015-09-28
[WIP] Setup for linting JS in our markdown files

mtthgn:
This PR sets up our test suite to make sure the JS in our markdown files is up to Ember's standard. We accomplish this by running a linter with effectively the same `.jshintrc` as the ember.js repo.

Tests are failing on purpose right now so that we can have a conversation about the following:

- What should the output look like when it fails?
- Should there be a test for each markdown file? Or should we instead have one test that makes sure _all_ js code passes?

Things to do:

- [x] Lint against a JS string, rather than an entire file
- [ ] Extract snippets from Markdown files
- [ ] Run linter against said snippets
- [ ] update gemfile to point to the original jshint gem provided
      [This PR](https://github.com/damian/jshint/pull/34) gets merged


michaelrkn:
Thanks for getting this going, @mtthgn!

I think the most straightforward way to achieve this goal is to raise an exception if code on a page doesn't pass linting. This approach is used to [prevent dead links in the table of contents](https://github.com/emberjs/guides/blob/master/lib/toc.rb#L32). That way, the page author will get immediate feedback in their development environment if there's a problem, and the CI build will fail if somehow bad code does get pushed.

mtthgn:
@michaelrkn Where would you suggest doing that for the markdown files? I'm not very familiar with middleman, but I can't find _where_ I'd raise that exception. With the table of contents, we have our own ruby file generating its DOM. 

I don't see the equivalent code where I could lint before middleman renders the page.

michaelrkn:
I think you'd need to make your own Middleman plugin to do this. You can reference toc.rb and highlighter.rb, as well as the [documentation](https://middlemanapp.com/advanced/custom_extensions/).

mtthgn:
Just wanted to let everyone know I haven't forgotten about this! I plan on getting some work done on this during the weekend :smiley:

michaelrkn:
Thanks @mtthgn!

michaelrkn:
Hey @mtthgn - any progress on this?

michaelrkn:
I'm adding "help wanted" back to the related issue, but @mtthgn feel free to ping me if you want to take this up again.

acorncom:
This works on #613

acorncom:
I'm going to close this for now, but we're happy to re-open it if folks want to work on it further!

PR #826, opened 2015-09-28
Add missing `default`

cafedomancer:


locks:
What missing default? :trollface:

PR #830, opened 2015-09-28
[WIP] Removes unnecessary CSS

locks:


acorncom:
I noticed this line could probably be cleaned up too:
https://github.com/emberjs/guides/blob/css-cleanup/source/stylesheets/guides.css.scss#L5

locks:
Let's do this post-redesign.

PR #832, opened 2015-09-29
Clarify purpose of Adapters

michaelrkn:


source/models/customizing-adapters.md
```diff
@@ -1,5 +1,6 @@
-In Ember Data, the logic for communicating with a backend data store
-lives in the `Adapter`. Ember Data's Adapter has some built-in
+In Ember Data, the Adapter determines the structure of requests to a
```
rwjblue: This definitely seems better, but I think `"structure of requests"` is a tad misleading.  The adapter does not control the structure of the request payload (the actual data being saved).

PR #833, opened 2015-09-29
Clean up and serializers and add examples

michaelrkn:
Clean up formatting and add examples of how to customize serializers.

### DIFF
--- a/source/models/customizing-serializers.md
+++ b/source/models/customizing-serializers.md
@@ -31,7 +31,7 @@ should be dash-cased. For example, if you request a record from
 {
   "data": {
     "type": "people",
-    "id": 123,
+    "id": "123",
     "attributes": {
       "first-name": "Jeff",
       "last-name": "Atwood"
@@ -47,14 +47,14 @@ A response that contains multiple records may have an array in its
 {
   "data": [{
     "type": "people",
-    "id": 123,
+    "id": "123",
     "attributes": {
       "first-name": "Jeff",
       "last-name": "Atwood"
     }
   }, {
     "type": "people",
-    "id": 124,
+    "id": "124",
     "attributes": {
       "first-name": "Yehuda",
       "last-name": "Katz"
@@ -144,7 +144,7 @@ Data:
 {
   "data": {
     "attributes": {
-      "id": 1,
+      "id": "1",
       "name": "My Product",
       "amount": 100,
       "currency": "SEK"
@@ -160,7 +160,7 @@ But our server expects data in this format:
 {
   "data": {
     "attributes": {
-      "id": 1,
+      "id": "1",
       "name": "My Product",
       "cost": {
         "amount": 100,
@@ -202,7 +202,7 @@ the server provides data that looks like:
 {
   "data": {
     "attributes": {
-      "id": 1,
+      "id": "1",
       "name": "My Product",
       "cost": {
         "amount": 100,
@@ -220,7 +220,7 @@ And so we need to change it to look like:
 {
   "data": {
     "attributes": {
-      "id": 1,
+      "id": "1",
       "name": "My Product",
       "amount": 100,
       "currency": "SEK"
@@ -287,7 +287,7 @@ in the document payload returned by your server:
 ```js
 {
   "data": {
-    "id": 44,
+    "id": "44",
     "type": "people",
     "attributes": {
       "first-name": "Barack",
@@ -482,10 +482,10 @@ to be a JSON object that looks similar to this:
 
 ```json
 {
-    "id": 1,
+    "id": "1",
     "title": "Rails is omakase",
     "tag": "rails",
-    "comments": [1, 2]
+    "comments": ["1", "2"]
 }
 ```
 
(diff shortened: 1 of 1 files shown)


source/models/customizing-serializers.md
```diff
@@ -135,7 +136,40 @@ import DS from 'ember-data';
...
-#### ID
+To change the format of the data that is sent to the backend store, you can use
+the `serialize` hook:
+
+```app/serializers/application.js
+import DS from 'ember-data';
+
+export default DS.JSONSerializer.extend({
+  serialize(snapshot, options) {
+    // help! what do i do here?
```
michaelrkn: Help! What do I do here? Can someone provide a code example similar to the one below for normalizing?

michaelrkn:
@wecc and/or @bmac, could you review this to make sure it's accurate, and help me with my partial example?

source/models/customizing-serializers.md
```diff
@@ -1,23 +1,24 @@
-## Serializers
+In Ember Data, serializers format the data sent to and recieved from
```
locks: `received`

source/models/customizing-serializers.md
```diff
@@ -1,23 +1,24 @@
...
+the backend store. By default, Ember Data serializes data using the
+[JSON API](http://jsonapi.org/) format. If your backend uses a different
+format, Ember Data allows you to customize the serializer or use a
+different serializer entirely.
 
-In Ember Data the Serializer is a class that is responsible for
-transforming the data returned by your adapter into a JSON API document
-that Ember Data can understand. Serializers can be customized to work
-with any backend. By default, Ember Data ships with 3 Serializers. The
+Ember Data ships with 3 Serializers. The
```
locks: which?

source/models/customizing-serializers.md
locks: @bmac @wecc

wecc: Something like this?

```js
  serialize: function(snapshot, options) {
    var json = this._super(...arguments);

    /*
     * In this example our server expects the attributes `amount` and `currency`
     * to be embedded in `cost`.
     * 
     * Before:
     *   { id: 1, name: 'My Product', amount: 100, currency: 'SEK' }
     * 
     * After:
     *   { id: 1, name: 'My Product', cost: { amount: 100, currency: 'SEK' } }
     */

    json.cost = {
      amount: json.amount,
      currency: json.currency
    };

    delete json.amount;
    delete json.currency;

    return json;
  },
```

wecc:
There are a couple of instances of `"id": 123` (outside of your changes), they should be `"id": "123"` as IDs in JSON API are strings.

source/models/customizing-serializers.md
```diff
@@ -135,7 +136,40 @@ import DS from 'ember-data';
...
+```
+
+If your backend store provides data in a format other than JSON API, you can
+use the `normalizeResponse` hook to change the data into the JSON API format:
+
+```app/serializers/application.js
+import DS from 'ember-data';
+
+export default DS.JSONSerializer.extend({
+  normalizeResponse(store, primaryModelClass, payload, id, requestType) {
```
wecc: We should really encourage people to use `this._super`.

Basically:
Before you call super, the payload is in the format that the serializer expects. Munge it but keep that format.
After you call super, the payload is now in JSON API format, feel free to munge as long as you keep it JSON API.

Delegating conversion to super gives you A LOT for free + you have the flexibility to make customizations in two different formats.

michaelrkn: :+1:

So like this?

```
  normalizeResponse: function(store, primaryModelClass, payload, id, requestType) {
    /*
     * In this example our server provides the attributes `amount` and `currency`
     * embedded in `cost`.
     * 
     * Before:
     *   { id: 1, name: 'My Product', cost: { amount: 100, currency: 'SEK' } }
     * 
     * After:
     *   { id: 1, name: 'My Product', amount: 100, currency: 'SEK' }
     */

    payload.amount = payload.cost.amount;
    payload.amount = payload.cost.currency;

    delete payload.cost;

    return this._super(store, primaryModelClass, payload, id, requestType);
  },
```

wecc: LGTM!

wecc: Oh wait, no. `normalizeResponse` is for the full payload, so it kind of depends.

wecc: If you want to manipulate the data for a single resource then the `normalize` hook is better.

wecc: You can use `normalizeResponse` to remove top-level data for server-responses. Like if your server kind of complies to JSONSerializer but puts everything in a `{ "response": ... }`, you can get do `payload = payload.response`

michaelrkn: So if you use an application serializer, use `normalizeResponse`, and if you use a model serializer, then use `normalize`?

wecc: I'd say that's a pretty common pattern, yes.

michaelrkn:
@wecc Can you do one last review? It took me a couple days to find the time to verify all the code examples worked properly.

PR #834, opened 2015-09-29
Updated Component Actions Guide

### LINKED ISSUE #507: Update Guides with closure actions
@mixonic is on this!

tomdale:
This PR isn't ready for merge yet, but I wanted to open it up for review before I headed to bed. I'll try to get this finished off tomorrow.

### Open Questions

With the introduction of "closure actions" (a term I've avoided here, since it's too insider baseball), we have three different APIs:

* `onClick=(action 'doSomething')`
* `<button {{action 'doSomething'}}>`
* `this.sendAction('doSomething')`

Explaining all of these in a coherent way, given that they have the same name for similar-but-different concepts, is tricky. I'd like to figure out which of these is "legacy" and we can begin to phase out in the guides, and which ones we still need to keep around until Routable Components land.

### What's Left

- [ ] Bubbling to routes (I temporarily removed this, but I suspect I'll have to add it back)
- [ ] Threading actions through multiple layers of components
- [ ] Passing arguments
- [ ] Argument currying
- [ ] Return values
- [ ] `target`?
- [ ] `value`?

Not sure about the last two—I don't know if I have a strong enough mental model of them myself to know when they'd be useful.

This fixes #507.


trek:
<3

mixonic:
> I'd like to figure out which of these is "legacy" and we can begin to phase out in the guides, and which ones we still need to keep around until Routable Components land.

Agreed. There are two concepts that are legacy, but we must keep around until later work completes.

* Actions on controllers behave differently. If they are targeted with the element-space helper (`<button {{action 'save'}}>`) then they bubble through to routing. With the introduction of routable components, this case will no longer remain.
* The `<button {{action 'save'}}>` syntax in general is unlike any other helper in Ember 2.0. It is the only element-space helper we have today. We may introduce `{{attrs` or something, but regardless I find the inconsistency of attaching an action to an element vs a component unfortunate. We've discussed for some time but not made moves toward "kabob actions". These would add support for attaching actions as element attributes, for example `<button on-click={{action 'save'}}>`. Importantly, this is without using simply a native event handler (ala `onclick={{action`) but instead opts into Ember's event management.

Either routable components or kabob actions can resolve most of the inconsistencies alone.

I'm happy to renew a push to get kabobs into an RFC, just need to get the component composition stuff I'm working on under control.

source/components/triggering-changes-with-actions.md
```diff
@@ -0,0 +1,277 @@
+So far, you've learned how parent components can pass attributes in to a
+child component, and how that component can use those attributes from
+both JavaScript and its Handlebars template.
```
mixonic: I suggest dropping "Handlebars". Ember templates are simply Ember templates. They just happen to have Handlebars syntax.

source/components/triggering-changes-with-actions.md
```diff
@@ -0,0 +1,277 @@
...
+Let's take it step by step.
+
+#### Implementing the Action
+
+First, we'll define what we want to happen when the user double clicks
+the button. In this case, we'll find the user's account and delete it.
+
+Where do you put the code for an action? In Ember, each component can
+have a property called `actions`, where you put functions that can be
+triggered by child components.
```
mixonic: "where you can put functions that can be triggered" ~ "where you can define action handlers"

source/components/triggering-changes-with-actions.md
```diff
@@ -0,0 +1,277 @@
...
+
+Let's look at the parent component's JavaScript file. In this example,
+imagine we have a parent component called `user-profile` that shows the
+user's profile to them.
+
+We'll implement an action on the parent component called
+`userDidDeleteAccount()` that, when called, gets a hypothetical `login`
+service and calls the service's `deleteUser()` method.
```
mixonic: In a straight read-through of the guides, services have not yet been discussed.

source/components/triggering-changes-with-actions.md
```diff
@@ -0,0 +1,277 @@
...
+```
+
+That means that calling the component's `onDoubleClick` method will call
+the parent's `userDidDeleteAccount` action.
+
+```app/components/double-click-button.js
+doubleClick() {
+  // Call the component's `onDoubleClick()` method to trigger
+  // the parent's `userDidDeleteAccount` action.
+  this.onDoubleClick();
```
mixonic: `this.sendAction('onDoubleClick')` is consistent with the strings-passing world and not deprecated. If you wanted to directly address the action function, I think you would need to suggest `this.get('onDoubleClick')();`

source/components/triggering-changes-with-actions.md
tomdale: I agree that my sentence isn't ideal and can be improved but I want to push back on your suggestion.

Previously, we've used the term "sending actions." This was great when actions were like a message in a bottle that you'd toss into the sea and hope some route or controller found it on a distant coast.

However, given the "closure action" semantics, describing components as "sending actions" (as in, a child component "sends an action" to a parent) just didn't fit the mental model. That's why, in this guide, I've switched to saying that a parent component "provides an action" or "gives a child component an action," and the child component "triggers the action."

IMO, this clears up a lot of the ambiguity we had previously, where the term "action" would refer to the function that got run (e.g. "action handlers" in your comment) and `sendAction`, which would actually trigger it.

tomdale: Relevant: https://twitter.com/csallen/status/648682071666241536

source/components/triggering-changes-with-actions.md
tomdale: Would you be satisfied with a link to the Services section? I'm not sure how to avoid the circular dependency problem entirely. In this case, the example is sufficiently simple that I suspect must people will be able to suspend disbelief and guess at what's happening.

source/components/triggering-changes-with-actions.md
tomdale: Seems legit

source/components/triggering-changes-with-actions.md
tomdale: > I think you would need to suggest `this.get('onDoubleClick')();`

Why do you say that?

source/components/triggering-changes-with-actions.md
```diff
@@ -0,0 +1,277 @@
...
+
+```js
+function onClick(element, onClickFunc) {
+  element.addEventListener('click', onClickFunc);
+}
+
+let element = document.getElementById('delete-button');
+let displayAlert = () => {
+  alert("button clicked!");
+};
```
acorncom: Should we use non-ES6 syntax here when demonstrating older callback syntax?  Callbacks are used heavily by folks in the jQuery world, but ES6 isn't.  Seems like we're trying to help explain actions based on something that is familiar, but using syntax that may be unfamiliar may make that harder to grok ...

tomdale:
Oh, one other thing about actions: routes and controllers can call `send()` (not `sendAction()`, confusingly!) to participate in the action bubbling chain.

source/components/triggering-changes-with-actions.md
rwjblue: Calling `this.onDoubleClick()` only happens to work properly due to the way we give initial attrs to components. You should generally be using `this.get('foo')` when your component is provided an attr via `{{thing-name foo="blah"}}`.

source/components/triggering-changes-with-actions.md
rwjblue: I actually think this is much simpler to understand:

```
function displayAlert() {

}
```

Preserving `this` context in the root of a module is not useful (since it is `undefined` anyways)...

source/components/triggering-changes-with-actions.md
mixonic: specifically I'm thinking about the word "triggers" as being funky. IMO it implies some evented system that does not exist, instead they are just "called". In Just JavaScript, I would never describe `foo()` as triggering a function. I'm agreed with your thoughts on direction though.

source/components/triggering-changes-with-actions.md
mixonic: I'm ok with cross linking sections a bit. Other parties have been less encouraging.

mixonic:
@tomdale correct you are :-/

source/components/triggering-changes-with-actions.md
praxxis: I've taken to saying "invoking" actions, which also implies that they're just functions being passed around

michaelrkn:
As far as I know, we are shipping 2.1 tomorrow, so I have gone ahead and updated this based on the comments here and discussions I've had with @tomdale. I've also opened #853 to address the remaining items on Tom's What's Left list. If anybody has time today or tomorrow, it would be great to get any high priority items from #853 taken care of before 2.1 ships. If not, then onwards to 2.2!

toddjordan:
@michaelrkn is there any word on which of the items in https://github.com/emberjs/guides/issues/853 would be considered high priority?

michaelrkn:
@toddjordan I'm sorry but I really don't know. If nothing jumps out at you, I'd suggest asking in the slack channel if anybody else has an opinion.

Issue #840, opened 2015-09-30
Add a section in Testing Components for stubbing services


toddjordan:
With the onset of component integration tests, I'm seeing a lot of questions between the slack testing channel and the discuss forum on how one should stub services.  This is a common enough requirement where I think a guide section is warranted.

toddjordan:
I have a PR to go along with.  I'd like some feedback, especially around the way I'm registering and looking up the stub service.

PR #841, opened 2015-09-30
Add section on stubbing services

toddjordan:
PR for issue https://github.com/emberjs/guides/issues/840

I've validated the below approach is functional, just need feedback on whether these apis are sanctioned.

### DIFF
--- a/source/testing/testing-components.md
+++ b/source/testing/testing-components.md
@@ -180,3 +180,102 @@ test('should trigger external action on form submit', function(assert) {
   this.$('input').click();
 });
 ```
+### Stubbing Services
+
+In cases where components have dependencies on Ember services, it is possible to stub these
+dependencies for integration tests. Stub Ember services by using the built-in register
+function to register your stub service in place of the default.
+
+Imagine you have the following component that uses a location service to display the city
+and country of your current location:
+
+> You can follow along by generating your own component with `ember generate
+> component location-indicator`.
+
+```app/components/location-indicator.js
+export default Ember.Component.extend({
+  locationService: Ember.inject.service('location-service'),
+
+  //when the coordinates change, call the location service to evaluate what the city and country would be
+  city: Ember.computed('locationService.currentLocation', function () {
+    return this.get('locationService').getCurrentCity();
+  }),
+
+  country: Ember.computed('locationService.currentLocation', function () {
+    return this.get('locationService').getCurrentCountry();
+  })
+});
+```
+
+```app/templates/components/location-indicator.hbs
+You currently are located in {{city}}, {{country}}
+```
+To stub the location service in your test, create a local stub object that extends
+`Ember.Service`, and register the stub as the service your tests need in the
+beforeEach function.  In this case we initially force location to New York.
+
+
+```tests/integration/components/location-indicator-test.js
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import Ember from 'ember';
+
+//keep a reference to the location service stub instance
+//for manipulating and asserting within tests
+let locationService;
+
+//Stub location service
+const locationStub = Ember.Service.extend({
+  city: 'New York',
+  country: 'USA',
+  currentLocation: {
+    x: 1234,
+    y: 5678
+  },
+
+  init() {
+    this._super(...arguments);
+    locationService = this;
+  },
+  getCurrentCity() {
+    return this.get('city');
+  },
+  getCurrentCountry() {
+    return this.get('country');
+  }
+});
+
+moduleForComponent('location-indicator', 'Integration | Component | location indicator', {
+  integration: true,
+
+  beforeEach: function () {
+    this.register('service:location-service', locationStub);
+  }
+});
+```
+
+Once the stub service is registered the test simply needs to check that the stub data that
+is being returned from the service is reflected in the component output.
+
+```tests/integration/components/location-indicator-test.js
+test('should reveal current location', function(assert) {
+  this.render(hbs`{{location-indicator}}`);
+  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA');
+});
+```
+
+In the next example, we'll add another test that validates that the display changes
+when we modify the values on the service.
+
+```tests/integration/components/location-indicator-test.js
+test('should change displayed location when current location changes', function (assert) {
+  this.render(hbs`{{location-indicator}}`);
+  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA', 'origin location should display');
+  Ember.run(() => {
+    locationService.set('city', 'Beijing');
+    locationService.set('country', 'China');
+    locationService.set('currentLocation', { x: 11111, y: 222222 });
+  });
+  assert.equal(this.$().text().trim(), 'You currently are located in Beijing, China', 'location display should change');
+});
+```


locks:
:tada:

benoittgt:
Thanks @toddjordan for this PR. I read it and it's ok for me. :+1:

source/testing/testing-components.md
```diff
@@ -180,3 +180,96 @@ test('should trigger external action on form submit', function(assert) {
   this.$('input').click();
 });
 ```
+### Stubbing Services
+
+In cases where components have dependencies on Ember services, its possible to stub these
```
kevinkucharczyk: I think "its" should be "it's" (missing apostrophe)

source/testing/testing-components.md
```diff
@@ -180,3 +180,96 @@ test('should trigger external action on form submit', function(assert) {
...
+```tests/integration/components/location-indicator-test.js
+test('should reveal current location', function(assert) {
+  this.render(hbs`{{location-indicator}}`);
+  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA');
+});
+```
+
+Since services default to being singletons, we can get a reference to the service within our test
+and modify values on it so that we can assert on how we expect our component to react. In next the
+example, we'll add another test that validates that the display changes when we modify the values
```
kevinkucharczyk: "In next the example" => "In the next example"

source/testing/testing-components.md
toddjordan: Yeah, I'll just change it to "it is"

source/testing/testing-components.md
toddjordan: ha, good catch.  Was writing like yoda there

source/testing/testing-components.md
```diff
@@ -180,3 +180,96 @@ test('should trigger external action on form submit', function(assert) {
...
+
+```tests/integration/components/location-indicator-test.js
+test('should reveal current location', function(assert) {
+  this.render(hbs`{{location-indicator}}`);
+  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA');
+});
+```
+
+Since services default to being singletons, we can get a reference to the service within our test
+and modify values on it so that we can assert on how we expect our component to react. In the next
```
michaelrkn: This first sentence is kind of hairy :) I think you could just remove it entirely without losing any meaning, and it would make it clearer to most people.

source/testing/testing-components.md
```diff
@@ -180,3 +180,96 @@ test('should trigger external action on form submit', function(assert) {
...
+test('should change displayed location when current location changes', function (assert) {
+  this.render(hbs`{{location-indicator}}`);
+  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA', 'origin location should display');
+  let locationService = this.container.lookup('service:location-service');
+  Ember.run(() => {
+    locationService.set('city', 'Beijing');
+    locationService.set('country', 'China');
+    locationService.set('currentLocation', { x: 11111, y: 222222 });
+  });
+  assert.equal(this.$().text().trim(), 'You currently are located in Beijing, China', 'location display should change');
```
michaelrkn: In the first test, you didn't have this last argument, which seems redundant with line 264's `'should change displayed location when current location changes'`. I think either remove it here or add it above as well.

source/testing/testing-components.md
toddjordan: ok, so just to be clear, remove the sentence starting with "Since services default...."

michaelrkn:
Thanks @toddjordan! I've added a couple notes, and also asked in the _testing_ Slack channel for someone with more expertise to review.

source/testing/testing-components.md
michaelrkn: Yup

source/testing/testing-components.md
toddjordan: So as a general practice, I add the end arguments on asserts to explain what they are asserting.  Otherwise they show up in the testem ui output as just 'ok'.  When they fail (or pass), its nice to have the extra info.  In this case there's 2 asserts, first one asserts that the location displays after render, second asserts that the location is updated onscreen after the service changes.

source/testing/testing-components.md
toddjordan: cool, will update

source/testing/testing-components.md
michaelrkn: Oh, I missed that there were two tests in here. Sounds good to me! I take back everything ;)

source/testing/testing-components.md
```diff
@@ -180,3 +180,96 @@ test('should trigger external action on form submit', function(assert) {
   this.$('input').click();
 });
 ```
+### Stubbing Services
+
+In cases where components have dependencies on Ember services, it is possible to stub these
+dependencies for integration tests. Stub Ember services by using the container
```
toddjordan: I'll update this to "by using the registry"

pixelhandler:
@toddjordan @michaelrkn I think the section on stubbing a service is well done :+1:

source/testing/testing-components.md
```diff
@@ -180,3 +180,94 @@ test('should trigger external action on form submit', function(assert) {
...
+  getCurrentCountry() {
+    return this.get('country');
+  }
+});
+
+moduleForComponent('location-indicator', 'Integration | Component | location indicator', {
+  integration: true,
+
+  beforeEach: function () {
+    this.registry.register('service:location-service', locationStub);
```
rwjblue: `this.registry` is the private internal registry.  I do not think we should recommend using it here.

source/testing/testing-components.md
```diff
@@ -180,3 +180,94 @@ test('should trigger external action on form submit', function(assert) {
...
+```
+
+In the next example, we'll add another test that validates that the display changes
+when we modify the values on the service.
+
+```tests/integration/components/location-indicator-test.js
+test('should change displayed location when current location changes', function (assert) {
+  this.render(hbs`{{location-indicator}}`);
+  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA', 'origin location should display');
+  let locationService = this.container.lookup('service:location-service');
```
rwjblue: `this.container` is private.  This usage can be avoided via a small tweak to the locationStub:

```javascript
init() {
  this._super(...arguments);
  locationService = this;
}
```

source/testing/testing-components.md
toddjordan: ok, then what would you suggest we use to register the stub service?

rwjblue:
Can you update the private usage of `this.registry` and `this.container` to use the techniques being added in https://github.com/switchfly/ember-test-helpers/pull/105?  As soon as that PR is updated (I left a tiny tweaking comment), we can land and release an updated ember-qunit version.

source/testing/testing-components.md
rwjblue: The code being added in https://github.com/switchfly/ember-test-helpers/pull/105

toddjordan:
@rwjblue updates made.  Now we'll just wait for the next ember-qunit release to make sure it actually works ;-)

rwjblue:
Sounds good, I'll update here once @ef4 updates that PR to fix that last comment...

toddjordan:
I see ember-qunit 0.4.12 is out.  Trying it out on the examples...

toddjordan:
Validated that the updated example works on ember-qunit 0.4.12.  @michaelrkn @rwjblue  I believe we are good to go

ef4:
I added a followup in #846 that takes advantage of the new `this.inject` as well.

rwjblue:
:+1:

rwjblue:
Thanks for the hard work, y'all!

PR #844, opened 2015-10-01
Clean up passing properties to components

michaelrkn:


source/components/passing-properties-to-a-component.md
```diff
@@ -1,97 +1,77 @@
-By default a component does not have access to any properties normally
-available in the template in which it is used (sometimes referred to
-as the template `scope`)
+Components only have access to properties passed to them.
```
locks: but what you pass in, aren't they attributes? :D

source/components/passing-properties-to-a-component.md
```diff
@@ -1,97 +1,77 @@
...
 It is important to note that these properties stay in sync (technically
 known as being "bound"). That is, if the value of `componentProperty`
 changes in the component, `outerProperty` will be updated to reflect that
 change. The reverse is true as well.
 
-You can also bind properties from inside an `{{#each}}` loop. This will
-create a component for each item and bind it to each model in the loop.
+## Positional Params
+
+Instead of passing parameters in by name, you can pass them in by position.
```
locks: you can also use both of them at the same time, this sentence kinda makes it feel XOR

source/components/passing-properties-to-a-component.md
```diff
@@ -1,97 +1,77 @@
...
 ```
-If you are using the `{{component}}` helper to render your component, you can
-pass properties to the chosen component in the same manner:
-
-```handlebars
-{{component componentName title=model.title name=model.name}}
-```
 
-### Positional Params
+To set the component up to recieve parameters this way, you need
```
locks: receive (you need a text expansion for this one)

source/components/passing-properties-to-a-component.md
```diff
@@ -1,97 +1,77 @@
...
-Apart from passing attributes to a component, you can also pass in positional parameters
-by setting the `positionalParams` attribute in your component class.
-
-```app/components/x-visit.js
-const MyComponent = Ember.Component.extend();
-
-MyComponent.reopenClass({
-  positionalParams: ['name', 'model']
+```app/components/blog-post.js
+export default Ember.Component.extend({
```
locks: AFAIK this is the old way, which is the wrong way.

locks:
I feel like some valuable information is lost in your edit of positional arguments.

source/components/passing-properties-to-a-component.md
michaelrkn: i just need to learn how to spell ;)

source/components/passing-properties-to-a-component.md
```diff
@@ -1,97 +1,88 @@
-By default a component does not have access to any properties normally
-available in the template in which it is used (sometimes referred to
-as the template `scope`)
+Components only have access to attributes passed to them.
```
locks: I like terse, but maybe expand a tiny bit? Remember you removed the whole index.

```
Components are isolated from their surroundings, so any data that the component needs has to be passed in.
```

?

michaelrkn:
All done?!

source/components/passing-properties-to-a-component.md
```diff
@@ -1,97 +1,89 @@
...
+BlogPostComponent.reopenClass({
+  positionalParams: ['title', 'body']
 });
 
-export default MyComponent;
+export default BlogPostComponent;
 ```
 
-This will expose the first parameter in your template as the `{{name}}` attribute and the second as `{{model}}`. They will also be available as regular attributes in your component via `this.get('name')` and `this.get('model')`.
+The you can use the attributes in the component exactly as if they had been
```
locks: you accidentally something at the beginning of the sentence

michaelrkn:
Ok how about NOW?

locks:
Not even waiting for CI. Que sera, sera.

michaelrkn:
:boom:

PR #846, opened 2015-10-01
Use injection to access service from test context

ef4:
This behavior is supported in ember-qunit 0.4.12, and I think it's nicer than asking users to save their own local variable during init.

### DIFF
--- a/source/testing/testing-components.md
+++ b/source/testing/testing-components.md
@@ -220,10 +220,6 @@ import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
 import Ember from 'ember';
 
-//keep a reference to the location service stub instance
-//for manipulating and asserting within tests
-let locationService;
-
 //Stub location service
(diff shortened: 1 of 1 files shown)


source/testing/testing-components.md
```diff
@@ -272,9 +265,9 @@ test('should change displayed location when current location changes', function
   this.render(hbs`{{location-indicator}}`);
   assert.equal(this.$().text().trim(), 'You currently are located in New York, USA', 'origin location should display');
   Ember.run(() => {
```
toddjordan: I think we can remove the ember.run using this.set now

ef4: Oh yes, that's true. The sets handle their own run loops.

toddjordan:
I ran the example and it passes so we are good from that end.

toddjordan:
@michaelrkn could we get second merge when the above comment is updated?

ef4:
I'll do a followup PR.

toddjordan:
cool

Issue #789, opened 2015-09-23
Update initializers page to match 2.1 initializer changes


acorncom:
Per https://github.com/emberjs/ember.js/issues/12336, there are changes come to the `initializers` page that will need to be put in.

@locks @michaelrkn Seems like we'd need to wait until 2.1 lands (and guides have split) to do this, but do we want to have a pull request pending to make the needed changes as soon as possible?

michaelrkn:
@acorncom see https://github.com/emberjs/ember.js/issues/12336#issuecomment-142636333. A PR would be very welcome!!

dgeb:
As requested by @michaelrkn, I'll work on this.

michaelrkn:
@dgeb, can you also take care of #771/#772 while you're at this? It seems like this will supplant/fix those.

michaelrkn:
@dgeb What's the status here? I believe 2.1 ships in the next couple days, so we need this ASAP to get it in!

rwjblue:
I am generally shooting for release on Sunday, but it might end up being Monday (depends on this and a few other factors).

michaelrkn:
Sounds good!

dgeb:
I'm seeing how much of #827 I can complete before the release, since the initializer docs will be much improved by additional context. Given @rwjblue's schedule, I'll shoot for a PR by Friday.

michaelrkn:
Sounds good. If you have the bandwidth, it would be great to get a PR by Thursday so we have Friday to review (the weekends are often pretty quiet in terms of folks helping out with docs).

dgeb:
> it would be great to get a PR by Thursday so we have Friday to review

ok, will try!

PR #848, opened 2015-10-01
Introduce "Application Concerns" section.

### LINKED ISSUE #827: Expand coverage of applications and their lifecycles
While working on #789, I've come to realize that the entire "Services and Initializers" section needs a refresh.

I propose that we expand this section to cover applications, application instances, registrations (via `RegistryProxy` methods) as well as lookups (via `ContainerProxy` methods).

This whole section should probably be titled something like "Application Lifecycle" or simply "Applications" with subsections such as:

* Applications and instances
* Registrations and lookups
* Initializers [covering both application and instance initializers]
* Services

Open to suggestions here. I will do what's needed to launch the registry / container reform this week, but I also feel that much of this is needed to put the changes in the proper context.

dgeb:
Replaces the "Services and Initializers" section and expands its
content to include pages explaining Applications and Instances, as
well as Dependency Injection.

All the examples and text have been reviewed and updated for
clarity and compatibility with Ember 2.1.

One caveat is that the Initializers page references the
`instance-initializer` generator, which is still a WIP at the
moment. I'll try to get that into CLI ASAP.

Closes #827, #789, and #771.


source/applications/dependency-injection.md
```diff
@@ -0,0 +1,183 @@
...
+factory with `application.register('user:user-to-register')`.
+
+Factory registrations must be performed either in application or application
+instance initializers (with the former being much more common).
+
+For example, an application initializer could register a `Logger` factory with
+the key `logger:main`:
+
+```app/initializers/logger.js
+export function initialize(application) {
```
mixonic: Exporting a bare function like this is OK? Doesn't this need the `Ember.Application.initializer(` call?

source/applications/dependency-injection.md
```diff
@@ -0,0 +1,183 @@
...
+instantiated with the property `logger` injected. The value of `logger` will
+come from the factory named `logger:main`.
+
+Routes in this example application can now access the injected logger:
+
+```app/routes/index.js
+export default Ember.Route.extend({
+  activate() {
+    // The logger property is injected into all routes
+    this.logger.log('Entered the index route!');
```
mixonic: Though I know it isn't required, I would prefer we continue to suggest `this.get('logger')` here as it is more refactoring-friendly (works with injected services).

source/applications/dependency-injection.md
```diff
@@ -0,0 +1,183 @@
...
+Injections can also be made on a specific factory by using its full key:
+
+```js
+application.inject('route:index', 'logger', 'logger:main');
+```
+
+In this case, the logger will only be injected on the index route.
+
+Injections can be made onto any class that requires instantiation. This includes
+all of Ember's major framework classes, such as components, routes, and the
```
mixonic: and helpers! :-)

source/applications/dependency-injection.md
dgeb: Sorry, this is only a partial initializer implementation. The part that's left off would be:

```

export default {
  name: 'logger',
  initialize: initialize
};
```

I should just add this for clarity.

source/applications/initializers.md
```diff
@@ -0,0 +1,100 @@
...
+primary means to configure [dependency injections](./dependency-injection) in
+your application.
+
+Application instance initializers are run as an application instance is loaded.
+They provide a way to configure the initial state of your application, as well
+as to set up dependency injections that are local to the application instance
+(e.g. A/B testing confurations).
+
+Initializers can perform any configuration operations, as long as they are
+synchronous. The user experience is degraded by attempting to block the booting
```
mixonic: I thoughts initializers could `defer`? For example, if you wanted to lazily load some code on application boot then register it you could use an initializer.

source/applications/dependency-injection.md
dgeb: ... and to all the other initializer examples on this page.

source/applications/dependency-injection.md
dgeb: :+1:

source/applications/services.md
```diff
@@ -13,18 +13,31 @@ Example uses of services include:
...
+Like any Ember object, a service is initialized and can have properties and
+methods of its own.
 
 ```app/services/shopping-cart.js
 export default Ember.Service.extend({
-  items: [],
+  items: null,
+
+  init() {
+    this._super();
```
mixonic: `this._super(...arguments);`

mixonic:
:clap:

source/applications/initializers.md
dgeb: > I thoughts initializers could defer? 

Application initializers (not application instance initializers) could technically call `deferReadiness`, perform something async, and then call `advanceReadiness`.

However, with the application / instance split (and the accompanied registry / container split), it is no longer possible to do anything state related because there isn't even a container that's accessible from an application initializer.

I think the use cases for advancing / deferring readiness are so slim now that it's almost always an anti-pattern that's better handled in a route hook (see the comment below). However, I'm open to hearing objections :)

> For example, if you wanted to lazily load some code on application boot then register it you could use an initializer.

As for lazy code loading, this should eventually be handled by engines, but the asynchrony will be integrated into the router. I'd suggest that in the future lazy code loading shouldn't defer loading of an application, but will be reason to create a lazy-loaded engine. Obviously, we don't want to confuse the docs with engines at this point - just clarifying my position.

dgeb:
Thanks for the review @mixonic. Just pushed a revision.

source/applications/initializers.md
mixonic: Hm. I 100% agree there are no cases for deferring loading in instanceInitializers and we should just say it isn't possible.

I'm less sure of initializers. In Torii we use defer in an initializer to keep the app from booting while we do some coordination with the parent window, and that seems legitimate to me. I'd like to just shuffle the wording to not imply that you *cannot* do anything async, but simply emphasize that you should use other patterns. ala:

> Instance Initializers can perform synchronous configuration operations. Asynchronous operations that are intended for the first render of the application should take place in routing hooks. Moving asynchronicity into routing allows for a better user experience including a rendered "loading" substate.

dgeb: Well, that sounds like a very good example of one of those "slim" use cases, which may not be as rare as I think!

Here's an attempt to straddle this issue by heavily discouraging asynchrony, but recognizing it as a possibility:

> Operations performed in initializers should be kept as lightweight as possible
to minimize delays in loading your application. Although advanced techniques
exist for allowing asynchrony in application initializers (i.e. `deferReadiness` 
and `advanceReadiness`), these techniques should generally be avoided. Any
asynchronous loading conditions (e.g. user authorization) are almost always
better handled in your application route's hooks, which allows for DOM
interaction while waiting for conditions to resolve.

If that works for you, I'll update again.

mixonic: :+1: Thanks sir :bow:

dgeb: Sure thing! And thanks for your careful review :100:

Just pushed the above revision.

michaelrkn:
This looks really awesome! I just have two thoughts about it:

1. The Guides are typically structured from most common/useful features to know about, to more obscure/abstract ideas. The current order makes sense from the perspective of how an application works, but presents the most commonly-used feature (services) last. I'd suggest re-ordering it to:

* Services
* Initializers
* Applications and Instances
* Dependency Injection
* The Container

2. I'd also suggest we move the Run Loop guide into here. Currently it's in Components, which is really not great. It doesn't totally fit in with the rest of the content here, but it's much more closely aligned.

Thanks for the great work!

dgeb:
I agree that the run loop section would be a better fit here.

I think that we need to lead with the "applications and instances" section because the distinction is a bit confusing without an explanation, and the distinction is referenced throughout the initializers and DI sections.

The container section has been absorbed by the DI section and no longer exists on its own.

michaelrkn:
Thanks for all your work on this, @dgeb!

dgeb:
@michaelrkn glad to help! 

Let's revisit your ordering / organization concerns with separate PRs. I strongly agree about the run loop section being a better fit here than under components. Also, maybe the sub-section order could be tweaked a bit.

Still waiting to resolve the `instance-initializer` generator issue over in https://github.com/ember-cli/ember-cli/pull/4218  Hope we can get that merged before the 2.1 guides go live - otherwise, we should make a note in the new instance initializer section.

michaelrkn:
Sounds good!!

PR #849, opened 2015-10-01
Fix up the passing params positional section

mixonic:
The positional params section has some incorrect syntax. This fixes it and cleans up a confusing example with `each` at the end of the section.


michaelrkn:
Thanks @mixonic!

PR #851, opened 2015-10-02
Add Handling Events page to Components, lists event names

### LINKED ISSUE #621: Add a list of supported event type name on actions
Hi there. I think would be very helpful to have the list of events names that we can use when specifying  a action, e.g. `action 'someAction' on="submit"`. 

I assume that the best place for it would be here: http://guides.emberjs.com/v2.0.0/templates/actions/#toc_specifying-the-type-of-event Also, I tried out that example with the dash in the name and didn't worked (ember v2.0.0). Digging a little bit I found [this in the code](https://github.com/emberjs/ember.js/blob/v2.0.0/packages/ember-views/lib/system/event_dispatcher.js#L42), in there contain a list of events from the event dispatcher class.

I have 3 questions here:
- Are those types of events that that Ember supports?
- Which way is the correct way to specify the event name? `mouseUp` or `mouse-up`?
- Can I open a PR adding the list that is in the code as event names? Or do you have any other suggestion?

Thank you.

pixelhandler:
[rendered page](https://github.com/pixelhandler/guides/blob/components-handling-events/source/components/handling-events.md)

[Fixes #621]

### DIFF
--- a/source/components/handling-events.md
+++ b/source/components/handling-events.md
@@ -1,0 +1,120 @@
+You can respond to user events on your component like double-clicking, hovering,
+and key presses through event handlers. Simply implement the name of the event
+you want to respond to as a method on your component.
+
+For example, imagine we have a template like this:
+
+```hbs
+{{#double-clickable}}
+  This is a double clickable area!
+{{/double-clickable}}
+```
+
+Let's implement `double-clickable` such that when it is
+clicked, an alert is displayed:
+
+```app/components/double-clickable.js
+export default Ember.Component.extend({
+  doubleClick: function() {
+    alert("DoubleClickableComponent was clicked!");
+  }
+});
+```
+
+Browser events may bubble up the DOM which potentially target parent component(s)
+in succession. To enable bubbling `return true;` from the event handler method
+in your component.
+
+```app/components/double-clickable.js
+export default Ember.Component.extend({
+  doubleClick: function() {
+    Ember.Logger.info("DoubleClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
+See the list of event names at the end of this page. Any event can be defined
+as an event handler in your component.
+
+## Sending Actions
+
+In some cases your component needs to define event handlers, perhaps to support
+various draggable behaviors. For example, a component may need to send an `id`
+when it receives a drop event:
+
+```hbs
+{{drop-target action="didDrop"}}
+```
+
+You can define the component's event handlers to manage the drop event.
+And if you need to, you may also stop events from bubbling, by using
+`return false;`.
+
+```app/components/drop-target.js
+export default Ember.Component.extend({
+  attributeBindings: ['draggable'],
+  draggable: 'true',
+
+  dragOver: function() {
+    return false;
+  },
+
+  drop: function(event) {
+    let id = event.dataTransfer.getData('text/data');
+    this.sendAction('action', id);
+  }
+});
+```
+
+## Event Names
+
+The event handling examples described above respond to one set of events.
+The names of the built-in events are listed below. Custom events can be
+registered by using [Ember.Application.customEvents][customEvents].
+
+Touch events:
+
+* `touchStart`
+* `touchMove`
+* `touchEnd`
+* `touchCancel`
+
+Keyboard events
+
+* `keyDown`
+* `keyUp`
+* `keyPress`
+
+Mouse events
+
+* `mouseDown`
+* `mouseUp`
+* `contextMenu`
+* `click`
+* `doubleClick`
+* `mouseMove`
+* `focusIn`
+* `focusOut`
+* `mouseEnter`
+* `mouseLeave`
+
+Form events:
+
+* `submit`
+* `change`
+* `focusIn`
+* `focusOut`
+* `input`
+
+HTML5 drag and drop events:
+
+* `dragStart`
+* `drag`
+* `dragEnter`
+* `dragLeave`
+* `dragOver`
+* `dragEnd`
+* `drop`
+
+[customEvents]: http://emberjs.com/api/classes/Ember.Application.html#property_customEvents
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -94,6 +94,8 @@
       url: "customizing-a-components-element"
     - title: "Sending Actions from Components to Your Application"
       url: "sending-actions-from-components-to-your-application"
+    - title: "Handling Events"
+      url: "handling-events"
 
 - title: "Controllers"
   url: 'controllers'


source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    alert("XClickableComponent was clicked!");
+  }
+});
+```
+
+Browser events may bubble up from the target component to each parent component
```
locks: this one is tricky, because they bubble up the DOM, not up component hierarchy. this matters for example for block components.

source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+  click: function(evt) {
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
+### Event Names
+
+  All of the event handling approaches described above respond to the same set
```
locks: There's only the one approach, no?

source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
+### Event Names
```
locks: Move it to the bottom please

source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+  * `dragEnter`
+  * `dragLeave`
+  * `dragOver`
+  * `dragEnd`
+  * `drop`
+
+###Sending Actions
+
+To have the click event from `x-clickable` affect the state of your application,
+simply send an action to the component's template context (the controller):
```
locks: remove ´(the controller)`, it might be another component

source/components/handling-events.md
pixelhandler: true, and that dom includes the parent components so need to update this verbage

source/components/handling-events.md
pixelhandler: There is bubbling and not bubbling (as approaches)

source/components/handling-events.md
locks: I believe the headers should be `##`, check some of michael's PRs for that

source/components/handling-events.md
locks: I would also prefer if this paragraph was phrased around converting events to actions, the current description is a bit jargony.

pixelhandler: maybe parent component or template's controller? Should anything be mentioned about targetObject here ?

source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+
+````app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    this.sendAction('turnItUp', 11);
+  }
+});
+````
+
+If the template's controller has an action handler called `turnItUp`, it will
```
locks: Introduce the concept in the general (hence https://github.com/emberjs/guides/pull/851/files#r41076912), then introduce the specific example so we have context.

source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+export default Ember.Controller.extend({
+  actions: {
+    turnItUp: function(level){
+      //Do your thing
+    }
+  }
+});
+````
+
+If it doesn't, the message will be passed to the current route:
```
locks: I would remove this, as I believe action bubbling is already explained elsewhere, and you also link below.

source/components/handling-events.md
```diff
@@ -0,0 +1,130 @@
...
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
+### Event Names
+
+  All of the event handling approaches described above respond to the same set
+  of events. The names of the built-in events are listed below. (The hash of
```
locks: I don't like the leak of implementation detail here (both `Ember` classes), I would remove it.
It might be nice to have a section of registering custom events, pending team consensus.

source/components/handling-events.md
pixelhandler: See https://raw.githubusercontent.com/emberjs/guides/master/source/components/sending-actions-from-components-to-your-application.md I noticed three (3) `###`

pixelhandler:
@locks I pushed a commit based responses to your review. I can squash once we're good to merge :)

source/components/handling-events.md
```diff
@@ -0,0 +1,125 @@
...
+
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+###Sending Actions
```
locks: h2, not h3 (confirm @michaelrkn)

source/components/handling-events.md
```diff
@@ -0,0 +1,125 @@
...
+
+  * `dragStart`
+  * `drag`
+  * `dragEnter`
+  * `dragLeave`
+  * `dragOver`
+  * `dragEnd`
+  * `drop`
+
+
```
locks: unnecessary extra space

source/components/handling-events.md
```diff
@@ -0,0 +1,125 @@
...
+export default Ember.Controller.extend({
+  actions: {
+    turnItUp: function(level){
+      //Do your thing
+    }
+  }
+});
+````
+
+#### Actions and Bubbling
```
locks: h3 https://github.com/emberjs/guides/pull/851/files#r41079505

source/components/handling-events.md
```diff
@@ -0,0 +1,125 @@
...
+
+Read more about how actions bubble and how to send actions to your application:
+
+- [Sending Actions from Components to Your Application](../sending-actions-from-components-to-your-application)
+- [Action Bubbling](../../templates/actions/#toc_action-bubbling)
+
+
+### Event Names
```
locks: h2 https://github.com/emberjs/guides/pull/851/files#r41079505

source/components/handling-events.md
```diff
@@ -0,0 +1,125 @@
...
+
+  The event handling examples described above respond to one set of events.
+  The names of the built-in events are listed below. Custom events can be
+  registered by using `Ember.Application.customEvents`.
```
locks: can you actually link to the API page here?

source/components/handling-events.md
```diff
@@ -0,0 +1,125 @@
...
+
+````app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    this.sendAction('turnItUp', 11);
+  }
+});
+````
+
+By default a component's `targetObject` is the template's controller instance.
```
locks: nnooo, nonono

source/components/handling-events.md
locks: I just remembered, we're moving towards closure actions so I'd not sure how that would affect this.

source/components/handling-events.md
pixelhandler: Oh so targetObject is private now, http://emberjs.com/api/classes/Ember.Component.html#property_targetObject oops.

source/components/handling-events.md
pixelhandler: I don't think closure actions should not be covered here

locks: I'm so confused by the double negative xD

pixelhandler:
@locks I changed up the context of the event narrative, I hope it does a better job of illustrating how/why to use event handlers in a component.

locks:
lmgt

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
+Instead of having to register event listeners on elements you'd like to
+respond to, simply implement the name of the event you want to respond to
+as a method on your component.
```
michaelrkn: I'd suggest a more positive introduction, like "You can respond to user events on your component like double-clicking, hovering, and key presses through event handlers." Starting out by contrasting Ember to another approach feels negative and assumes the reader comes from a particular background.

I think it would better to introduce event handling as a solution to a problem, rather than as an alternative to

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+Let's implement `x-clickable` such that when it is
+clicked, an alert is displayed:
+
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    alert("XClickableComponent was clicked!");
+  }
+});
+```
```
michaelrkn: Since clicking is the most common action and can be handled in other ways, I'd use doubleClick or some other event to make the use case for this feature more obvious.

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+{{x-draggable}}
+```
+
+You can define the component's event handlers to manage the drop event.
+And if you need to, you may prevent some events from bubbling as well.
+
+```app/components/x-draggable.js
+export default Ember.Component.extend({
+  attributeBindings: ['draggable'],
+  draggable: 'true',
```
michaelrkn: What's this `draggable: 'true'` thing? Is `"true"` supposed to be a string?

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
```
michaelrkn: Should we maybe talk about `event.preventDefault()` here?

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+  This is a clickable area!
+{{/x-clickable}}
+```
+
+Let's implement `x-clickable` such that when it is
+clicked, an alert is displayed:
+
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
```
michaelrkn: Personally I'm partial to not abrvng things, especially when it only saves 2 letters (`evt` vs `event`).

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
+Any registered event can be defined as an event handler in your component.
+(See The list of event names at the end of this page.)
```
michaelrkn: "The" should not be capitalized.

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+```app/components/x-clickable.js
+export default Ember.Component.extend({
+  click: function(evt) {
+    Ember.Logger.info("XClickableComponent was clicked!");
+    return true;
+  }
+});
+```
+
+Any registered event can be defined as an event handler in your component.
```
michaelrkn: What's a "registered event"? Maybe just change to "See the end of this page for a full list of event names." or something.

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Sending Actions
+
+In some cases your component needs to define event handlers, perhaps to support
+various draggable behaviors. For example, a component may need to send an `id`
+when it receives a drop event:
+
+```hbs
+{{x-draggable action="didDrop"}}
+  {{! …an element as a drop affordance may go in here… }}
```
michaelrkn: What's an "affordance"?

source/components/handling-events.md
```diff
@@ -0,0 +1,129 @@
...
+
+  drop: function(evt) {
+    let id = evt.dataTransfer.getData('text/data');
+    this.sendAction('action', id);
+  },
+  // …any other event handling your draggable component needs…
+});
+```
+
+### Actions and Bubbling
```
michaelrkn: I think these sections are easy enough to find without linking to them here.

michaelrkn:
@pixelhandler Thanks so much for putting all this together! I added a few comments with feedback.

source/components/handling-events.md
pixelhandler: @michaelrkn see http://www.html5rocks.com/en/tutorials/dnd/basics/ its an attribute of the element `draggable="true"`

source/components/handling-events.md
pixelhandler: Or just `return false;`

source/components/handling-events.md
pixelhandler: these can be removed

source/components/handling-events.md
pixelhandler: a ui term for an element that you can interact with, I'll remove it, it distracts from the topic

josemarluedke:
This is looking good! Thank you for taking care of this.

source/components/handling-events.md
locks: @pixelhandler you may be thinking of `event.stopPropagation();`.
`return false;` is for actions.

pixelhandler: In vanilla JS return false works, right?

PR #852, opened 2015-10-03
Add workaround for unversioned search links in Swiftype overlay

### LINKED ISSUE #514: Search results have unversioned links
1. Go to http://guides.emberjs.com/. 
2. Type a term like "Ember" in the search box. Don't press the down arrow key. Press Enter. A modal opens.
3. Click a link. 

Expected behavior: You are taken to a URL with the version number, eg http://guides.emberjs.com/v1.13.0/getting-started/. 

Actual behavior: You are taken to a URL with the version number, eg http://guides.emberjs.com/getting-started.

This issue is not present if you enter a search term, use the down arrow to navigate through the results, and press Enter.

This issue is present in the 1.13 and 1.12 guides, making me think (/hope) that I did not introduce it myself when updating search.

Versions 1.10 and 1.11 of the guides simply refresh the page when you press Enter without pressing the down arrow... I'm not sure what's going on there.

kevinkucharczyk:
This temporarily resolves #514 

This will modify links in the Swiftype overlay whenever they appear, adding the currentRevision to the URL. It's a temporary workaround which will hopefully be removed in the future, once we find a more permanent solution.

### DIFF
--- a/source/javascripts/app/swiftype.js.erb
+++ b/source/javascripts/app/swiftype.js.erb
@@ -37,4 +37,15 @@ $(function() {
     documentTypes: Swiftype.searchDocumentTypes,
     onComplete: onComplete
   });
+
+  $(document).on('DOMNodeInserted', '.st-result', function() {
+    var searchLabel = $(this).find('h3');
+    var searchLink = $(this).find('a');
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thank you @kevinkucharczyk!

PR #859, opened 2015-10-07
Add additional explanation/warning around running tests with ember s

### LINKED ISSUE #857: Explain the difference between running your tests via ember server and ember test --server
Interesting discussion in the Slack testing channel today where I learned that running your tests through ember server (via localhost:4200/tests)  could give you different results then running them through ember test --server.  

Really the difference comes down to `ember server` runs in `environment=development`, while running via `ember test --server` runs in `environment=test`.  This can have runtime ramifications, especially if you load some libraries only in test mode.

There's already a blurb about being able to run with /tests in the testing guide intro.  We just need to elaborate a bit to let people know the difference and to steer them towards using ember test.

I'll work on a PR to accompany this...

toddjordan:
Fixes #857

### DIFF
--- a/source/testing/index.md
+++ b/source/testing/index.md
@@ -64,9 +64,15 @@ third-party addons.
 ### How to Run Your Tests
 
 Run your tests with `ember test` on the command-line. You can re-run your tests on every
-file-change with `ember test --server`. Another way to run the tests when you are running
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thanks @toddjordan!

PR #866, opened 2015-10-08
Link to store.peekRecord method description

acorncom:
If we want to do this type of thing, I can run through the rest of this page adding in appropriate links as well ...


locks:
I like it! Can you trawl for the other ones and update the PR accordingly?

acorncom:
Yep, will do.  Aiming for tomorrow on this.

PR #867, opened 2015-10-09
Add warning about viewing an old version of the guides (v1.13.0)

SaladFork:
Based on [676c867](https://github.com/emberjs/guides/commit/676c867d9395374d26f4589eff0fd46d5e28fb63).


locks:
We still need to rebase 1.13.0 to remove `{{each-in` and `{{get`.

locks:
The `v1.13.0` branch doesn't seem to have the helpers, and now I understand nothing.

PR #868, opened 2015-10-09
Add warning about viewing an old version of the guides (v1.12.0)

SaladFork:
A cherry pick of [7144d62](https://github.com/SaladFork/emberjs-guides/commit/7144d62b7db5c1586cf2406644563d80f3c66779) with modifications for different templating approach in this version.


SaladFork:
Ran tests on `v1.12.0` branch (without this commit/PR) locally and it looks like it fails the same way.

SaladFork:
Fixed a broken test caused by a change in error format. Thanks to @locks.

PR #869, opened 2015-10-09
Add warning about viewing an old version of the guides (v1.11.0)

SaladFork:
A cherry pick of [41f1395](https://github.com/emberjs/guides/commit/4ef139578a37aee6e9b963ec6ef48c3254d39927) with modification for version number.


locks:
:+1:

PR #875, opened 2015-10-13
Added links to API details on models pages

acorncom:
I've removed the code back ticks, as it was hard to distinguish that we had links with our old `code` styling.

But I've got a CSS suggestion.  Here's a picture of what we've got and what I'm proposing:
screen shot 2015-10-12 at 9 11 25 pm

Thoughts?  I'll update the CSS and amend this commit if we like that idea ...

### DIFF
--- a/source/models/creating-updating-and-deleting-records.md
+++ b/source/models/creating-updating-and-deleting-records.md
@@ -1,6 +1,8 @@
 ## Creating Records
 
-You can create records by calling the `createRecord` method on the store.
+You can create records by calling the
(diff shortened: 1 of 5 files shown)


michaelrkn:
I like it! Do it!

acorncom:
@michaelrkn I've done this for the models section, might be worth opening a meta issue to have folks run through all the guides applying the same type of styling / linking for the rest of the guides ...

michaelrkn:
Sweet! Can you open an issue to do this globally?

PR #876, opened 2015-10-13
Document preferences on how to link to API methods

acorncom:
A question came up as I was adding links to API pages (https://github.com/emberjs/guides/pull/875) about how we want to link to methods.  @locks suggested I open this PR so we can come up with a standard approach to this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,6 +37,12 @@ In code samples:
 * In fenced code blocks, include the filename or language after the triple-backticks, like <code>&#96;&#96;&#96;routes/kittens.js</code> or <code>&#96;&#96;&#96;hbs</code>.
 * Write paths relative to the project root.
 
+When linking to API pages:
+
+* use code backticks as part of the link text, i.e., \[<code>&#96;store.push()&#96;</code>]
+(api link)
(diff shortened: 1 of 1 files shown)


locks:
My knee jerk reaction is that I don't like adding the parenthesis that much because it might induce the error that the function doesn't have params.

acorncom:
I'm happy doing either, just think we should have it consistent and documented (I've seen both :-)

michaelrkn:
> it might induce the error that the function doesn't have params

Can you elaborate? I don't understand this problem.

locks:
> Can you elaborate? I don't understand this problem.

That `this.store()` doesn't have any arguments. I know it's silly, but… :P

michaelrkn:
Oh, I see. But it still needs to be called with `()`... I personally don't see this as confusing.

acorncom:
I'd been thinking we would only add the `()` to methods that needed it when you called them (instead of properties).  But I'm fine either direction.  Mainly was interested in clarifying which way we do this, as we've got an interesting mix on the site right now ;-)

michaelrkn:
Right. Only Ruby developers would be confused by invoking an argument-less function with parens ;)

acorncom:
Oh, is that the issue?  I'm not super familiar with Ruby, so that escaped me :-)

PR #884, opened 2015-10-15
Update customizing-serializer.md

jiezhang917:
change the wrong word "amount" in the second "payload.data.attributes.amount" to "payload.data.attributes.currency".


michaelrkn:
Thanks!!

PR #888, opened 2015-10-15
Add blurb about {{get}} helper

### LINKED ISSUE #861: Add section about get and mut helpers on input helpers
Just had a [discussion](https://embercommunity.slack.com/archives/needhelp/p1444223216002500) with @rwjblue and @kamilbiela on the Slack channel about using the `(get)` helper with an `{{input}}` helper and requiring the `(mut)` helper to enable two way binding. Although there is an example in the [API docs](http://emberjs.com/api/classes/Ember.Templates.helpers.html#method_get), I think it would be helpful to make note of this situation in the guides, such as on the [Input Helpers page](http://guides.emberjs.com/v2.1.0/templates/input-helpers/). If I get some spare time, I could write a quick PR to add a mention, but wanted to add a note here before I forgot.

michaelrkn:
Fixes #861.

### DIFF
--- a/source/templates/input-helpers.md
+++ b/source/templates/input-helpers.md
@@ -110,3 +110,19 @@ Will bind the value of the text area to `name` on the current context.
 * `form`
 * `spellcheck`
 * `required`
+
+### The `{{get}}` Helper
+
+The [`{{get}}`](https://github.com/emberjs/guides/issues/861) helper allows you to
+dynamically look up a property on an object. For example, these two usages are equivilent:
+
+```handlebars
+{{person.height}}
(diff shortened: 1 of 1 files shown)


rwjblue:
Since this is the input helpers section, can the example use inputs?

michaelrkn:
Yes but can you give me an example so I don't have to think of it and then
make sure it works? :)

rwjblue:
Ya, definitely. I'll put together a twiddle when I'm back at my computer.

michaelrkn:
Closed in favor of #891.

PR #890, opened 2015-10-16
[input-helpers] Tweaks header levels

locks:


mixonic:
Is there some context to this that motivates the change? I'd be happy to merge stuff like this if there was some information about what is happening.

locks:
@mixonic I believe this is in line with previous edits: https://github.com/emberjs/guides/commit/0c1f977de385619736a29b2e18baa5ce21a53563.

The first header is superfluous regardless of the level, since it gets repeated due to the section name being the same.

mixonic:
Seems good!

locks:
:beers: :smile:

Issue #893, opened 2015-10-17
Observer doesn't fire in`Getting started / Observers` 


ryrych:
I tried to run the [first example](http://guides.emberjs.com/v2.1.0/object-model/observers/) both with Chrome snippet and 'regular' console, but every time I change `firstName` or `lastName` the `fullNameChanged` doesn't fire:

```
  fullNameChanged: Ember.observer('fullName', function() {
    // deal with the change
    console.log('fullName changed to', this.get('fullName'));
  })
```

acorncom:
@ryrych Further down the page, it mentions

> If you never get a computed property, its observers will not fire even if its dependent keys change. You can think of the value changing from one unknown value to another.

Does that fix your issue?  Might be worth asking on Stackoverflow unless you're sure this is a bug in Ember / guides

PR #895, opened 2015-10-17
Make clear that `fullNameChanged` fires if `fullName` is consumed first

ryrych:
Fixes https://github.com/emberjs/guides/issues/893#issuecomment-148925813

### DIFF
--- a/source/object-model/observers.md
+++ b/source/object-model/observers.md
@@ -22,6 +22,7 @@ Person = Ember.Object.extend({
 
   fullNameChanged: Ember.observer('fullName', function() {
     // deal with the change
+    console.log(`fullName changed to: ${this.get('fullName')}`);
   })
 });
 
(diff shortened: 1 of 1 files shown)


locks:
Hm, I actually don't like how the original example complicates the explanation.
Thoughts @michaelrkn?

source/object-model/observers.md
```diff
@@ -30,7 +31,9 @@ var person = Person.create({
   lastName: 'Katz'
 });
 
-person.set('firstName', 'Brohuda'); // observer will fire
+// observer won't fire until `fullName` is consumed first
```
locks: what do you think of switching it around to `// consume "fullName" so the observer is triggered`?

locks:
p.s. sorry for slightly editing your post, but I think this way the issue will get closed when this gets merged. And thanks for your contribution!

ryrych:
@locks no problem, you can switch the order :+1:

michaelrkn:
I think it's pretty clear how it is. @locks why don't you open another PR if you think there's a better way to explain it.

Issue #887, opened 2015-10-15
Add API links throughout Guides


michaelrkn:
Someone should go through all of the Guides and add links to the API docs, following the format at the bottom of the [contributing guidelines](https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md).

locks:
@acorncom

acorncom:
I would suggest we break things up by major section of the guides and have people take them in chunks as desired.  In addition, I was trying to link the first mention of a new concept to the API guides, but in a paragraph where it gets mention 2 or 3 times, it seems redundant to link every single time ...

- [ ] Getting Started
- [ ] The Object Model
- [ ] Routing
- [ ] Templates
- [ ] Components
- [ ] Controllers
- [x] Models (I've already done this one)
- [ ] Application Concerns
- [ ] Testing
- [x] Ember Inspector (done as well)
- [ ] Addons and Dependencies
- [ ] Configuring Ember.js
- [ ] Contributing to Ember.JS

acorncom:
I'm going to recommend skipping linking to APIs on the Controllers pages, as they link to older docs that reference Controllers and Views.

acorncom:
Everything else here is up for grabs.  I'm going to work away on this as I have time, but would love to get help! ;-)

acorncom:
@michaelrkn @locks We're finally finished here.  Can we get this pushed up to the live site, I'd love to see it in action ;-)

locks:
:computer: says :no_entry:

michaelrkn:
@acorncom you are a true hero!!!

PR #896, opened 2015-10-17
Adding in api links to the routing pages

acorncom:
Part of #887


locks:
We need a way to automate this or something :P

acorncom:
It's called increased access :-p

PR #897, opened 2015-10-17
Update classes-and-instances.md

marpo60:
Use string interpolation in all examples


locks:
Thanks for the contribution!

PR #899, opened 2015-10-18
Doc inflector use for model pluralization

### LINKED ISSUE #898: Doc how and where to specify inflectors for pluralization of adapters
Currently the model guide is vague on where or how to handle non-standard pluralization.

EricSchank:
Fixes #898 
Added documentation of how and where to handle pluralization.

### DIFF
--- a/source/models/customizing-adapters.md
+++ b/source/models/customizing-adapters.md
@@ -104,13 +104,22 @@ Data comes bundled with
 [Ember Inflector](https://github.com/stefanpenner/ember-inflector), a
 ActiveSupport::Inflector compatible library for inflecting words
 between plural and singular forms. Irregular or uncountable
-pluralizations can be specified via `Ember.Inflector.inflector`:
+pluralizations can be specified via `Ember.Inflector.inflector`.
+A common way to do this is:
 
-```js
+```app/app.js
+// sets up Ember.Inflector
+import './models/custom-inflector-rules';
+```
+
+```app/models/custom-inflector-rules.js
 var inflector = Ember.Inflector.inflector;
 
 inflector.irregular('formula', 'formulae');
 inflector.uncountable('advice');
+
+// Meet Ember Inspector's expectation of an export
+export default {};
 ```
 
 This will tell the JSON API adapter that requests for `formula`


bmac:
:+1:

locks:
I think I'd rather the person do it inside the model.js file.

michaelrkn:
@bmac @EricSchank do either of you have thoughts on putting it in an initializers vs a model file?

EricSchank:
@michaelrkn 
I prefer having all the inflector changes together in a single file and getting them all setup before the "main" code that relies on them is executed.  But I also see the appeal of having each inflector call in the model that it affects.

rwjblue:
I'd prefer to avoid an initializer for these sorts of things, personally.  I see the following downsides:

* initializers are not run before `moduleForModel` or other non full application acceptance tests
* initializers are ran once for each test, but this setup work is only needed one time

I also completely agree that having these rules all in the same place would be great (as opposed to sprinkled throughout many models). Something like the following would solve both concerns:

```
// in app/app.js

// sets up Ember.Inflector
import './models/custom-inflector-rules';
```

```js
// in app/models/custom-inflector-rules.js
var inflector = Ember.Inflector.inflector;
inflector.irregular('formula', 'formulae');
inflector.uncountable('advice');
```

michaelrkn:
Maybe there should be a CLI command then like `ember generate inflectors` that does the "best practice" setup for you? Otherwise everybody will just do it their own way.

rwjblue:
Also, the last part was just a pattern that I have followed in my apps, but I'm not terribly worried where we put them as long as we aren't using an initializer.

rwjblue:
@michaelrkn - Ya, makes sense.

michaelrkn:
On the docs call today, we agreed that the documentation should be updated now with the code example in https://github.com/emberjs/guides/pull/899#issuecomment-149340100, and then when a generator for inflectors is added to Ember CLI, we will update the Guides again. @EricSchank can you update this PR to be along the lines of https://github.com/emberjs/guides/pull/899#issuecomment-149340100? Thank you!

EricSchank:
@michaelrkn 
Sure.  I'll update the PR.

fsmanuel:
@EricSchank @michaelrkn we need to add

```js
export default {};
```

otherwise the inspector will crash 

```bash
Ember Inspector has errored.
This is likely a bug in the inspector itself.
You can report bugs at https://github.com/emberjs/ember-inspector.
Error message:  Expected to find: 'model:custom-inflector-rules' within 'inflector-bug/models/custom-inflector-rules' but got 'undefined'. Did you forget to `export default` within 'inflector-bug/models/custom-inflector-rules'?
Stack trace: Error:  Expected to find: 'model:custom-inflector-rules' within 'inflector-bug/models/custom-inflector-rules' but got 'undefined'. Did you forget to `export default` within 'inflector-bug/models/custom-inflector-rules'?
    at resolveOther (http://localhost:4200/assets/vendor.js:63230:15)
    at superWrapper (http://localhost:4200/assets/vendor.js:31622:20)
    at exports.default._emberRuntimeSystemObject.default.extend.resolve (http://localhost:4200/assets/vendor.js:15306:35)
    at Object.resolve [as resolver] (http://localhost:4200/assets/vendor.js:15040:23)
    at resolve (http://localhost:4200/assets/vendor.js:12706:29)
    at Object.Registry.resolve (http://localhost:4200/assets/vendor.js:12274:21)
    at factoryFor (http://localhost:4200/assets/vendor.js:11924:28)
    at Object.Container.lookupFactory (http://localhost:4200/assets/vendor.js:11839:14)
    at exports.default._emberRuntimeSystemObject.default.extend._nameToClass (http://localhost:4200/assets/vendor.js:16267:31)
    at http://localhost:4200/assets/vendor.js:16464:25
```

EricSchank:
Thanks @fsmanuel -- I've added that to the change now.

michaelrkn:
Thanks @EricSchank!

figwit:
Following the recommendation in this thread, given this:

```js
inflector.irregular('rota', 'rotas');
```

I'd expect `store.find('rota')` to result in `GET /rotas` but instead you get `GET /rota`. However, with the following:

```js
// app/models/department.js
rotas: DS.hasMany('rota', {async: true});
```

It's all rosy – `department.rotas` results in `GET /departments/:department_id/rotas`.

Of note, `Ember.Inflector.inflector.pluralize('rota')` returns, `"rota"`, which is wrong. And if you instead use an application initializer to set-up inflection rules then `Ember.Inflector.inflector.pluralize('rota')` returns `"rotas"` (win), `store.find('rota')` produces `GET /rotas` (win), but the model definition results in `No model was found for 'rotum'`.

Hair.pull.out.

This is `v1.13.x` btw.

figwit:
UPDATE: tested on a clean slate (using `ember-cli@1.13.15`) and my scenarios above work as expected using the method outlined in this PR...clearly I or an app dependency fubar'ed something along the way :|

fwiw, this method doesn't appear to work against a new `ember-cli@2.3.0-beta.1` app, resulting in `Uncaught TypeError: Cannot read property 'inflector' of undefined` (`Ember.Inflector` being `undefined`). Using an initializer does work, however.

My initializer, for reference:

```js
// app/initializers/inflections.js
import Ember from 'ember';

export function initialize(/* application */) {
  var inflector = Ember.Inflector.inflector;

  inflector.irregular('rota', 'rotas');
}

export default {
  name: 'inflections',
  initialize
};
```

Subtletree:
Getting this in Ember 2.8.0 too. 

    var inflector = Ember.Inflector.inflector;  //=> Uncaught TypeError: Cannot read property 'inflector' of undefined

PR #912, opened 2015-10-23
Update rendering-a-template.md

tylerneu:
I believe this documentation should read that the `posts.new` route will render into the `posts.hbs`'s `{{outlet}}` not `post.hbs`'s.


michaelrkn:
Thanks for fixing this!

PR #913, opened 2015-10-23
Elaborate a little more about `_super()`

grapho:
this gives the guides a bit more specific link target if you want to reference `_super()`

### DIFF
--- a/source/object-model/classes-and-instances.md
+++ b/source/object-model/classes-and-instances.md
@@ -33,6 +33,8 @@ export default Ember.Component.extend({
 });
 ```
 
+####Overriding Parent Class Methods
+
 When defining a subclass, you can override methods but still access the
 implementation of your parent class by calling the special `_super()`
 method:
@@ -59,6 +61,25 @@ var yehuda = Soldier.create({
 
 yehuda.say('Yes'); // alerts "Yehuda Katz says: Yes, sir!"
 ```
+
+In certain cases, you will want to pass arguments to `_super()` before or after overriding.
+
+This allows the original method to continue operating as it normally would.
+
+One common example is when overriding the [`normalizeResponse()`][3] hook in one of Ember-Data's serializers.
+
+A handy shortcut for this is to use a "spread operator", like `...arguments`:
+
+[3]: http://emberjs.com/api/data/classes/DS.JSONAPISerializer.html#method_normalizeResponse
+
+```javascript
+normalizeResponse(store, primaryModelClass, payload, id, requestType)  {
+  // Customize my JSON payload for Ember-Data
+  return this._super(...arguments);
+}
+```
+
+The above example returns the original arguments (after your customizations) back to the parent class, so it can continue with its normal operations.
 
 ### Creating Instances
 


source/object-model/classes-and-instances.md
```diff
@@ -33,6 +33,8 @@ export default Ember.Component.extend({
 });
 ```
 
+####Overwritig Subclass Methods
```
Dhaulagiri: Overwriting

source/object-model/classes-and-instances.md
```diff
@@ -59,6 +61,13 @@ var yehuda = Soldier.create({
 
 yehuda.say('Yes'); // alerts "Yehuda Katz says: Yes, sir!"
 ```
+In certain cases, you will want to pass arguments to `_super()` before or after overriding, so that the original method can continue operating as it normally would.  One common example is when overriding the `setupController()` hook in a route.  A handy shortcut for this is to use a "spread operator", like `...arguments`:
```
michaelrkn: Can you:

* break this onto multiple lines;
* add a line break before the code fences;
* change the example to something that doesn't involve a controller, since controllers are going away?

Thanks!

michaelrkn:
This looks great! I just added a couple requests; if you can address those and squash your commits, we'll get this merged in!

source/object-model/classes-and-instances.md
```diff
@@ -2,7 +2,7 @@ As you learn about Ember, you'll see code like `Ember.Component.extend()` and
 `DS.Model.extend()`. Here, you'll learn about this `extend()` method, as well
 as other major features of the Ember object model.
 
-### Defining Classes
+### Defining Classes and Subclasses
```
rwjblue: Isn't every class a subclass :P

rwjblue:
We should also discuss the return value from this._super (especially for methods that can return promises).

grapho:
@michaelrkn  good point about the example, the only other example i can draw on (from my own experince) would be overriding Serializer's `normalizePayload()`, do you think that would be a better long-term example? and @rwjblue would that example also illustrate the "return value" of a _super?

rwjblue:
>  @rwjblue would that example also illustrate the "return value" of a _super?

Yep

michaelrkn:
Looks good! Can you add line breaks, change two spaces after periods to one, and squash your commits?

grapho:
@michaelrkn  to be clear, when you say "line breaks" you actually mean new lines? I have already been trying to break long sentences up into smaller ones? But you prefer to break them into new lines?

source/object-model/classes-and-instances.md
grapho: yeah i guess that is a good point

michaelrkn:
@grapho yes, sorry, new lines.

grapho:
@michaelrkn  here is a final look at the PR, do you see any last minute edits?

source/object-model/classes-and-instances.md
```diff
@@ -33,6 +33,8 @@ export default Ember.Component.extend({
 });
 ```
 
+####Overriding Subclass Methods
```
michaelrkn: Shouldn't this be "Overriding Parent Class Methods"?

source/object-model/classes-and-instances.md
```diff
@@ -60,6 +62,21 @@ var yehuda = Soldier.create({
 yehuda.say('Yes'); // alerts "Yehuda Katz says: Yes, sir!"
 ```
 
+In certain cases, you will want to pass arguments to `_super()` before or after overriding, so that the original method can continue operating as it normally would.
```
michaelrkn: Please break lines before 120 characters.

michaelrkn:
We're getting real close! Just a couple comments for you. Looks like you got the commit squashing figured out :)

michaelrkn:
Thanks @grapho!

PR #914, opened 2015-10-23
Ex. Objects & arrays leaking across instances

grapho:
This illustrates how arrays and objects can leak across instances if not initialized properly.

### DIFF
--- a/source/object-model/classes-and-instances.md
+++ b/source/object-model/classes-and-instances.md
@@ -131,6 +131,57 @@ override the `init()` method, make sure you call `this._super(...arguments)`!
 If you don't, a parent class may not have an opportunity to do important
 setup work, and you'll see strange behavior in your application.
 
+Arrays and objects defined directly on any `Ember.Object` are shared across all instances of that object. 
+```js
+Person = Ember.Object.extend({
+  shoppingList: ['eggs', 'cheese']
+});
+
+Person.create({
+  name: 'Stefan Penner',
+  addItem() {
+    this.get('shoppingList').pushObject('bacon');
+  }
+});
+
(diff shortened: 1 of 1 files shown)


source/object-model/classes-and-instances.md
```diff
@@ -131,6 +131,54 @@ override the `init()` method, make sure you call `this._super(...arguments)`!
 If you don't, a parent class may not have an opportunity to do important
 setup work, and you'll see strange behavior in your application.
 
+Arrays and objects defined directly on any Ember.Object are shared across all instances of that object.
```
michaelrkn: add a line break below

michaelrkn: add backticks around Ember.Object

source/object-model/classes-and-instances.md
```diff
@@ -131,6 +131,54 @@ override the `init()` method, make sure you call `this._super(...arguments)`!
...
+Person.create({
+  name: 'Robert Jackson',
+  addItem() {
+    this.get('shoppingList').pushObject('sausage');
+  }
+});
+
+// If Stefan and Robert trigger their addItem, they will both end up with ['eggs', 'cheese', 'bacon', 'sausage']
+```
+To avoid this behavior, it is encouraged to initialize arrays and object properties during init() if they are meant to be unique to each instance.
```
michaelrkn: Add backticks around init(). Also break this onto two shorted lines.

michaelrkn:
Looks good! Just added a couple comments.

michaelrkn:
Can you squash your commits so we can get this merged in? Thanks!

PR #916, opened 2015-10-25
Add HTML Proofer

### LINKED ISSUE #598: Test guides for dead links
We often find that guides end up with dead links as a result of reorganizing content. It would great to have something that would raise an exception if Middleman tries to serve or build a page with a dead link - serve so that you catch it in development if you're working on that page, build so that it gets caught by Travis making sure it builds if you delete a page that somewhere else links to.

bmichelsen:
The HTML Proofer is a gem that validates HTML output. It checks for lots of things, for example whether image references are correct, and that images have alt tags. The main reason we're looking into it, is because it checks for dead links, which is great after reorganizing content.

Have a look at the discussion over at #598, as well as the [html-proofer options](https://github.com/gjtorikian/html-proofer#configuration).

There are still a few issues remaining, which I'm hoping to get feedback on:

- What other options should be enabled? I'm thinking ```check_html``` might be good, unless it's completely out of scope.
- The PR in its current form should have Travis complaining about dead links. However, the ```middleman server``` command still needs work.

Fixes #598.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,7 @@ gem "builder"
 gem "middleman-alias"
 gem "middleman-swiftype", :git => "https://github.com/LeonB/middleman-swiftype.git"
 gem "underscore-rails"
+gem "html-proofer"
 
 source 'https://rails-assets.org' do
   gem "rails-assets-js-md5"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -24,6 +24,7 @@ GEM
       minitest (~> 5.1)
       thread_safe (~> 0.1)
       tzinfo (~> 1.1)
+    addressable (2.3.8)
     builder (3.2.2)
     capybara (2.4.4)
       mime-types (>= 1.16)
@@ -39,6 +40,7 @@ GEM
       coffee-script-source
       execjs
     coffee-script-source (1.9.0)
+    colored (1.2)
     compass (1.0.3)
       chunky_png (~> 1.2)
       compass-core (~> 1.0.2)
@@ -54,6 +56,8 @@ GEM
     daemons (1.1.9)
     diff-lcs (1.2.5)
     erubis (2.7.0)
+    ethon (0.8.0)
+      ffi (>= 1.3.0)
     eventmachine (1.0.7)
     execjs (2.3.0)
     ffi (1.9.6)
@@ -67,6 +71,14 @@ GEM
     hitimes (1.2.2-x86-mingw32)
     hooks (0.4.0)
       uber (~> 0.0.4)
+    html-proofer (2.5.1)
+      addressable (~> 2.3)
+      colored (~> 1.2)
+      mercenary (~> 0.3.2)
+      nokogiri (~> 1.5)
+      parallel (~> 1.3)
+      typhoeus (~> 0.7)
+      yell (~> 2.0)
     i18n (0.6.11)
     json (1.8.3)
     kramdown (1.5.0)
@@ -74,6 +86,7 @@ GEM
       celluloid (>= 0.15.2)
       rb-fsevent (>= 0.9.3)
       rb-inotify (>= 0.9)
+    mercenary (0.3.5)
     method_source (0.8.2)
     middleman (3.3.7)
       coffee-script (~> 2.2)
@@ -119,6 +132,7 @@ GEM
       tilt (~> 1.4.1)
     padrino-support (0.12.4)
       activesupport (>= 3.1)
+    parallel (1.6.1)
     poltergeist (1.6.0)
       capybara (~> 2.1)
       cliver (~> 0.3.1)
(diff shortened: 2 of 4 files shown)


michaelrkn:
This is fantastic! And so awesome (or maybe not awesome) to see Travis failing the PR because this has already found dead links. 

Re: other options, `check_html` sounds nice, but at the end of the day, most of the HTML gets generated from Markdown, so hopefully it's all nice and clean. Try enabling it yourself and see if you get a bunch of noise or if it's useful. No other options jump out at me, but if there's anything else you'd like to discuss, let's do so here.

Re: `middleman server`, can you try using:

```ruby
configure :development do
  activate :html_proofer
end
```

and see if this causes exceptions using the development server?

Finally, can you squash your commits, and then also add a commit to fix the problems HTML Proofer has identified, so that Travis is green again?

Thank you for your great work!

michaelrkn:
Hey @bmichelsen - any progress on this?

bmichelsen:
The following errors have been identified by the HTML Proofer:

```
- ./build/applications/initializers/index.html
  *  internally linking to ./dependency-injection, which does not exist (line 87)
```

```
- ./build/configuring-ember/embedding-applications/index.html
  *  internally linking to ../../routing/specifying-the-location-api, which does not exist (line 114)
  *  internally linking to ../../templates/the-application-template, which does not exist (line 88)
```

```
- ./build/ember-inspector/troubleshooting/index.html
  *  linking to internal hash #toc_using-the-inspector-with-jsbin that does not exist (line 99)
```

```
- ./build/models/customizing-adapters/index.html
  *  internally linking to /api/classes/Ember.ComputedProperty.html#method_volatile, which does not exist (line 382)
  *  trying to find hash of /api/classes/Ember.ComputedProperty.html#method_volatile, but /Users/bjorn/Source/Projects/emberjs-guides/build/api/classes/Ember.ComputedProperty.html does not exist (line 382)
```

```
- ./build/testing/unit-testing-basics/index.html
  *  internally linking to ../unit-test-helpers, which does not exist (line 157)
```

With one exception (the first one referring to line 87), I have a hard time figuring out where these URLs should link to. Are you able to help out? I'm happy to have a discussion on chat in order to have this fixed, just let me know.

Also, links pointing to ```/api``` are failing as well, but they are correct (at least in this case). Should we ignore ```/api``` links or is there a better way to deal with them?

bmichelsen:
I'd like to skip the ```check_html```option due to scope since the HTML is automatically being generated by Markdown.

I've been doing some thinking about running the HTML Proofer in development mode, with the hopes of it causing exceptions using the development server. Before I look more into it, I'd like to express a concern with regards to development speed. If we activate the Proofer during development, then every time the file is saved, the Proofer will run. For me, that resulted in a 2 seconds delay which I find unacceptable. I'm sure there are ways we can optimize this, but I'm not too sure it's the way forward.

Any thoughts? :-)

michaelrkn:
Thanks for the update! I've opened #965 to fix the broken links. 

Skipping `check_html` is fine with me. There is the chance that somebody writes borked HTML in their Markdown, but let's cross that bridge when we come to it.

I agree about not slowing down development speed. If we just have the Proofer check for internal links in development mode, and only external links in CI (or not at all), would that make the speed more acceptable?

bmichelsen:
Hm, the Proofer is only checking internal links at the moment so that won't speed things up unfortunately.

michaelrkn:
I just worry that if we don't have this in development, we'll get a lot of PRs where people didn't realize the build was failing, and we'll have to do the whole back-and-forth thing. @locks do you have any thoughts? 

I'm trying to configure this to run in dev so I can get a sense of the slowdown, but I'm having trouble figuring out how to configure it properly. @bmichelsen can you point me in the right direction?

michaelrkn:
@bmichelsen - just want to ping you on getting some help setting this up in development. Thanks!

bmichelsen:
In order to get a sense of how the Proofer affects development, please do the following:

1. Download [a.patch.txt](https://github.com/emberjs/guides/files/43862/a.patch.txt)
2. Inside the project, run ```git checkout -b proofer-takes-forever```
3. Then ```git apply a.patch.txt``` within the newly created branch
4. Run ```middleman server```
5. Go to ```http://localhost:4567/routing/rendering-a-template/index.html```
6. Reload the page

When you're done experimenting, do a ```git checkout .``` to restore the working directory.

bmichelsen:
I think this gets us closer to the goal, but I also see your point about avoiding the back-and-forth. What if we update the documentation on how to contribute, merge the changes and see if it becomes a hassle? If it does, we'll simply revert it and know for sure that it's the wrong approach. :-)

michaelrkn:
Boom! Merged :) Thanks @bmichelsen! In case we do want to enable this in development in the future, do you have any pointers on how it might work to have it more generically check every page as it's built? (Just in case you're not around to work on it more at that point.)

PR #920, opened 2015-10-27
Fixed testing-components doc related to ember-qunit

dsaenztagarro:
Documentation was outdated. Verified while I was trying to create some acceptance tests for components. Ember-qunit documentation is the correct path.

### DIFF
--- a/source/testing/testing-components.md
+++ b/source/testing/testing-components.md
@@ -45,14 +45,16 @@ component's `style` attribute and is reflected in the  rendered HTML:
 test('should change colors', function(assert) {
   assert.expect(2);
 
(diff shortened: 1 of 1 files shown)


michaelrkn:
@dsaenztagarro thank you for updating this! @toddjordan can you please help review?

rwjblue:
This section is describing component integration tests. Calling this.set is definitely appropriate in component integration tests. These changes swap it to be a unit test (which is when you call this.subject), which we do not really want folks to do (at least not as part of the default happy path).

See https://github.com/rwjblue/ember-qunit#component-integration-tests for the ember-qunit docs on integration tests.

PR #921, opened 2015-10-27
Fixed testing-components doc related to ember-qunit

dsaenztagarro:


rwjblue:
This section is describing component integration tests.  Calling `this.set` is definitely appropriate in component integration tests.  These changes swap it to be a unit test (which is when you call `this.subject`), which we do not really want folks to do (at least not as part of the default happy path).

dsaenztagarro:
I suppose I am missing something... When I create integration tests for components `this.set` is always `undefined`, doesn't matter if attribute `integration` is `true`... Could anyone help me?

PR #923, opened 2015-10-27
Fix reference to logger

StevenNunez:


locks:
@StevenNunez what's wrong with the current initializer?

StevenNunez:
Hey @locks, 

Maybe I'm misunderstanding how the initializers work. The example that registers `notification:message` is using the filename `logger.js`, and exporting `logger`. Is the exported `name` tied to the filename, or the registered service?

locks:
The `name` property is the name that the initializer is registered with so you can define the ordering with `before`/`after`, not sure it's related to anything else.
I agree that the names are confusing because the other examples use the concept of logger and this one is a notification message.

Issue #853, opened 2015-10-03
Update component actions


michaelrkn:
Several items didn't get addressed in #834. We should add in this content:

- [x] Threading actions through multiple layers of components
- [x] Passing arguments
- [x] Argument currying
- [x] Return values
- [x] `target`?
- [x] `value`?
- [ ] Bubbling to routes?

szines:
We suggest more places in the guide, that a special event name should use dasherized version. A code snippet for example:

http://emberjs.com/api/classes/Ember.Templates.helpers.html#toc_specifying-dom-event-type
```
<div {{action "anActionName" on="double-click"}}>
  click me
</div>
```

My experience that this feature works only if we use the original camelCase version. Should we update these sections or it is already updated just not released yet?

This instruction is misleading as well.
http://guides.emberjs.com/v2.1.0/templates/actions/#toc_specifying-the-type-of-event

"You should use the dasherized event names, so two-word names like keypress become key-press."

I'm happy to give a hand if it haven't updated yet.

acorncom:
Here's a Twiddle demo showing the problem:

https://ember-twiddle.com/ee3626653d2777a75678

michaelrkn:
@szines thanks for reporting this! If you can open a PR on the Guides and the API docs to fix this that would be awesome.

szines:
@michaelrkn I opened a PR on the Guides. Hope it is fine. Let me know if it should improve further.

I feel, there could be a little confusion, because in the following case, we have to use the dasherized even name: http://guides.emberjs.com/v2.1.0/templates/input-helpers/#toc_actions

michaelrkn:
@szines Hrm, perhaps. I'm not sure what to do about it. We could add a note to the Input Helpers guide just saying that it's different than the `action` helper syntax.

toddjordan:
@michaelrkn I'll spend some time rounding out component actions over the next week or two

michaelrkn:
@toddjordan hooray!

jdmarshall:
The section on bubbling is inaccurate at least in 1.13 and beyond.  You have to have an action on the component or you get an error.  It implies a default behavior you should rely on but that behavior no longer exists, and it provided no alternative.

I'm trying to have one component that has a list of items to select and another that displays the currently selected item, and the guides provide no leads on how that might be accomplished.

michaelrkn:
[...]

PR #926, opened 2015-10-28
Add Closure Actions to actions.md

lmcardle:
Closure Actions were introduced in 1.13 and while there is a great blog post on them (http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_closure-actions) the guides that talk about actions do not mention them. This PR adds a simple explanation of closure actions and why you would use them instead of the traditional method of calling 'sendAction'.

### DIFF
--- a/source/templates/actions.md
+++ b/source/templates/actions.md
@@ -158,3 +158,100 @@ For example:
   cursor: pointer;
 }
 ```
+
+## Closure Actions and Components
+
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thanks for the contribution, @lmcardle! Closure actions are actually [already documented in the Guides](http://guides.emberjs.com/v2.1.0/components/triggering-changes-with-actions/). As noted in the [Contributing Guidelines](https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md), "no attempt is made to update content, layout, or styles for older versions of the Guides"; unfortunately, closure actions weren't documented when the 1.13 Guides were released, so they are only documented going forward. If you're interested in improving the documentation of closure actions, there is [an issue noting several outstanding items](https://github.com/emberjs/guides/issues/853).

PR #927, opened 2015-10-29
Loading event: add reference to transition promise

vmalloc:
Add an example for using the transition promise to effectively use the `loading` event handler

### DIFF
--- a/source/routing/loading-and-error-substates.md
+++ b/source/routing/loading-and-error-substates.md
@@ -84,7 +84,8 @@ export default Ember.Route.extend({
   },
   actions: {
     loading(transition, originRoute) {
-      alert('Sorry this page is taking so long to load!');
+      let controller = this.controllerFor('foo');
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thanks for contributing @vmalloc! Is there any way to do this without using controllers?

michaelrkn:
Also, I think it might be more concise to just replace the example above (`alert('Sorry this page is taking so long to load!');`) with your contribution, which seems more realistic anyways.

acorncom:
Don't think this would be doable without controllers at the moment ...

michaelrkn:
Cool - then @vmalloc if you can just replace the original example with yours, then we'll get this merged in!

vmalloc:
@michaelrkn I changed the upper example as well. Let me know if any other changes are needed.

michaelrkn:
Awesome! Thanks!!

PR #928, opened 2015-10-29
Fix a typo

PavelBastov:


michaelrkn:
Thanks for contributing, @PavelBastov! The first part of the sentence is plural ("translates requests"), so it probably makes more sense to switch the second part to plural as well ("into requests", rather than "into a request"). Can you make that change real quick? Feel free to open a new PR if you don't want to deal with rebasing and squashing :)

PavelBastov:
Yep, you are right.

Created a new pr https://github.com/emberjs/guides/pull/929

PR #938, opened 2015-11-05
Nudge toward 'owner' instead of 'container'

mixonic:
Preliminary stuff to merge before more in depth changes re: https://github.com/emberjs/ember.js/issues/12555


locks:
:beer:

PR #947, opened 2015-11-06
[Redesign] WIP: Setup new header

acorncom:


source/_header.erb
```diff
@@ -1,36 +1,38 @@
...
-      </li>
-    </ul>
+    <li class="header-search">
+      <form>
+        <input type="text" id="st-search-input" class="st-search-input" placeholder="Search the guides"/>
+      </form>
+      <div id="st-results-container"></div>
+    </li>
+    <li>
+        <div id="github">
```
locks: you can kill this

josemarluedke:
I would like to do some suggestions in the design of the header.

- Keep a solid color as the background, without the image.
- Change the background color to the ember orange.
- Change the height of the header to be smaller.

screen shot 2015-11-06 at 10 47 59 am

Not sure if you are open for this kind of suggestions, but I thought it was worth the shot.

acorncom:
@josemarluedke With the exception of dealing with the background image, those are exactly what we've been aiming for.  I just haven't had the time to finish it up yet (but those were the criticism we were all making of the current design).  Do you happen to have time to make the tweaks? ;-)

acorncom:
@josemarluedke If you don't, no worries.  I'll use your screenshot as a guide for the background color (which was the part I was puzzling over)

locks:
@josemarluedke https://github.com/emberjs/core-notes/blob/michaelrkn-patch-4/docs-team/2015-11/november-5.md we're already going in that direction :P

acorncom:
@locks this is ready to be merged, @josemaluedke just did most of the work :-)

I'd like to clean up headers, spacing and remove some CSS under the hood, but those aren't blockers. Let's make this more visible :-)

PR #949, opened 2015-11-07
Just purging

locks:
Triple puns.


michaelrkn:
This looks just fine to me.

PR #951, opened 2015-11-07
Correct initializers guide for v2.0.

arenoir:
In ember 2.0 initializers expect two arguments `container` and `application`.  Also the `applicationInstance` argument of the instance initializers do not have a method `lookup` one must call `lookup` on the `container` attribute of the `applicationInstance`.


rwjblue:
In general we don't change past versions.  I'm a little confused how this branch got messed up though (since at the time of branching these guides were not changed yet).

rwjblue:
Confirmed with @locks about this.

PR #952, opened 2015-11-07
[Redesign] Cleanup unnecessary styles

josemarluedke:


josemarluedke:
I think there is more styles to be removed, but I think this is a good start.

locks:
wait, solarized was actually in the styles? xD

PR #953, opened 2015-11-07
[Redesign] Remove unused images

josemarluedke:


josemarluedke:
The tests failed because some styles are using some images, however, these styles are being removed at #952.

PR #956, opened 2015-11-08
Fix code block on classes and instances object model

josemarluedke:
## Before

screen shot 2015-11-07 at 5 46 12 pm

## After

screen shot 2015-11-07 at 5 46 23 pm


michaelrkn:
Thanks @josemarluedke!

PR #650, opened 2015-08-23
[WIP] Style redesign

locks:
- [x] Port over [thoughtbot/ember-guides/](https://github.com/thoughtbot/ember-guides/)
- [x] Tweak contrast
- [x] Tweak structure
- [x] Tweak header
- [x] Fix sidebar/content layout
- [x] Keep old code sample styling, but move flag to the path name
- [x] Clean up paragraph and color CSS
- [x] Swap footer back to old footer (for now)
- [ ] Test in browsers
  - [x] Chrome
  - [ ] Firefox
  - [ ] Safari
  - [ ] Shiira
  - [ ] Camino
  - [ ] Mobile

PR #960, opened 2015-11-08
[Redesign] Cleaned up the headers, spacing a bit

acorncom:
Helps with #650

### DIFF
--- a/source/layout.erb
+++ b/source/layout.erb
@@ -31,19 +31,17 @@
       <%= partial "sidebar" %>
 
       <article>
-        <div class="chapter">
-          <h1>
-            <%= page_title %>
-            <a href="https://github.com/emberjs/guides/edit/master/source/<%= current_page.path.gsub('html', 'md') %>"
-               target="_blank" class="edit-page icon-pencil">Edit Page</a>
-          </h1>
-          <hr>
+        <h1>
+          <%= page_title %>
+          <a href="https://github.com/emberjs/guides/edit/master/source/<%= current_page.path.gsub('html', 'md') %>"
+             target="_blank" class="edit-page icon-pencil">Edit Page</a>
+        </h1>
+        <hr>
 
-          <%= yield %>
+        <%= yield %>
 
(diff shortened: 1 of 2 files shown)


source/layout.erb
```diff
@@ -31,19 +31,17 @@
...
-        <div class="chapter">
-          <h1>
-            <%= page_title %>
-            <a href="https://github.com/emberjs/guides/edit/master/source/<%= current_page.path.gsub('html', 'md') %>"
-               target="_blank" class="edit-page icon-pencil">Edit Page</a>
-          </h1>
-          <hr>
+        <h1>
+          <%= page_title %>
+          <a href="https://github.com/emberjs/guides/edit/master/source/<%= current_page.path.gsub('html', 'md') %>"
```
michaelrkn: Why is the pencil inside the H1? That doesn't make semantic sense to me. (I realize it was like that before your change.) Can you try moving it out and see if it would take much CSS readjustment?

source/layout.erb
```diff
@@ -31,19 +31,17 @@
       <%= partial "sidebar" %>
 
       <article>
-        <div class="chapter">
```
josemarluedke: There are some styles that depends on this class. Such as [here](https://github.com/emberjs/guides/blob/redesign/source/stylesheets/base/_lists.scss#L45).

Also, you there are styles that adds the `chapter-mark`, it would be removed and the image as well.

source/layout.erb
acorncom: @michaelrkn Looks like it gets to be a headache to not have it inside of the H1, yes.  Understood on the non-semantic nature.  Could we add an issue to clean this up later (post redesign and tutorial)?  Would probably help with SEO in the long-run ...

source/layout.erb
acorncom: @josemarluedke Other than the `.chapter .li`, I'm not seeing any other reliance on `.chapter`, have I missed anything?

PR #963, opened 2015-11-09
Tweaks example for pushing to the store with pushPayload

### LINKED ISSUE #945: Tweak the pushing records example for pushPayload
Under Pushing Payloads (http://guides.emberjs.com/v2.1.0/models/pushing-records-into-the-store/#toc_pushing-records) the second example is an example of a JSON-API payload.  It might make more sense to demonstrate a non-JSON API payload in the interests of explaining clearly that the store can handle other payloads and serialize them for the user ...

jsangilve:
Fixes #945 pushing non-JSON API record to the store.


michaelrkn:
Thanks @jsangilve!

PR #969, opened 2015-11-11
[Redesign] Tweaking code styling

acorncom:
Helps with #650 

Interested in what you guys have to think about this change @michaelrkn @locks @josemarluedke  I'm not sure if it's better or not.  Makes the code snippets stick out quite a bit more, but is that a good thing?

Before:
screen shot 2015-11-10 at 8 17 58 pm

After:
screen shot 2015-11-10 at 8 18 12 pm

My feelings will not be hurt if we decide to kill this :-)


josemarluedke:
I personally think that without the bold looks better because the it calls to much attention in the paragraph.

locks:
i have to agree with @josemarluedke.

The background and different colour already call enough attention, the bold is a bit too much.
We might want to keep bold for calling attention to something else.

acorncom:
Thanks guys, I'm inclined to agree. Good to get other opinions on it :+1:

PR #970, opened 2015-11-12
Grammar updates and additional clarification

tleskin:


source/getting-started/index.md
```diff
@@ -16,8 +17,8 @@ This tool provides:
 ### Node.js and npm
 
 Ember CLI is built with JavaScript, and expects the [Node.js](https://nodejs.org/)
-runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/). npm is packaged with Node.js, so if your computer has Node.js
-installed you are ready to go.
+runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/), Node Package manager. npm is packaged with Node.js, so if your computer has Node.js
```
michaelrkn: "manager" should be capitalized.

source/getting-started/index.md
```diff
@@ -45,18 +46,23 @@ download and run [this Git installer](http://git-scm.com/download/win).
 
 ### Watchman (optional)
 
-On Mac and Linux, you can improve file watching performance by installing [Watchman](https://facebook.github.io/watchman/docs/install.html).
+On Mac and Linux, you can improve file watching performance and raise events when files change or get
```
michaelrkn: What does "raise events" mean?

michaelrkn:
Thanks for the contribution, @tleskin! I've added a couple line notes.

tleskin:
@michaelrkn No problem, it's my first time contributing. I'll make these changes.

source/getting-started/index.md
```diff
@@ -16,8 +17,8 @@ This tool provides:
 ### Node.js and npm
 
 Ember CLI is built with JavaScript, and expects the [Node.js](https://nodejs.org/)
-runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/). npm is packaged with Node.js, so if your computer has Node.js
-installed you are ready to go.
+runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/), Node Package Manager. npm is packaged with Node.js, so if your computer has Node.js
```
locks: npm doesn't stand for "node package manager"

tleskin:
Thanks for that clarification. I had noticed that many places had it written differently, like "Node Package manager" as opposed to the way I had written it.

ashleygwilliams:
ya'll might write a note about updating npm. npm releases more often than node: https://docs.npmjs.com/getting-started/installing-node

tleskin:
@ashleygwilliams I added the note about updating npm. I also added one other note about using `-g` since last night at our Ember meetup in Denver we had a beginner who asked why they had to install something globally, since they didn't quite understand that. I feel like it could be helpful for people like him.

source/getting-started/index.md
```diff
@@ -37,6 +38,16 @@ installing Homebrew, run `brew install node` to install Node.js.
 Once you've got Node.js installed, re-run `node --version` to verify your
 computer's setup.
 
+If you already have npm and Node.js installed, it's a good idea to make sure you're on the latest verson of npm since it gets updated more frequently than Node.js does. 
+
+To update npm, run:
+
+```bash
+sudo npm install npm -g
```
michaelrkn: Telling people to use sudo generally strikes me as a bad idea without the proper context and explanation, which I don't think we should get into here.

source/getting-started/index.md
```diff
@@ -37,6 +38,16 @@ installing Homebrew, run `brew install node` to install Node.js.
...
+```
+
+For those unfamiliar with npm, the reason why you are using `-g` is to install the tool globally, so you can use it as a command line tool. You will do the same later when installing Ember and Ember CLI. On the other hand, if you wanted to depend on the package from your own module using something like Node's require, then you would want to install it locally without the `-g`.
```
michaelrkn: "globally", "module", "require"... these are all terms that I don't think we can assume our readers are familiar with. I like the intent of this paragraph, but I think it will end up causing more confusing than clarity.

michaelrkn:
I left a couple more comments. Also make sure to squash your commits. Thanks!

source/getting-started/index.md
tleskin: @michaelrkn Okay. What would you recommend? Trying to add more clarification for these terms or just taking this paragraph out?

source/getting-started/index.md
```diff
@@ -45,18 +56,23 @@ download and run [this Git installer](http://git-scm.com/download/win).
...
 ### Watchman (optional)
 
-On Mac and Linux, you can improve file watching performance by installing [Watchman](https://facebook.github.io/watchman/docs/install.html).
+On Mac and Linux, you can improve file watching performance and receive alerts when files change or get
+deleted/added by installing [Watchman](https://facebook.github.io/watchman/docs/install.html).
 
 ## Installation
 
-Install Ember using npm:
+Ember and Ember CLI are installed using npm.
```
locks: might be misleading, since the last Ember deployed to npm is 1.0.0-beta.2 or something like that?

locks:
It should be mentioned somewhere that Ember CLI's binary is actually called `ember`.

locks:
> If the result is something like version: 1.13.0, you're ready to go.

I thought this had already been replaced with something more generic, must've slipped. Can you address it as well?
Something like "If the previous command printed a version to the terminal, you're ready to go".

tleskin:
@locks I made the changes. Let me know if that's what you meant. Thanks!

source/getting-started/index.md
```diff
@@ -65,10 +81,12 @@ npm install -g phantomjs2
 If you run into any problems with this step, consult the [PhantomJS download
 instructions](http://phantomjs.org/download.html).
 
-To verify that your installation was successful, run:
+Once we have everything installed, we'll be able to run commands with Ember CLI's binary, which is `ember`. There are many commands available by running the binary `ember` along with another command. A list of all commands is available by simply running `ember` or `ember help` from the command line.
```
michaelrkn: I don't think `ember` is technically binary... maybe "executable" is a better word?

source/getting-started/index.md
```diff
@@ -16,8 +17,8 @@ This tool provides:
 ### Node.js and npm
 
 Ember CLI is built with JavaScript, and expects the [Node.js](https://nodejs.org/)
-runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/). npm is packaged with Node.js, so if your computer has Node.js
-installed you are ready to go.
+runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/), the package manager for Node.js. npm is packaged with Node.js, so if your computer has Node.js
```
michaelrkn: Please manually break lines that are longer than 120 characters per the [Style Guide](https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md#style-guide).

michaelrkn:
@tleskin Sorry for the silence over here - I didn't realize you had updated this PR since we last commented. I'm looking through and I think most of the content changes ended up getting made in parallel from other PRs, except for: the PhantomJS install instructions, which because they don't work reliably across platforms currently, we removed in favor of linking to the official install page; and the explanation of the `-g` flag for npm, which I think is outside the scope of the Ember Guides. I'm going to close this out for now - I'm sorry for doing so after you put in this work and took the time to update things. I hope you won't get discouraged from continuing to get involved - there are a bunch of things we could use help on at https://github.com/emberjs/guides/issues - and I'll buy you a beer if you're ever in Portland!

PR #801, opened 2015-09-25
Tutorial

michaelrkn:
This is more or less what we're thinking in terms of the outline of the tutorial:

- [x] What we will be building (gives context)
- [x] Core concepts and how they relate to our tutorial app
- [x] Installing ember cli and other dependencies, a brief introduction to bower and npm (for the people coming from jQuery backgrounds)
- [x] Ember CLI commands and generators, generate new app
- [x] Router basics, templates
  - An about page
  - A contact page
  - A homepage
  - link-to
- [x] Using a model
  - Rentals
  - use POJOs
- [x] Ember Data (mirage)
- [x] Simple component with action
  - Hide/show an image
- [ ] Autocomplete Component
- [ ] Google Maps service and component
- [ ] Helpers
  - Format currency
  - Moment for formatting time?
- [ ] CSP (since it will be needed for both initializers features) 
- [ ] Deploying
- [ ] Addons
  - dotenv
  - pagefront
- [ ] Upgrading

Fixes #445.

PR #974, opened 2015-11-12
[WIP] Draft of updates to the Ember Data page

### LINKED ISSUE #877: Firebase and Tutorial
Switch to Mirage, put Firebase and Divshot on an optional "Extra credits" section.

acorncom:
Helps with: #801

These instructions don't lead to a working setup yet:

- [ ] fix `Error while processing route: index Adapter operation failed Error: Adapter operation failed` errors
- [ ] handle JSON-API upgrade issues, probably best to get people using JSON-API by default

Fixes #877.


michaelrkn:
Looking good! Yes, I think that we should get things working with JSON API if at all possible.

michaelrkn:
Subsumed in #445.

PR #981, opened 2015-11-14
Update Factory Lookup section.

rwjblue:
This section is talking about how to get a singleton instance from the container, and not how to grab the factory itself (so that you might later be able to instantiate it yourself).  

Before this change, based on the title alone I would have expected it to discuss `appInstance._lookupFactory` (or `container.lookupFactory`), which is private.

After this change, it seems clearer that we are looking up an instance not the class itself.


michaelrkn:
Thanks @rwjblue!

PR #985, opened 2015-11-15
Life-cycle -> Lifecycle

tomdale:
Sorry for the nit-picky PR, but we've typically called these "lifecycle hooks," not "life-cycle".


locks:
:+1:

PR #986, opened 2015-11-15
Tweaks to component lifecycle guide

tomdale:
- General language/clarity cleanup
- Move from numeric IDs to semantic IDs (this is more friendly to
  refactorings in the future)
- Remove mention of `_super()` - `didInsertElement` should not require
  this on Ember.Component subclasses. Subclasses of other components
  should follow the normal rules of `_super()` invocation

### DIFF
--- a/source/components/the-component-lifecycle.md
+++ b/source/components/the-component-lifecycle.md
@@ -1,62 +1,118 @@
-Part of what makes a component such a useful tool is that it is closely tied to your app's templates and the DOM itself.
+Part of what makes components so useful is that they let you take
+complete control of a section of the DOM.  This allows for direct DOM
+manipulation, listening and responding to browser events, and using 3rd
+party JavaScript libraries in your Ember app.
 
-Components are your primary means for direct DOM manipulation, listening and responding to browser events, and attaching 3rd party JS libraries into your Ember app.
+As components are rendered, re-rendered and finally removed, Ember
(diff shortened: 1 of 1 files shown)


source/components/the-component-lifecycle.md
```diff
@@ -1,62 +1,118 @@
-Part of what makes a component such a useful tool is that it is closely tied to your app's templates and the DOM itself.
+Part of what makes a component such a useful tool is that it allows you
```
michaelrkn: "a component" -> "components" (to match "They" in the following sentence)

source/components/the-component-lifecycle.md
```diff
@@ -1,62 +1,118 @@
...
+  this.$('input.date').myDatePickerLib();
+}
+```
+
+[`didInsertElement()`][did-insert-element] is also a good place to
+attach event listeners. This is particularly useful for custom events or
+other browser events which do not have a [built-in event
+handler][event-names].
+
+For example, Perhaps you have some custom CSS animations trigger when the component
```
michaelrkn: lowercase Perhaps

michaelrkn:
Thanks for the improvements, @tomdale! I pointed out just a couple small grammar things I noticed.

tomdale:
@michaelrkn Updated to incorporate your feedback

PR #989, opened 2015-11-17
Added `type` to acceptable input attributes

Glennvd:
Type is obviously accepted as an attribute to the {{input}} helper, since it's used in an example below. I feel like it should be included in the list of accepted attributes, for sake of completeness.


rwjblue:
:+1:

PR #994, opened 2015-11-19
Remove PhantomJS install instructions

### LINKED ISSUE #992: npm install -g phantomjs2 is stuck on OSX 10.11
The [Installing Ember page](http://guides.emberjs.com/v2.2.0/getting-started/) asks the users to run `npm install -g phantomjs2`. This runs forever on OSX 10.11. After some googling around, I saw that phantomjs 2 doesn't build on 10.11 yet, and the effort of getting it to work is stalled.

However, `npm install -g phantomjs` works. Is phantomjs 1.x sufficient for ember? If so, will you please consider changing the guides? Getting stuck this early in the guide is a pretty bad experience for people who want to give ember a try.

locks:
Fixes #992.

It's more trouble than it's worth.

### DIFF
--- a/source/getting-started/index.md
+++ b/source/getting-started/index.md
@@ -47,6 +47,11 @@ download and run [this Git installer](http://git-scm.com/download/win).
 
 On Mac and Linux, you can improve file watching performance by installing [Watchman](https://facebook.github.io/watchman/docs/install.html).
 
+### PhantomJS (optional)
+
+You can run your tests from the command line with PhantomJS, without the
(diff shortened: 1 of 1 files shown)


rwjblue:
We should likely still mention PhantomJS, but I agree having the exact install steps is a path to pain.  Maybe we can just point them to `phantomjs.com` and call it a day?

locks:
@rwjblue something like this?

rwjblue:
:+1:

michaelrkn:
There's something weird with your PR - it pulls in "Fix quotes in customizing element example", which already got merged.

PR #996, opened 2015-11-20
Removing need/have extra word

trcarden:
There seems to be an extra word in the documentation


michaelrkn:
Good catch! Can you actually remove "need" instead of "have", so that it flows better: "If you _have_ more extensive "data sharing" needs..."? If easier, feel free to open another PR and close this one so you don't have to rebase.

trcarden:
Closing in favor of https://github.com/emberjs/guides/pull/999

PR #999, opened 2015-11-21
Removing need/have extra word

trcarden:
Removing need and leaving have per @michaelrkn recommendation


michaelrkn:
Thanks @trcarden!

Issue #858, opened 2015-10-06
Document all available config options for .ember-cli


HeroicEric:
There really isn't any place to see all of the options that can be specified in `.ember-cli` and I've seen people ask about it a lot.

I would make  PR instead of just this issue but I must first figure out what they are :smile: This is a reminder.

- `disableAnalytics`
- `usePods`
- `proxy`

If anybody knows any that are missing off the top of their head, please comment here.

locks:
I personally don't want to document pod usage until the final structure is here.
The other PRs are welcome ;)

michaelrkn:
@HeroicEric this ties in with #823 - do you think you could look into the issues brought up there while you're at it?

josenberg:
From [Ember-cli Docs](http://www.ember-cli.com/#runtime-configuration).

```
"skipGit" : true,
"port" : 999,
"host" : "0.1.0.1",
"liveReload" : true,
"environment" : "mock-development",
"checkForUpdates" : false
```

Lxrd-AJ:
In addition to what @josenberg wrote
```
"output-path": "dist/"
```

Lxrd-AJ:
Can we also include config options for `config/environment.js` @michaelrkn

michaelrkn:
@Lxrd-AJ I think we should have one page for Ember CLI, and another page (or pages) for other configuration options. This issue is in particular for Ember CLI, but I've also opened #823 for documenting configuration more generally.

locks:
@HeroicEric @josenberg @Lxrd-AJ could one of you send a PR so it can get reviewed and merged? thanks!

PR #1000, opened 2015-11-22
Ember build directory config

Lxrd-AJ:
Added command to allow the user instruct ember-cli for default build directory for compiled EmberJS web app

### DIFF
--- a/source/getting-started/first-app.md
+++ b/source/getting-started/first-app.md
@@ -28,8 +28,15 @@ server. To build your app for production, run:
 ember build --environment=production
 ```
 
-When this finishes the `dist/` directory will contain a version of your
+When this finishes the `dist/` directory (default) will contain a version of your
 application suitable for production use. Copy the contents of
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thanks for your suggestion, @Lxrd-AJ. Note from the [Contributing guidelines](https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md) that "the Guides are primarily meant to cover the "Ember happy path", and are not intended to be comprehensive." Since the "happy path" here is the default output directory, I don't think it's appropriate to include the configuration option - that's a better fit for a future guide on configuration options (which I see you've [also updated](, and for updating https://github.com/emberjs/guides/issues/858#issuecomment-158781828) - thanks!). If you'd like to help out with #858, that would be awesome!

Lxrd-AJ:
Many thanks @michaelrkn , would try my best to contribute to #858 😅

PR #1001, opened 2015-11-23
Revive glossary

tomdale:
As discussed in one of the previous meetings, this revives the glossary as a way to familiarize those totally new to the frontend space with terms they may encounter when learning Ember that aren't Ember specific.

+@wycats @michaelrkn

### DIFF
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -12,6 +12,8 @@
       url: "index"
     - title: "Your First Ember App"
       url: "first-app"
+    - title: "Glossary"
+      url: "glossary"
     - title: "Core Concepts"
       url: "core-concepts"
 
--- a/source/getting-started/glossary.md
+++ b/source/getting-started/glossary.md
@@ -1,0 +1,85 @@
+Joining a web development community can be a challenge within itself, especially when all the resources you visit assume you're familiar with other technologies that you're not familiar with.
+
+Our goal is to help you avoid that mess and come up to speed as fast as possible; you can consider us your internet friend.
+
+## CDN
+Content Delivery Network
+
+This is typically a paid service you can use to get great performance for your app. Many CDNs act as caching proxies to your origin server; some require you to upload your assets to them. They give you a URL for each resource in your app. This URL will resolve differently for folks depending on where they're browsing.
+
+Behind the scenes, the CDN will distribute your content geographically with the goal of end-users being able to fetch your content with the lowest latency possible. For example, if a user is in India, they'd likely get content served from India faster than from the United States.
+
+
+## CoffeeScript, TypeScript
+These are both languages that compile to JavaScript. You're able to write your code using the syntax they offer and when ready you compile your TypeScript or CoffeeScript into JavaScript.
+
+[CoffeeScript vs TypeScript](http://www.stoutsystems.com/articles/coffeescript-versus-typescript/)
+
+
+## Evergreen browsers
+Browsers that update themselves (without user intervention).
+
+[Evergreen Browsers](http://tomdale.net/2013/05/evergreen-browsers/)
+
+
+## ES3, ES5, ES5.1, ES6 (aka ES2015), etc
+ES stands for ECMAScript, which is the specification that JavaScript is based on. The number that follows is the version of the specification.
+
+Most browsers support at least ES5, and some even have ES6 (also known as ES2015) support. You can check each browser's support (including yours) here:
+
+* [ES5 support](http://kangax.github.io/compat-table/es5/)
+* [ES6 support](http://kangax.github.io/compat-table/es6/)
+
+[ECMAScript](https://en.wikipedia.org/wiki/ECMAScript)
+
+
+## LESS, Sass
+Both LESS and Sass are types of CSS preprocessor markup intended to give you much more control over your CSS. During the build process, the LESS or Sass resources compile down to vanilla CSS (which can be executed in a browser).
+
+[Sass/Less Comparison](https://gist.github.com/chriseppstein/674726)
+
+
+## Linter, linting, jslint, jshint
+A validation tool which checks for common issues in your JavaScript. You'd usually use this in your build process to enforce quality in your codebase. A great example of something to check for: *making sure you've always got your semicolons*.
+
+[An example of some of the options you can configure](http://jshint.com/docs/options/)
+
+
+## Polyfill
+This is a concept that typically means providing JavaScript which tests for features that are missing (prototypes not defined, etc) and "fills" them by providing an implementation.
+
+
+## Promise
+Asynchronous calls typically return a promise (or deferred). This is an object which has a state: it can be given handlers for when it's fulfilled or rejected.
+
+Ember.js makes use of these in places like the model hook for a route. Until the promise resolves, Ember is able to put the route into a "loading" state.
+
+* [An open standard for sound, interoperable JavaScript promises](https://promisesaplus.com/)
+* [emberjs.com - A word on promises](http://emberjs.com/guides/routing/asynchronous-routing/#toc_a-word-on-promises)
+
+
+## SSR
+Server Side Rendering
+
+[Inside FastBoot: The Road to Server-Side Rendering](http://emberjs.com/blog/2014/12/22/inside-fastboot-the-road-to-server-side-rendering.html)
+
+
+## Transpile
+When related to JavaScript, this can be part of your build process which "transpiles" (converts) your ES6 syntax JavaScript to JavaScript that is supported by current browsers.
+
+Besides ES6, you'll see a lot of content about compiling/transpiling CoffeeScript, a short-hand language which can "compile" to JavaScript.
+
+* Ember CLI specifically uses [Babel](https://babeljs.io/) via the [ember-cli-babel](https://github.com/babel/ember-cli-babel) plugin.
+
+
+## Shadow DOM
+Not to be confused with Virtual DOM. Shadow DOM is still a work in progress, but basically a proposed way to have an "isolated" DOM encapsulated within your app's DOM.
+
+Creating a re-usable "widget" or control might be a good use-case for this. Browsers implement some of their controls using their own version of a shadow DOM.
+
+* [W3C Working Draft](http://www.w3.org/TR/shadow-dom/)
+* [What the Heck is Shadow DOM?](http://glazkov.com/2011/01/14/what-the-heck-is-shadow-dom/)
+
+
+## Virtual DOM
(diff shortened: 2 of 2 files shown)


michaelrkn:
cc @locks

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
+Joining a web development community can be a challenge within itself, especially when all the resources you visit assume you're familiar with other technologies that you're not familiar with.
+
+Our goal is to help you avoid that mess and come up to speed as fast as possible; you can consider us your internet friend.
+
+## CDN
+Content Delivery Network
+
+This is typically a paid service you can use to get great performance for your app. Many CDNs act as caching proxies to your origin server; some require you to upload your assets to them. They give you a URL for each resource in your app. This URL will resolve differently for folks depending on where they're browsing.
```
michaelrkn: "caching proxy" is not a great term if we're trying to educate people who don't have a lot of context. "They give you a URL for each resource [which] will resolve differently" - I don't think this really explains what CDNs do very well.

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+Behind the scenes, the CDN will distribute your content geographically with the goal of end-users being able to fetch your content with the lowest latency possible. For example, if a user is in India, they'd likely get content served from India faster than from the United States.
+
+
+## CoffeeScript, TypeScript
+These are both languages that compile to JavaScript. You're able to write your code using the syntax they offer and when ready you compile your TypeScript or CoffeeScript into JavaScript.
```
michaelrkn: Why would you use CoffeeScript or TypeScript?

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+[CoffeeScript vs TypeScript](http://www.stoutsystems.com/articles/coffeescript-versus-typescript/)
```
michaelrkn: Why are we linking to a comparison instead of an explanation of what each one is?

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+Most browsers support at least ES5, and some even have ES6 (also known as ES2015) support. You can check each browser's support (including yours) here:
+
+* [ES5 support](http://kangax.github.io/compat-table/es5/)
+* [ES6 support](http://kangax.github.io/compat-table/es6/)
+
+[ECMAScript](https://en.wikipedia.org/wiki/ECMAScript)
+
+
+## LESS, Sass
+Both LESS and Sass are types of CSS preprocessor markup intended to give you much more control over your CSS. During the build process, the LESS or Sass resources compile down to vanilla CSS (which can be executed in a browser).
```
michaelrkn: CSS isn't really executed, is it?

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Linter, linting, jslint, jshint
+A validation tool which checks for common issues in your JavaScript. You'd usually use this in your build process to enforce quality in your codebase. A great example of something to check for: *making sure you've always got your semicolons*.
+
+[An example of some of the options you can configure](http://jshint.com/docs/options/)
+
+
+## Polyfill
+This is a concept that typically means providing JavaScript which tests for features that are missing (prototypes not defined, etc) and "fills" them by providing an implementation.
```
michaelrkn: Link to another resource with more info

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Promise
+Asynchronous calls typically return a promise (or deferred). This is an object which has a state: it can be given handlers for when it's fulfilled or rejected.
```
michaelrkn: "handlers", "fulfilled", "rejected" - these terms aren't going to mean much to someone who isn't already familiar with promises.

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+Ember.js makes use of these in places like the model hook for a route. Until the promise resolves, Ember is able to put the route into a "loading" state.
+
+* [An open standard for sound, interoperable JavaScript promises](https://promisesaplus.com/)
+* [emberjs.com - A word on promises](http://emberjs.com/guides/routing/asynchronous-routing/#toc_a-word-on-promises)
+
+
+## SSR
+Server Side Rendering
+
+[Inside FastBoot: The Road to Server-Side Rendering](http://emberjs.com/blog/2014/12/22/inside-fastboot-the-road-to-server-side-rendering.html)
```
michaelrkn: All of the other entries have short explanations; this probably should too, rather than just linking away.

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Transpile
+When related to JavaScript, this can be part of your build process which "transpiles" (converts) your ES6 syntax JavaScript to JavaScript that is supported by current browsers.
+
+Besides ES6, you'll see a lot of content about compiling/transpiling CoffeeScript, a short-hand language which can "compile" to JavaScript.
+
+* Ember CLI specifically uses [Babel](https://babeljs.io/) via the [ember-cli-babel](https://github.com/babel/ember-cli-babel) plugin.
```
michaelrkn: Why is this a list item?

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## SSR
+Server Side Rendering
+
+[Inside FastBoot: The Road to Server-Side Rendering](http://emberjs.com/blog/2014/12/22/inside-fastboot-the-road-to-server-side-rendering.html)
+
+
+## Transpile
+When related to JavaScript, this can be part of your build process which "transpiles" (converts) your ES6 syntax JavaScript to JavaScript that is supported by current browsers.
```
michaelrkn: We shouldn't use the word "transpile" in the definition of "transpile"... that's a bit circular.

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+Besides ES6, you'll see a lot of content about compiling/transpiling CoffeeScript, a short-hand language which can "compile" to JavaScript.
```
michaelrkn: And Typescript - see above.

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+* Ember CLI specifically uses [Babel](https://babeljs.io/) via the [ember-cli-babel](https://github.com/babel/ember-cli-babel) plugin.
+
+
+## Shadow DOM
+Not to be confused with Virtual DOM. Shadow DOM is still a work in progress, but basically a proposed way to have an "isolated" DOM encapsulated within your app's DOM.
```
michaelrkn: "an "isolated" DOM encapsulated within your app's DOM" - this is not a good explanation for someone unfamiliar with web development.

source/getting-started/glossary.md
```diff
@@ -0,0 +1,85 @@
...
+
+Creating a re-usable "widget" or control might be a good use-case for this. Browsers implement some of their controls using their own version of a shadow DOM.
+
+* [W3C Working Draft](http://www.w3.org/TR/shadow-dom/)
+* [What the Heck is Shadow DOM?](http://glazkov.com/2011/01/14/what-the-heck-is-shadow-dom/)
+
+
+## Virtual DOM
+Not to be confused with Shadow DOM. The concept of a virtual DOM means abstracting your code (or in our case, Ember) away from using the browser's DOM in favor of a "virtual" DOM that can easily be accessed for read/writes or even serialized.
```
michaelrkn: Why would it be serialized?

michaelrkn:
As I expressed when this was discussed, I'm not a fan of this, but I'm okay to agree to disagree. That said, I think the quality of the content in its current form is not great, and I've left several comments on the parts I think are particularly bad. I think we need to get this in much better shape before merging.

wycats:
@michaelrkn I'd rather have something than nothing here that we can iterate on. 

Your ideological objection to the section is causing you to require perfection for v1, which is never necessary.

michaelrkn:
@wycats My ideological objection is definitely a part of the problem, but the other part of the problem is that the content is legitimately low-quality. I think that an open PR is a great place to start iterating and improving the quality. If @locks thinks it's high enough quality to merge, I'll again agree to disagree.

tomdale:
The copy can definitely use some improvement. I will incorporate @michaelrkn's feedback to get the worst of it polished off. I don't want to block forever awaiting perfection, but there are parts that do dip below what I'd consider the acceptable quality threshold that shouldn't be too hard to fix.

michaelrkn:
@tomdale :+1:

data/pages.yml
```diff
@@ -12,6 +12,8 @@
       url: "index"
     - title: "Your First Ember App"
       url: "first-app"
+    - title: "Glossary"
+      url: "glossary"
     - title: "Core Concepts"
       url: "core-concepts"
```
acorncom: I really like the idea of a Glossary, but wonder if putting it here in the Getting Started section is going to stop people cold (especially when we rearrange things when we merge in the tutorial).  Could we find a different spot to park the Glossary link and instead link to the appropriate section as needed when we first introduce a new concept in the tutorial / guides?

Seems like that would allow newcomers to catch up where needed without them needing to read the entire page as part of an intro to Ember ...

acorncom:
Per the docs meeting yesterday, we'll merge this in post-tutorial and find a slightly different home for it.

data/pages.yml
rtablada: IMO just like a traditional technical book, put the Glossary at the end?

I've found that most of the guides put a sprinkle of what different terms are.
I think the glossary is more of a reference back rather than an introduction?

locks: Another cool idea is if it contained terms actually used throughout the guides.

rtablada: I really like that thought, then as terms are introduced, they could give a quick overview and link to the fully glossary?

pixelhandler:
How about adding more info for new comers e.g. like this http://kylecoberly.github.io/ember-vocab/

data/pages.yml
acorncom: Yep, :+1: on linking to the entry in the glossary

locks:
@tomdale please to rebase and move glossary to the end of the TOC

acorncom:
Per a conversation we had a few weeks ago at the docs meeting, Tom may not have much time to do this soon, so if someone else wants to do this, I think it's up for grabs ( @tomdale, please correct me if I'm wrong )

PR #1003, opened 2015-11-24
Make example with loading substates for nested routes more consistent

hlubek:
The route names that are used in the example should be consistent.


michaelrkn:
Thanks @hlubek!

PR #1005, opened 2015-11-27
Update NodeJS installation info.

daniellutz:
I believe that `which` is a better way to find if a software is installed rather than trying to run it with `--version`.

And running `node --version` to re-run it again looked a bit confusing.

### DIFF
--- a/source/getting-started/index.md
+++ b/source/getting-started/index.md
@@ -20,22 +20,22 @@ runtime. It also requires dependencies fetched via [npm](https://www.npmjs.com/)
 installed you are ready to go.
 
 Ember requires Node.js 0.12 or higher and npm 2.7 or higher.
-If you're not sure whether you have Node.js, run this on your
+If you're not sure whether you have Node.js or the right version, run this on your
 command line:
 
 ```bash
 node --version
+npm --version
 ```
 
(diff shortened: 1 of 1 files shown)


daniellutz:
Oh, I forgot that `which` is not available to Windows users.

I will submit an alternative for them.

michaelrkn:
Thanks for pointing this out! How about something like:


Ember requires Node.js 0.12 or higher and npm 2.7 or higher.
If you're not sure whether you have Node.js or the right version, run this on your
command line:

```bash
node --version
npm --version
```

If you get a "command not found" error or an outdated version for Node:

* Windows or Mac users can download and run [this Node.js installer](http://nodejs.org/download/).
* Mac users often prefer to install Node using [Homebrew](http://brew.sh/). After
installing Homebrew, run `brew install node` to install Node.js.
* Linux users can use [this guide for Node.js installation on Linux](https://github.com/joyent/node/wiki/Installing-Node.js-via-package-manager).

If you get an out of date version of npm, run `npm install -g npm`.

daniellutz:
Yeah ! Amazing !

I was trying to think something like that but I needed to go to work, hahaha :smile: 

Now it's awesome !

michaelrkn:
Can you just squash your commits now? Thanks!!

PR #1006, opened 2015-11-28
Remove resetNamespace from defining routes guide

alexspeller:
So the first paragraph here is just plain wrong - you can't reuse routes in multiple route trees because routes have to be uniquely named.

But I think that there is only one edge case where resetNamespace actually makes sense, and even then it's only out of laziness: Moving a route under a new parent route without having to update link-to's and transitionTos. But (a) that only covers one case of moving a route, if you want to move a route out from under a parent it doesn't help, and (b) it's actually a bit of a hack because it removes the conventional mapping of route nesting &rarr; file location, which I suspect will become even more annoying as pods mature.

So I propose removing this section from the guide entirely rather than correcting it as the use case for it is quite narrow and it's just confusing and not really necessary.

### DIFF
--- a/source/routing/defining-your-routes.md
+++ b/source/routing/defining-your-routes.md
@@ -196,30 +196,6 @@ Router.map(function() {
 });
 ```
 
-## Resetting Nested Route Namespace
-
-When nesting routes, it may be beneficial for a child route to not inherit its ancestors name. This allows you to reference and reuse a given route in multiple route trees as well as keep the class name short.
-
-You can reset the current "namespace" with the aptly named `resetNamespace: true` option.
-
-```app/router.js
-Router.map(function() {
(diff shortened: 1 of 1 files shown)


danmcclain:
The place that I actually wanted the intended behavior has 2 sub routes that used alllll the same logic with a different parent, it would have been a perfect use case, but due to embers/ember.js#11692, could not use the feature. Since it is so broken, I agree that the section should be removed

michaelrkn:
Thanks!

locks:
@MarkMT `resetNamespace` is still around, the example in which it was used was not technically correct so it was removed. As for documentation, https://github.com/emberjs/ember.js/blob/master/packages/ember-routing/lib/system/router.js#L1000-L1009 is what you need to know.

MarkMT:
@locks thanks.

Issue #944, opened 2015-11-05
Update syntax highlighting


michaelrkn:
We currently use a fork of [CodeRay](https://github.com/rubychan/coderay) with support for Handlebars. CodeRay isn't well-maintained and doesn't support ES6. The Ember homepage uses [highlight.js](https://highlightjs.org/) client-side. We should probably switch to highlight.js.

If somebody can investigate this and make sure it's a good idea, and then implement it, that would be awesome!

---

Edit (**@acorncom**): we've got more data on what we're after and we're after a different option now.

We'd like to implement [Prism.js](http://prismjs.com) instead of Highlight.js as discussed below, as it supports [line highlighting](http://prismjs.com/plugins/line-highlight/) and [line numbers](http://prismjs.com/plugins/line-numbers/) as well as ES6, `.scss` and `.hbs` files

Here's what I think we need to do to get this working:
- [ ] Re-implement `_highlight` to produce HTML in a format that Prism.js supports.  [Our current implementation](https://github.com/emberjs/guides/blob/master/lib/highlighter.rb#L15-L30) uses CodeRay but we don't actually need to produce HTML using Coderay at all
- [ ] Setup the use of prism.js in a new JS file that handles things well.  We may want to make this globally available instead of versioned for each version of the guides (let's get this built for one version and then talk about this on a PR ...)
- [ ] Drop the use of the coderay gem (once we're sure we're ready to trash it).  We can't tell dgeb he can delete his fork yet, as `emberjs/website` still depends on it for API doc creation as well

acorncom:
And we may want this to be a universal JS file as mentioned on #934

michaelrkn:
I'm not sure about that. This isn't something we want to backport.

dustyburwell:
I'm not sure how valuable the line numbers in the code formatting are, or if they're ever referenced throughout the rest of the documentation, but I came across this: https://highlightjs.readthedocs.org/en/latest/line-numbers.html

If the line numbers are important to anyone, then highlight.js is clearly not a good choice.

dustyburwell:
@michaelrkn can you elaborate on what features are important/necessary in the code highlighting? From the description above, it sounds like HBS and ES6 support are a must. What else should be considered requirement?

acorncom:
[...]

PR #1007, opened 2015-11-28
Quick Start Guide

tomdale:
After a few discussions with @wycats, we realized that the fact that there's no "try Ember in 5 minutes" resource was extremely harmful to adoption. This guide is my attempt to rectify that. It significantly expands on (and replaces) the old "Your First Ember App" section.

Most other libraries and frameworks have something like this. You can compare it, for example, to [Angular 2's quickstart](https://angular.io/docs/js/latest/quickstart.html). One nice thing that Ember CLI does for us is eliminate almost all of the work happening in Angular's quickstart, so our quickstart can actually begin covering the conceptual model rather than just have the user type boilerplate code.

I'd love to hear feedback, particularly from beginners. And I want to make sure this integrates well with the tutorial work @michaelrkn has been doing.

### DIFF
--- a/source/getting-started/quick-start.md
+++ b/source/getting-started/quick-start.md
@@ -1,0 +1,262 @@
+This guide will teach you how to build a simple app using Ember from
+scratch.
+
+We'll cover these steps:
+
+1. Installing Ember
+2. Creating a new application
+3. Defining a route
+4. Writing a UI component
+5. Building your app to be deployed to production
+
+## Install Ember
+
+You can install Ember with a single command using npm, the Node.js package
+manager. Type this into your terminal:
+
+```sh
+npm install ember-cli -g
+```
+
+Don't have npm? [Learn how to install Node.js and npm here][npm].
+
+[npm]: https://docs.npmjs.com/getting-started/installing-node
+
+## Create a New Application
+
+Once you've installed Ember via npm, you will have access to a new
+`ember` command in your terminal. You can use the `ember new` command to
+create a new application.
+
+```sh
+ember new ember-quickstart
+```
+
+This one command will create a new directory called `ember-quickstart`
+and set up a new Ember application inside of it.  Out of the box, your
+application will include:
+
+* A development server
+* Template compilation
+* CSS preprocessing
+* JavaScript and CSS minification
+* ES2015 features via Babel
+
+By providing everything you need to build production-ready web
+applications in an integrated package, Ember makes starting new projects
+a breeze.
+
+Let's make sure everything is working properly. Open the
+`ember-quickstart` directory in your favorite editor, such as Sublime
+Text or Vim. Once you have your editor open, `cd` into the application
+directory and start the development server by typing:
+
+```sh
+cd ember-quickstart
+ember serve
+```
+
+After a few seconds, you should see output that looks like this:
+
+```text
+Livereload server on http://localhost:49152
+Serving on http://localhost:4200/
+```
+
+(To stop the server at any time, type Ctrl-C in your terminal.)
+
+Open [http://localhost:4200/](http://localhost:4200) in your browser of
+choice. You should see a page that says "Welcome to Ember" and not much
+else. Congratulations! You just created and booted your first Ember app.
+
+Switch to your editor and open `app/templates/application.hbs`. This is
+called the `application` template and it is always on screen while the
+user has your application loaded.
+
+In your editor, change the text inside the `<h1>` from `Welcome to
+Ember` to `PeopleTracker` and save the file. Notice that Ember detects
+the change you just made and automatically reloads the page for you in
+the background. You should see that "Welcome to Ember" has been
+replaced by "PeopleTracker".
+
+## Define a Route
+
+Let's build an application that shows a list of scientists. To do that,
+the first step is to create a route. For now, you can think of routes as
+being the different pages that make up your application.
+
+Ember comes with _generators_ that automate the boilerplate code for
+common tasks. To generate a route, type this in your terminal:
+
+```sh
+ember generate route scientists
+```
+
+You'll see output like this:
+
+```text
+installing route
+  create app/routes/scientists.js
+  create app/templates/scientists.hbs
+updating router
+  add route scientists
+installing route-test
+  create tests/unit/routes/scientists-test.js
+```
+
+That's Ember telling you that it has created:
+
+1. A template to be displayed when the user visits `/scientists`
+2. A `Route` object that fetches the model used by that template
+3. An entry in the application's router (located in `app/router.js`)
+4. A unit test for this route
+
+Open the newly-created template in `app/templates/scientists.hbs` and add
+the following HTML:
+
+```app/templates/scientists.hbs
+<h2>List of Scientists</h2>
+```
+
+In your browser, open
+[http://localhost:4200/scientists](http://localhost:4200/scientists). You should
+see the `<h2>` you put in the `scientists.hbs` template, right below the
+`<h1>` from our `application.hbs` template.
+
+Now that we've got the `scientists` template rendering, let's give it some
+data to render. We do that by specifying a _model_ for that route, and
+we can specify a model by editing `app/routes/scientists.js`.
+
+We'll take the code created for us by the generator and add a `model()`
+method to the `Route`:
+
+```app/routes/scientists.js{+4,+5,+6}
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+  model() {
+    return ['Marie Curie', 'Mae Jemison', 'Albert Hofmann'];
+  }
+});
+```
+
+(This code example uses the latest features in JavaScript, some of which
+you may not be familiar with. Learn more with this [overview of the
+newest JavaScript features][es6-bullet-points].)
+
+[es6-bullet-points]: https://ponyfoo.com/articles/es6
+
+In a route's `model()` method, you return whatever data you want to make
+available to the template. If you need to fetch data asynchronously, the
+`model()` method supports any library that uses [JavaScript
+Promises][promises].
+
+[promises]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise
+
+Now let's tell Ember how to turn that array of strings into HTML. Open
+the `scientists` template and add some Handlebars code to loop through the
+array and print it:
+
+```app/templates/scientists.hbs{+3,+4,+5,+6,+7}
+<h2>List of Scientists</h2>
+
+<ul>
+  {{#each model as |scientist|}}
+    <li>{{scientist}}</li>
+  {{/each}}
+<ul>
+```
+
+Here, we use the `each` helper to loop over each item in the array we
+provided from the `model()` hook and print it inside an `<li>` element.
+
+## Create a UI Component
+
+As your application grows and you notice you are sharing UI elements
+between multiple pages (or using them multiple times on the same page),
+Ember makes it easy to refactor your templates into reusable components.
+
+Let's create a `people-list` component that we can use
+in multiple places to show a list of people.
+
+As usual, there's a generator that makes this easy for us. Make a new
+component by typing:
+
+```sh
+ember generate component people-list
+```
+
+Copy and paste the `scientists` template into the `people-list`
+component's template and edit it to look as follows:
+
+```app/templates/components/people-list.hbs
+<h2>{{title}}</h2>
+
+<ul>
+  {{#each people as |person|}}
+    <li>{{person}}</li>
+  {{/each}}
+<ul>
+```
+
+Note that we've changed the title from a hard-coded string ("List of
+Scientists") to a dynamic property (`{{title}}`). We've also renamed
+`scientist` to the more-generic `person`, decreasing the coupling of our
+component to where it's used.
+
+Save this template and switch back to the `scientists` template. Replace all
+our old code with our new componentized version. Components look like
+HTML tags but instead of using angle brackets (`<tag>`) they use double
+curly braces (`{{component}}`). We're going to tell our component:
+
+1. What title to use, via the `title` attribute.
+2. What array of people to use, via the `people` attribute. We'll
+   provide this route's `model` as the list of people.
+
+```app/templates/people.hbs{-1,-2,-3,-4,-5,-6,-7,+8}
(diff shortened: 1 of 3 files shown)


source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+all.
+
+## Building For Production
+
+Now that we've written our application and verified that it works in
+development, it's time to get it ready to deploy to our users. To do so,
+run the following command:
+
+```sh
+ember build
```
stefanpenner: should we introduce --env production ?

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+updating router
+  add route scientists
+installing route-test
+  create tests/unit/routes/scientists-test.js
+```
+
+That's Ember telling you that it has created:
+
+1. A template to be displayed when the user visits `/scientists`
+2. A `Route` object that fetches the model used by that template
```
kimroen: Simplify:

> A `Route` object that prepares the data for that template

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+
+Open [http://localhost:4200/](http://localhost:4200) in your browser of
+choice. You should see a page that says "Welcome to Ember" and not much
+else. Congratulations! You just created and booted your first Ember app.
+
+Switch to your editor and open `app/templates/application.hbs`. This is
+called the `application` template and it is always on screen while the
+user has your application loaded.
+
+In your editor, change the text inside the &lt;h1&gt; from `Welcome to
```
kimroen: Backtick the `<h1>`.

source/getting-started/quickstart.md
kimroen: The next paragraph does say that it packages everything into an optimized bundle, so it should really include that (or the friendlier `-prod`)

steveklabnik:
hi I haven't read this yet but :+1: is it ever needed, lemme check it out now

steveklabnik:
Looks great here!

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+After a few seconds, you should see output that looks like this:
+
+```text
+Livereload server on http://localhost:49152
+Serving on http://localhost:4200/
+```
+
+(To stop the server at any time, type Ctrl-C in your terminal.)
```
maclover7: Should this be included in the `text` block?

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+You can install Ember with a single command using npm, the Node.js package
+manager. Type this into your terminal:
+
+```sh
+npm i ember-cli -g
+```
+
+Don't have npm? [Learn how to install Node.js and npm here][npm].
+
+[npm]: https://docs.npmjs.com/getting-started/installing-node
```
trek: Have they fixed these instructions so that they actually work? I remember not long ago the recommended way (at least on MacOS) caused all sorts of problems.

source/getting-started/quickstart.md
trek: I think this is just part of Tom's narration.

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+called the `application` template and it is always on screen while the
+user has your application loaded.
+
+In your editor, change the text inside the &lt;h1&gt; from `Welcome to
+Ember` to `PeopleTracker` and save the file. Notice that Ember detects
+the change you just made and automatically reloads the page for you in
+the background. You should see that "Welcome to Ember" has been
+replaced by "PeopleTracker".
+
+## Define a Route
```
trek: The last time we all talked about quick start the idea of introducing routes before components was shot down. Did something change?

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+4. Writing a UI component
+5. Building your app to be deployed to production
+
+## Install Ember
+
+You can install Ember with a single command using npm, the Node.js package
+manager. Type this into your terminal:
+
+```sh
+npm i ember-cli -g
```
trek: We've tended to prefer full command names over the short versions, for clarity to newer people.

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+`<h1>` from our `application.hbs` template.
+
+Now that we've got the `scientists` template rendering, let's give it some
+data to render. We do that by specifying a _model_ for that route, and
+we can specify a model by editing `app/routes/scientists.js`.
+
+We'll take the code created for us by the generator and add a `model()`
+method to the `Route`:
+
+```js{+4,+5,+6}
```
trek: you can also specify the file name here too, with our markdown parser.

source/getting-started/quickstart.md
```diff
@@ -0,0 +1,262 @@
...
+
+1. A template to be displayed when the user visits `/scientists`
+2. A `Route` object that fetches the model used by that template
+3. An entry in the application's router (located in `app/router.js`)
+4. A unit test for this route
+
+Open the newly-created template in `app/templates/scientists.hbs` and add
+the following HTML:
+
+```hbs
```
trek: you can also specify the file name here too, with our markdown parser.

source/getting-started/quickstart.md
trek: Why use a route at all? To someone who doesn't know Ember, the use of a route here doesn't add anything over defining that data property inside a component.

michaelrkn:
Awesome! This is a great idea.

What do you think about integrating this with the tutorial, so that a new Ember dev following along with the Guides starting at the beginning can use this page to create their app, and then the tutorial continues to build it out where the quickstart leaves off?

Right now, the structure of the Getting Started and Tutorial sections looks like:

- Getting Started
    - Installing Ember
    - Core Concepts
    - Ember CLI (ember new, directory structure, ember s)
- Tutorial
    - Routes and Templates
    - The Model Hook
    - Ember Data
    - A Simple Component

Maybe we could change it to something like:

- Getting Started
    - Quick Start (installing, ember new, first template, component, model hook, building)
    - Core Concepts
    - Tooling (directory structure, etc)
- Tutorial
    - Routes and Templates
    - Ember Data
    - A Simple Component

If this or some variation makes sense, can you change this PR to be against the `tutorial` branch?

Thanks!!

source/getting-started/quickstart.md
tomdale: @kimroen I specifically wanted to use the word "model" to strengthen the mental model that "`model()` hook means template model".

@trek It's key for the refactoring step from route template to component template. I agree that this should be replaced by a routable component once those land.

source/getting-started/quickstart.md
tomdale: Good call, I forgot backticks were escaped when I started writing this but you can see I remembered about halfway through. :P Thanks for catching this.

source/getting-started/quickstart.md
tomdale: @trek is correct, but @maclover7 does raise the good point that we don't tell newbs how to kill the server once they've started it. :stuck_out_tongue: Maybe we should open an issue or PR on ember-cli.

source/getting-started/quickstart.md
tomdale: Unknown, I stole the link from the Angular quickstart. It seems like it recommends the binary installers from nodejs.org, which AFAIK is the least error-prone way to go and avoids package manager hell.

source/getting-started/quickstart.md
tomdale: I have a pretty fault memory, so I don't remember the exact discussion you're alluding to, but I have a vague recollection. I think there are two key factors.

First, at the time we discussed it, I was probably operating under the assumption that we would have routable components for Ember 2.0. And (I think?) we all agree that teaching one concept rather than two is a huge win. I suspect this quick start drops routes once RCs land.

Second, I actually still am in favor of the component-first approach for more in-depth tutorials. Here's why: components are ultimately useful for two things. The first is to make UI elements reusable (as we demonstrate in the quick start). The second is *managing the complexity of sophisticated UI*.

By definition, a quick start does not give you enough time to demonstrate the benefits of breaking a large UI into small components and wiring them together. A proper tutorial does. I do think we have time to demonstrate the refactoring-towards-reusability benefit, though, which is why I chose the narrative flow you see here.

*Disclaimer*: This may all be wrong and I would love further critiques of the approach, particularly from folks who may be able to recruit Ember beginners to go through it!

source/getting-started/quickstart.md
tomdale: Good call

source/getting-started/quickstart.md
tomdale: Good call

source/getting-started/quick-start.md
```diff
@@ -0,0 +1,266 @@
...
+In your editor, change the text inside the `<h1>` from `Welcome to
+Ember` to `PeopleTracker` and save the file. Notice that Ember detects
+the change you just made and automatically reloads the page for you in
+the background. You should see that "Welcome to Ember" has been
+replaced by "PeopleTracker".
+
+## Define a Route
+
+Ember makes URLs the heart of your application, so you never build apps
+that feel broken to your users. Before doing anything in Ember, first think
```
dherman: > URLs the heart... never build apps that feel broken to your users.

For someone who hasn't internalized the zen of Ember they won't understand what this really means, or what makes apps broken wrt URLs.

source/getting-started/quick-start.md
```diff
@@ -0,0 +1,266 @@
...
+about your application's URLs.
```
dherman: > first think about your application's URLs.

This isn't actionable advice because you haven't helped the reader understand how to think about URLs. If you can't explain it right here, then you should instead try to get them excited about how either this guide will eventually teach it to them, or if you can't teach it to them in this guide, you should try to get them excited about how _using Ember_ will teach it to them. Example: "Working with Ember will help you learn how to think about designing beautiful URL architectures for your apps."

source/getting-started/quick-start.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+Here, we use the `each` helper to loop over each item in the array we
+provided from the `model()` hook and print it inside an `<li>` element.
+
+## Create a UI Component
+
+As your application grows and you notice you are sharing UI elements
+between multiple pages (or using them multiple times on the same page),
+Ember makes it easy to refactor your templates into reusable components.
```
dherman: Should take the opportunity to also mention your ecosystem of reusable third-part components?

source/getting-started/quick-start.md
```diff
@@ -0,0 +1,262 @@
...
+<h2>List of Scientists</h2>
+
+<ul>
+  {{#each model as |scientist|}}
+    <li>{{scientist}}</li>
+  {{/each}}
+<ul>
+```
+
+Here, we use the `each` helper to loop over each item in the array we
```
dherman: Stuff like this should link to the relevant docs to provide confidence to the reader that they'll be able to chase down that learning thread if/when they need to.

source/getting-started/quick-start.md
```diff
@@ -0,0 +1,262 @@
...
+application&mdash;JavaScript, templates, CSS, web fonts, images, and
+more.
+
+In this case, we told Ember to build for the production environment via
+the `--env` flag. This creates an optimized bundle that's ready to
+upload to your web host. Once the build finishes, you'll find all of the
+concatenated and minified assets in your application's `dist/`
+directory.
+
+The Ember community values collaboration and building common tools that
```
dherman: "community values collaboration" is a slightly confusing garden path noun pileup -- I think I prefer "The Ember community places a high value on collaboration..."

dherman:
I'm so glad you're doing this -- TBH this will help me get started personally!

Is it worth having some sort of hints about where to go from here once you get to the end of the quick start guide?

michaelrkn:
@dherman I'd personally like to see this transition into the rest of the tutorial.

michaelrkn:
I had another idea - what about making this the first page of the Guides? We could shorten up the front matter that's there currently, and get people coding an Ember app right off the bat.

acorncom:
I like the idea of using this quick start to also get people started toward the rest of our more "in-depth" tutorial :+1:

trek:
> I had another idea - what about making this the first page of the Guides? We could shorten up the front matter that's there currently, and get people coding an Ember app right off the bat.

Love this idea.

source/getting-started/quick-start.md
```diff
@@ -0,0 +1,262 @@
...
+curly braces (`{{component}}`). We're going to tell our component:
+
+1. What title to use, via the `title` attribute.
+2. What array of people to use, via the `people` attribute. We'll
+   provide this route's `model` as the list of people.
+
+```app/templates/people.hbs{-1,-2,-3,-4,-5,-6,-7,+8}
+<h2>List of Scientists</h2>
+
+<ul>
```
praxxis: should this be a diff view? it shows both the old code and the new component even though we said "Replace" above

tomdale: When rendered, it shows the previous code as "removed" (note the wonky diff syntax in the file name).

acorncom: @tomdale Re the wonky diff syntax, I'm not sure we're continuing to support that with the new redesign. It's not something we've had a need for up until now, but I've been seeing a few places here (and in the tutorial) where it might be handy.

@locks do you remember if we're adding in styling for diffs?  We'd been contemplating killing that I thought ...

michaelrkn: With the [proposed switch to highlight.js](https://github.com/emberjs/guides/issues/944), does highlight.js support diff styling?

acorncom: @michaelrkn Per this: https://github.com/isagalaev/highlight.js/issues/480 it'll either handle styling a .diff or a .js file but not both

rtablada:
Especially on quick start guide we should avoid using `you` and `you're` POV. Keep instructions using `we` and `our`

tomdale:
@rtablada 
> Especially on quick start guide we should avoid using you and you're POV. Keep instructions using we
> and our. I can work on a PR to update this.

Why?

locks:
@homu r+

PR #1008, opened 2015-11-29
adds comment about extending adapter.handelResponse to include messages

remkoboschker:


michaelrkn:
Thanks for contributing, @remkoboschker! However, as noted in the [Contributing Guidelines](https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md), "The Guides are primarily meant to cover the "Ember happy path", and are not intended to be comprehensive." I think this falls outside the Ember happy path. I'm going to defer to @locks for a final decision on this, though.

remkoboschker:
Alright, no problem. I just think that the example is misleading as is suggests that there will be a `error.status` and, as far as I understand it, `error` will be `true` unless you add some extra error handling. It took me some time to understand what is going on. A comment or link to something like the answer to this question http://stackoverflow.com/questions/31918565/handling-errors-with-the-now-default-ember-data-json-api-adapter will be helpful. I understand that the guides should be concise so as to not overwhelm. And I'll say that I appreciate them very much.

michaelrkn:
Ah, I see, thanks for the explanation and link (and I'm glad that you are finding the Guides useful!). @wecc or @bmac, on https://github.com/emberjs/guides/blob/master/source/routing/loading-and-error-substates.md#the-error-event, can one of you verify that `error.status === 400` doesn't work for the JSON API adapter, and suggest an alternate code snippet? Thanks!!

michaelrkn:
Closed in favor of #1014.

PR #1012, opened 2015-11-30
Remove hyphen from "well separated" after noun.

nubs:
When a compound adjective comes after the noun, it generally shouldn't be hyphenated: http://www.grammarly.com/handbook/punctuation/hyphen/4/hyphen-with-noun-adjective-or-adverb-and-present-p/.


michaelrkn:
Yes sir!

PR #1014, opened 2015-12-01
Remove reference to error.status

michaelrkn:
Fixes the problem noted in #1008.


acorncom:
:+1: Looks good to me

PR #1016, opened 2015-12-02
Add info about upgrading ember

michaelrkn:


michaelrkn:
@locks can you review? If it looks good, please merge into tutorial, then merge tutorial into master.

acorncom:
If folks are going from 1.13 -> 2.0, there are a number of other dependencies that need to get upgraded as well.  If we only put in this advice, folks are likely to end up in a broken state ..

Per https://github.com/ember-cli/ember-cli/issues/4671#issuecomment-161527953, it may be a bit longer before we have an Ember-CLI release that is easy to describe in our tutorial ...

Dhaulagiri:
> there are a number of other dependencies that need to get upgraded as well

@acorncom do you know what those are?  I have upgraded a few new ember-cli apps to latest ember/ember data in the past couple of weeks without needing to do anything with other deps that I can recall.

acorncom:
@Dhaulagiri I could be wrong. From what I remember, there were additional dependencies that were also out of date, but that was prior to the latest releases of Ember-CLI.

acorncom:
@Dhaulagiri I stand corrected, yes, things work fine with ember-cli 1.13.13, thanks!

Regarding the instructions though, I'm a bit more ambivalent about making these recommendations.  When I follow the instructions, I'm presented with the following (bower's choose your own adventure per stefanpenner):

```
Unable to find a suitable version for ember, please choose one:
    1) ember#>= 1.8.1 < 2.0.0 which resolved to 1.13.11 and is required by ember-data#1.13.15
    2) ember#2.2.0 which resolved to 2.2.0 and is required by ember-rentals
    3) ember#>=1.4 which resolved to 2.2.0 and is required by ember-cli-shims#0.0.6
    4) ember#> 1.5.0-beta.3 which resolved to 2.2.0 and is required by ember-resolver#0.1.21
    5) ember#^2.0.0 which resolved to 2.2.0 and is required by ember-data#2.2.0

Prefix the choice with ! to persist it to bower.json

? Answer
```

I don't see that being a great experience when folks are first getting going either ...  It also doesn't properly update `ember-data` in `package.json` which should probably be done if people are going to use this app in the long-run.  That would then need an `npm install` followed by a `bower install`

Per the [latest core notes](https://github.com/emberjs/core-notes/blob/master/ember.js/2015-11/november-13.md) we should be seeing Ember-CLI 2.x coming out at the end of next week, should we just punt this down the road until then (or wait to see if it's needed)?

michaelrkn:
@acorncom agreed, that is not a good user experience. I would like to hold off, but several times over the last couple months I have heard that CLI's release is a week or two away, so I'm not sure how much stock to put in that. Someone on the Ember CLI slack channel said not to worry too much about updating Ember Data in `package.json`, for whatever that's worth.

acorncom:
Sounds good, we'll merge and clean this up once Ember-CLI releases

PR #1027, opened 2015-12-05
Use quotations consistent with rest of guides

HeroicEric:
The rest of the guides use single quotes for javascript and double-quotes in .hbs templates. This makes the updates to keep the tutorial consistent with the rest of the guides.


michaelrkn:
@HeroicEric you are the best!

PR #1028, opened 2015-12-06
Make mirage paths consistent with the rest

gpluta:
I think it is more clear when we keep pointing the full path  every time. Otherwise things might get a bit confusing.


michaelrkn:
Thanks @gpluta!

gpluta:
Cheers

Issue #982, opened 2015-11-14
Add Ember.getOwner section to Factory Lookup section.


rwjblue:
http://guides.emberjs.com/v2.1.0/applications/dependency-injection/#toc_factory-lookups

In the 2.3 guides we should also explain how to use `Ember.getOwner` to do the same things from within an instance of a class that was instantiated by the container.  See the [deprecation guide entry](http://emberjs.com/deprecations/v2.x/#toc_injected-container-access) for a few more details.

tomdale:
@mixonic Is this something you can tackle?

mixonic:
I'm feeling fairly swamped by getting the actions/events stuff fixed up, sorry. Will keep it in mind though.

mitchlloyd:
I just started using this API in an addon and I'm excited about it. I can take a shot at it today.

rwjblue:
Awesome! Thank you!

PR #1029, opened 2015-12-06
Add documentation for `getOwner`

mitchlloyd:
This is a PR to address https://github.com/emberjs/guides/issues/982.

### DIFF
--- a/source/applications/dependency-injection.md
+++ b/source/applications/dependency-injection.md
@@ -182,15 +182,26 @@ export default Ember.Component.extend({
 
 ## Factory Instance Lookups
 
-The vast majority of Ember registrations and lookups are performed implicitly.
+To fetch an instantiated factory from the running application you can call the
+[`lookup`][3] method on an application instance. This method takes a string
+to identify a factory and returns the appropriate object.
 
-In the rare cases in which you want to perform an explicit lookup of an instance of a registered factory,
-you can do so on an application instance in its associated instance initializer.
-For example:
+```javascript
+applicationInstance.lookup('factory-type:factory-name');
+```
+
+The application instance is passed to Ember's instance initializer hooks and it
+is added as the "owner" of each object that was instantiated by the application
+instance.
+
+### Using an Application Instance Within an Instance Initializer
+
+Instance initializers receive an application instance as an argument, providing
+an opportunity to look up an instance of a registered factory.
 
 ```app/instance-initializers/logger.js
 export function initialize(applicationInstance) {
-  var logger = applicationInstance.lookup('logger:main');
+  let logger = applicationInstance.lookup('logger:main');
 
   logger.log('Hello from the instance initializer!');
 }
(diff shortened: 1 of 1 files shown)


michaelrkn:
Thanks @mitchlloyd! This looks great to me. @mixonic could you take a quick look over since I am not super familiar with this feature?

We'll hold off on merging this until we're closer to the 2.3 release so that we can re-re-release 2.2 with fixes before 2.3 drops.

source/applications/dependency-injection.md
```diff
@@ -191,11 +191,17 @@ export default Ember.Component.extend({
 
 ## Factory Instance Lookups
 
-The vast majority of Ember registrations and lookups are performed implicitly.
+You rarely need to write code to look up objects in an Ember application because
```
tomdale: "look up objects" seems both vague and like insider baseball here to me. Look up code from where? When might I be looking objects? I can imagine a developer new to Ember reading this and being confused.

mitchlloyd: I had written "look up dependencies" at one point, but that seemed pretty abstract. Also `import`ing modules is "looking up dependencies". Maybe "You rarely need to use registered factories directly in an Ember Application"?

Any suggestions here?

mixonic: Something like

"In some cases, fetching an instantiated factory from the running application can be useful. The `lookup` method on `Ember.ApplicationInstance` objects provides this functionality. It takes an argument of a factory name, and returns an appropriate object. The application instance object is passed to Ember's instance initializer hooks, and can additionally be referenced as the "owner" of any object that was itself instantiated by the application instance object."

source/applications/dependency-injection.md
```diff
@@ -209,3 +215,28 @@ export default {
   initialize: initialize
 };
 ```
+
+### Getting an Application Instance from a Factory Instance
+
+[`Ember.getOwner`][4] will retrieve the application instance from an Ember
+Object that was instantiated from a registered factory. This means that
```
mixonic: something like

"Factory instances in Ember are "owned" by the application instance object that created them. The `Ember.getOwner` function will retrieve the application instance object relevant to a factory instance, which may be useful for fetching other instances at runtime."

At the least "Ember Object" should probably not capitalize "Object" as it may imply something special about `Ember.Object`.

source/applications/dependency-injection.md
```diff
@@ -209,3 +215,28 @@ export default {
...
+```app/helpers/logger.js
+import Ember from 'ember';
+
+const { getOwner } = Ember;
+
+export default Ember.Helper.extend({
+  init() {
+    this._super(...arguments);
+
+    this.errorReporter = getOwner(this).lookup('error-reporter:sentry');
```
mixonic: A good example here might be to get the name of the error reporting service from an argument to the component:

```js
// Usage: {{logger 'error!!' reportWith='sentry'}}
import Ember from 'ember';
const { getOwner, Helper } = Ember;

export default Helper.extend({
  compute([message], {reportWith}) {
    let errorReportingServiceName = `service:${reportWith}`;
    let applicationInstance = getOwner(this);
    let errorReporter = applicationInstance.lookup(errorReportingServiceName);
    errorReporter.report(message);
  }
});
```

Even better might be an example that did currency formatting or date formatting. Generally you do not want helpers to do things with side-effects (like log), and thus the example of a logger might be an anti-pattern.

source/applications/dependency-injection.md
mitchlloyd: I removed the concept of "you rarely need to do this" and "in some cases you might do this" for now. I'd rather just tell people how to do it and not talk about how often they should look to this technique unless we have some clear guidelines.

Reading again I feel that "look up objects" is less insider than the other phrases I'm reading for this in the docs. However, I do want to be as consistent as possible with the rest of the documentation so I've changed that. I went with @mixonic's "fetching an instantiated factory".

source/applications/dependency-injection.md
mitchlloyd: I worked "owns" and "owner" into the added documentation. There is a description about the meaning of "owner" at the top of this documentation so I didn't want to repeat this too many times.

source/applications/dependency-injection.md
mitchlloyd: Yeah this logging example wasn't working. I stole the audio example from the Ember 2.3 beta release blog instead and made a few minor changes. If you have any other nice use cases in mind let me know.

My current use case for `getOwner` is to dig into the registry and pull out private Ember internals -- not an exemplary use case.

mixonic:
Seems pretty good to me, requires a rebase @mitchlloyd and lets do it

rwjblue:
Thank you, @mitchlloyd!

rwjblue:
Looks like I shouldn't have merged this just yet. Sorry about that.

rwjblue:
I missed https://github.com/emberjs/guides/pull/1029#issuecomment-162383888 when reviewing so I am going to revert for now. So sorry about the confusion and extra work y'all.

michaelrkn:
:) no worries!

PR #1031, opened 2015-12-07
Fix typo

Gpx:
The article can either be *"an"* or *"the"*. I think the latter is more appropriate because we're talking about a specific array: the one we're iterating over.


michaelrkn:
Thanks @Gpx!

Gpx:
:+1:

PR #1032, opened 2015-12-07
Update example to use ES6 template-strings

Gpx:
It seems to me that the Guide is using ES6 template-strings.


acorncom:
Thanks @Gpx

Gpx:
@acorncom :+1:

PR #1034, opened 2015-12-07
Removed webkit-font-smoothing to ensure legibility

pichfl:
Since the new Guides now use light on dark code blocks, the `-webkit-font-smoothing` property makes thinner fonts really thin. While that might look good in certain settings it also reduces readability.

### DIFF
--- a/source/stylesheets/components/_highlight.scss
+++ b/source/stylesheets/components/_highlight.scss
@@ -1,4 +1,5 @@
 .highlight {
+  -webkit-font-smoothing: auto;
   border-radius: $base-border-radius;
   font-family: $monospace-font-family;
   font-size: 0.8em;


michaelrkn:
Hmm... I know very little about this stuff. @acorncom @locks @josemarluedke do any of you have any thoughts on this?

pichfl:
Besides personal choices, I think the default browser behavior should not be changed in most cases.

Here is a quick visual comparison of the current state and my proposed change.
With `-webkit-font-smoothing`
1 before

Without `-webkit-font-smoothing`
2 after

michaelrkn:
It looks like this was from the CSS that Thoughtbot gave to us. I'm fine with changing it, but I want to wait for at least one other person to sign off, preferably someone who knows more than I do about this.

rwjblue:
I'm in the same boat as @michaelrkn, the "after" version looks better but I'd prefer for someone more in tune with the newer design to weigh in.

@wifelette - Thoughts?

acorncom:
@pichfl Agreed on the change increasing legibility improve in code blocks.  However your change also noticeably affects the look of the nav bars, headers and body font throughout the rest of the site.  I don't particularly like what the change does to the rest of the design outside the highlight boxes.

Could you amend your commit to the `.highlight` sections and then we'll merge it?  Thanks!

pichfl:
I'll amend my request right away but I wanted to say that others agree with me that you [should](http://usabilitypost.com/2010/08/26/font-smoothing/) [not change](http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/) [this](https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth) at all.

michaelrkn:
@pichfl Thanks for the links. The MDN seems to recommend just removing that line entirely. Does that seem like the best solution to you? I'm happy to follow the MDN's recommendations.

pichfl:
Yes, would be the best solution to me. I understand the personal preference and design implications though, so for now just changing the code blocks back works for me as well. 

So you merge this right away and change the global setting later on, if everyone is fine with that :smile:

michaelrkn:
@acorncom any thoughts?

michaelrkn:
If you remove the `-webkit-font-smoothing` and `-webkit-font-feature-settings`, the rendering in Chrome looks much more similar to the rendering in Firefox. That leads me to think even more that we should remove these lines.

acorncom:
@pichfl Thanks, just merged this.  Re: doing it for the rest of the site, I'm not a huge fan of having to redo all the CSS font-weights at the moment to clean up after adjust this for the body, as we just got the new design merged in ;-)

If you want to send in a new PR to make the new design more closely match what we have now (minus the font-smoothing), would be happy to merge that in.  Otherwise, let's leave the body styling alone for now. :+1:

krisselden:
Please stop, do not make fonts weights lighter after removing this, that was the point of normalizing in the first place, you will make the guides harder to read on everything but a Mac and that is much less ok than non standard CSS.   The only way it is ok to remove this is to ensure the weights look good on other platforms, if you think it is too thick on a Mac, than removing subpixel antialiasing is the only fix for it to look consistent across platform.

michaelrkn:
@krisselden Can you please move this conversation over to #1058 and/or open a PR if appropriate? Thanks!

PR #1043, opened 2015-12-10
Tweak the description of the .create() method

### LINKED ISSUE #1038: [The Object Model] - Classes and instances - description of the .create() method
In the guides it says:
For performance reasons, note that you ___cannot___ redefine an instance's computed properties or methods when calling create(), nor can you define new ones. You should only set simple properties...

Should it not state:
For performance reasons, note that you ___should not___ redefine...
?

Cheers,

gpluta:
The previous description was not in sync with reality. I hope my proposition is more accurate.

Fixes #1038.


michaelrkn:
Thanks @gpluta!

PR #1044, opened 2015-12-10
Fix incorrect documentation links

fzdvincent:
The links used in the "Escaping HTML Content" subsection were clobbering the links set in the "Class-based Helpers" subsection.


trek:
Thanks!

PR #1045, opened 2015-12-10
Fix code snippet showing Router definition.

nicholalexander:
I noticed a difference in the code that is generated by the ember new command and the example snippet and updated so that it matches what the developer sees.  Small change.


rwjblue:
Hmmm, this seems to included *MANY* commits due to targetting the tutorial branch, can you close this PR and submit against master?

PR #1046, opened 2015-12-10
Fix code snippet showing Router definition

nicholalexander:
I noticed a difference in the code that is generated by the ember new command and the example snippet and updated so that it matches what the developer sees. Small change.

(sorry, previously opened against wrong branch)


rwjblue:
:+1:

acorncom:
Thanks!

nicholalexander:
sure!

PR #1048, opened 2015-12-10
create original images folder and upgrade ember-core-concepts.png

### LINKED ISSUE #1025: Update core concepts image with higher quality
The nice new image was submitted by @alexG53090 in #987 (and originally #983), but it has a problem:

This image is of fairly low quality and doesn't look that great in practice, unfortunately - at least not on a retina screen. In use here: http://guides.emberjs.com/v2.2.0/getting-started/core-concepts/

Some possible solutions:

- Make it twice the size and serve that with width set on the `img`-tag
- Use some kind of method of serving a retina version to retina screens
- Recreate it inline with markup
- Use svg or canvas

The original files seem to be available in the original issue: #979

nicholalexander:
fixes #1025 - create /source/images/originals with originals, upgrade /source/images/ember-core-concepts/ember-core-concepts.png to full resolution (only 97k) so I thought it was fine but would happy to optimize it if needed.


michaelrkn:
Thanks @nicholalexander! For the originals, what do you think about just including the SVG to save space, since the others can be generated from it?

nicholalexander:
@michaelrkn No prob.  I'm neutral on it.  I had thought that having the different formats might be useful to someone further down the line if they needed to do something with them but didn't want - or didn't know how - to convert.  Also, the project is pretty small right now so I wasn't worried about space.  Happy to clean out the files, though and leave just the SVG.  Lemme know!

locks:
:+1: on just including the SVGs

nicholalexander:
updated the pr so that there is only the svg file.

michaelrkn:
Thanks! Can you squash the commits and we'll get this merged?

nicholalexander:
@michaelrkn squashed.  let me know if that's ok.  thanks again!

michaelrkn:
Thanks @nicholalexander!

nicholalexander:
Awesome. My pleasure, thanks @michaelrkn

PR #1053, opened 2015-12-11
Update line break guidelines

michaelrkn:


rtablada:
:aww:

locks:
:-1:

rtablada:
@locks more of a guideline for when to line break. It's not going to be enforcable automatically, but the 120 is already not enforced. This gives an idea of when to line break.

The line breaks will also help in git diff and merge conflict which can lead to a hard time contributing to docs.

michaelrkn:
@acorncom what are your thoughts?

acorncom:
I'm fine with merging it, but don't like the idea of simply out-voting people. @locks, after further thought are you ok with merging this?

locks:
No, but I've been out-voted. :cry: :cry:

PR #1054, opened 2015-12-11
added about to the URL in route example

mattw59:
Just a suggestion: it could be confusing for a new developer if left out.


michaelrkn:
Thanks!

PR #1059, opened 2015-12-12
Fix typo

AlexKVal:
It seems the `as` is out of place in this sentence:
`...meaning that Ember will call it for us as during different times in our app`
:cherries:


rwjblue:
:+1:

michaelrkn:
Thanks @AlexKVal!

PR #1060, opened 2015-12-13
[DOC] JSON definition of Sideloaded Data is wrong

frunjik:
The JSON is now valid,
but the data is wrong ... the ']' needed to be removed instead of added ...

(sorry)

### DIFF
--- a/source/models/customizing-serializers.md
+++ b/source/models/customizing-serializers.md
@@ -73,7 +73,7 @@ should look like this:
 
 ```js
 {
-  "data": [{
+  "data": {
     "type": "articles",
     "id": "1",
     "attributes": {
@@ -90,7 +90,7 @@ should look like this:
         ]
       }
     }
-  }],
+  },
   "included": [{
     "type": "comments",
     "id": "5",


locks:
Current one seems fine, see first example after http://jsonapi.org/format/#document-compound-documents. cc @dgeb

rtablada:
@locks the issue that @frunjik is pointing out is that the requested URL is `/people/1` so the `data` should be a single resource not an array?

locks:
@rtablada irrelevant

rtablada:
The official JSON API docs state:

> For example, the following primary data is a single resource object:

```
{
  "data": {
    "type": "articles",
    "id": "1",
    "attributes": {
      // ... this article's attributes
    },
    "relationships": {
      // ... this article's relationships
    }
  }
}
```

frunjik:
@locks My point is indeed that the example is requesting a single resource object via '/people/1'.
Since i'm an absolute beginner i don't know if 'data' being an array is breaking.
Showing a single object looks less confusing to me.

dgeb:
It's true that `data` in responses should be an array when resource collections are requested vs. a single object when individual resources are requested. See http://jsonapi.org/format/#fetching-resources-responses-200

If the request is to `/people/1`, then it's probably safe to assume that it's a request for an individual resource.

PR #1063, opened 2015-12-13
Fix #1058 Remove font smoothing

### LINKED ISSUE #1058: Remove font smoothing
Currently, our styles use font smoothing, which is [non-standard and should not be used](https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth). This should be removed, and the styles updated so that the font still looks good. 

The relevant properties are `-webkit-font-smoothing` and `-moz-osx-font-smoothing`, and they are in `_highlight.scss`, `_typography.scss`, and `_buttons.scss`.

nightire:


locks:
Appreciated!

PR #1066, opened 2015-12-13
Revert "Fix #1058 Remove font smoothing"

### LINKED ISSUE #1058: Remove font smoothing

michaelrkn:
Reverts emberjs/guides#1063

### DIFF
--- a/source/stylesheets/base/_buttons.scss
+++ b/source/stylesheets/base/_buttons.scss
@@ -1,6 +1,8 @@
 #{$all-button-inputs},
 button {
   @include appearance(none);
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
   background-color: $action-color;
   border-radius: $base-border-radius;
   border: none;
--- a/source/stylesheets/base/_typography.scss
+++ b/source/stylesheets/base/_typography.scss
@@ -2,6 +2,8 @@
 
 body {
   @include font-feature-settings("kern", "liga", "pnum");
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
   color: $base-font-color;
   font-family: $base-font-family;
   font-size: $base-font-size;
--- a/source/stylesheets/components/_highlight.scss
+++ b/source/stylesheets/components/_highlight.scss
@@ -1,4 +1,6 @@
 .highlight {
+  -webkit-font-smoothing: auto;
+  -moz-osx-font-smoothing: auto;
   border-radius: $base-border-radius;
   font-family: $monospace-font-family;
   font-size: 0.8em;


rwjblue:
Was there an issue with the PR?

michaelrkn:
Hey @locks and @nightire - Unfortunately, removing font smoothing is only half the story. As noted in #1058, the styles also need to be updated so that the font still looks good. Without the font smoothing, the type looks very thick. Maybe all that's needed is to reduce the font weight, but I don't know.

krisselden:
It only looks thick to people who have font smoothing supported and only when content is unscaled.  Personally I think that antialias is like browser reset css and the design should be based on what it looks like without font smoothing or it should be reset.  If you design with then it is not legible without.  I do not get why this is hard to understand.  Font smoothing is a fast dated hack that only works in some situations, even when smoothing and if you believe in normalize css files then you should support disabling it.

michaelrkn:
@krisselden I'm not a designer or a proficient CSS developer, so all of that is Greek to me :) If you're able to help out with getting this resolved we'd love for you to contribute!

krisselden:
Simplified, the font weight should be bumped up and the font smoothing set to antialias and designers that design on a Mac any cross platform content (including iOS) should disable it (you can do this in Sketch) so they'll pick a font weight that will look good on all devices.

On a high res screen disabling/enabling it looks equivalent to a font weight bump of about 100-200. So if the font weight is Light, just make it Normal, if Book, just use Bold, then explicitly set antialias to have everyone see the same thing.

And if someone says we shouldn't be "fixing" this ignore them, they most likely don't know what the actual trick is, and why it isn't used on iOS.  If you have an iPhone, doesn't the text look pretty?  At some point subpixel antialiasing will die, hopefully designers will start choosing weights based on what everyone will see (no phone does subpixel antialiasing that I know off).

nightire:
Since English is not my native language, maybe I can't feel exactly as you guys did about the problem `font-smoothing` causes here. But I've tested under:

1. windows chrome / firefox
2. os x chrome / firefox
3. iPhone 6 / iPad 2 Air chrome

and I don't see any problem with or without `font-smoothing` settings, there are some trivial differences but not too bad to reading.

`font-smoothing` is a non-standard property and not the standards track as well, it should not be used in any production level web application, so remove them is a correct step. For other problems (that not everyone will notice/know about), we should do other optimization to fix or improve them.

krisselden:
The non standard CSS does nothing other than make Mac look like everything else. We should not remove it.

It belongs in a CSS reset applied to all text.

If we remove it, someone looking on a Mac will attempt to try to use a Light weight font on body text again which will look illegible on every other device.

PR #1068, opened 2015-12-14
Quote curlies

chadhietala:
I believe that we should encourage people to quote attributes.

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -6,7 +6,7 @@ to an image:
 
 ```handlebars
 <div id="logo">
-  <img src={{logoUrl}} alt="Logo">
+  <img src="{{logoUrl}}" alt="Logo">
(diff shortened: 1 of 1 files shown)


rwjblue:
I completely agree that we should generally be suggesting use of quotes around curlies, but both of these examples might be correct as is.  

> `Logo`

When `logoUrl` is `undefined` or `null`, this example will not trigger any assets to be fetched.  If quoted, you will see a request for `"null"` or `"undefined"` endpoints.

> <input type="checkbox" disabled={{isAdministrator}}>

`disabled` is a boolean attribute, and its mere presence makes the `input` disabled.  When quoted, the value of `disabled` will always be a string (and therefore present), and no value for `isAdministrator` can make the `input` non-disabled.

locks:
I am closing this PR because I agree with @rwjblue's explanations of why they're unquoted.

Issue #1061, opened 2015-12-13
Add more details to the Transforms section


skaterdav85:
@acorncom mentioned on Twitter that I could improve the transforms documentation. I'd like to include more details under the transform section of the guide Defining Models about number, boolean, and date transforms similar to my blog post. Also, maybe it'd be useful to add a guide "Creating Custom Transforms" under Models, similar to the example in my blog post.

michaelrkn:
Awesome! Go ahead and open a PR.

skaterdav85:
@acorncom Do you think it'd be useful adding a new guide "Creating Custom Transforms" under Models?

acorncom:
@skaterdav85 Seems like it fits better in context on the "Defining Models" page, don't you think?  Seems like if we added another section beneath Transforms called "Custom Transforms" that that would give us the needed space.  What we might end up doing as I look at the pages would be to split out simple attributes from relationships (which could go on their own separate page).

@michaelrkn @locks Thoughts on the above?

michaelrkn:
@acorncom sounds good to me!

skaterdav85:
What do you think about taking the "Defining Relationships" section under Models and moving it to its own guide right before "Working with Relationships"?

locks:
@skaterdav85 that would be a different issue

skaterdav85:
ah yes, sorry. will open that.

PR #1069, opened 2015-12-14
adds more details to transforms section #1061

skaterdav85:
I tried the rebase and think I messed things up so I just reforked and made a single commit :)


acorncom:
:+1: Thanks!

PR #1074, opened 2015-12-15
links each mention of a serializer to its corresponding API page

### LINKED ISSUE #1071: Customizing Serializers enhancements
I think it would be useful to add a couple sections to the Customizing Serializers guide for the JSONSerializer and RESTSerializer. I wrote about it in a blog post because I had a hard time finding it in the guides: http://thejsguy.com/2015/12/05/which-ember-data-serializer-should-i-use.html

What do you think?

skaterdav85:
Fixes #1071

@acorncom


michaelrkn:
Thanks!!

PR #1075, opened 2015-12-15
adds a section on creating custom transforms

skaterdav85:
#1061 @acorncom If this is cool, we can chat about splitting up the attributes from relationships, like you mentioned :)


acorncom:
:+1:

acorncom:
Yep, let's discuss splitting the pages out, I think that could make the page less unwieldy

Issue #1076, opened 2015-12-15
break out "Defining Relationships" section


skaterdav85:
@acorncom As discussed in #1061, what do you think about taking the "Defining Relationships" section under Models and moving it to its own guide right before "Working with Relationships"?

acorncom:
Yes, I think that'd be fine. We should check the pages before it in the guides to make sure they don't use relationships though ...

PR #1077, opened 2015-12-15
#1076 break out "Defining Relationships" section

skaterdav85:
@acorncom looks like model relationships are mentioned in those sections so it makes more sense to keep it right after Defining Models


acorncom:
@locks @michaelrkn you guys ok with this?

michaelrkn:
Would it make sense to combine Working With Relationships into this page and just call it Relationships?

skaterdav85:
The only downside to that is that "Working with Relationships" talks about creating records, updating records, etc, which readers wouldn't have seen yet since "Creating, Updating, and Deleting" follows that.

acorncom:
Fixes #1076

michaelrkn:
What if we moved the new Relationships page after Creating, Updating and Deleting?

skaterdav85:
Creating, Updating, and Deleting mentions relationships :laughing:

michaelrkn:
Hrm, could we pull that info out? It looks like just two little things at the beginning and end.

skaterdav85:
Ya. I think the stuff about relationships in "Creating, Updating and Deleting" would be better suited for "Working with Relationships"

acorncom:
:+1:

PR #1078, opened 2015-12-15
typo

tomasznapieralski:


michaelrkn:
```shell
$ ember help
...
ember serve <options...>
  Builds and serves your app, rebuilding on file changes.
  aliases: server, s
```

trek:
Thanks for the submission, but [`serve`](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js#L18) is the correct command name. [`server`](https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js#L20) is only an alias.

tomasznapieralski:
ok, thanks for explanation :+1:

PR #1079, opened 2015-12-16
var store updates

skaterdav85:
replaces `var store` with `this.store` to be consistent with other store references


acorncom:
Thanks!

PR #1083, opened 2015-12-16
Inject stub service as proper property name

thijsvdanker:
The component uses the `locationService` to inject the location service, while the test injects it as `location`.


locks:
@homu r+

toddjordan:
This breaks the example.  `this.inject.service` injects the service into the test context, so you can access the registered service within your testcase.  If you look later on in the example, the location services is accessed as 'location'.  If you want to pull the guide testing examples for trying them out, you can look at https://github.com/toddjordan/ember-testing-guide-examples

thijsvdanker:
Ah, I overlooked that part, sorry about that and thanks for the link to the examples!

PR #1086, opened 2015-12-17
Fix the last change that broke example

toddjordan:
Fix examples by updating references to new variable name.  Added comment on inject to explain where its being injected.


locks:
@homu r+

PR #1089, opened 2015-12-17
Add lifecycle hook documentation

toddjordan:
The approach here is as follows: 
List the hooks in order similar to the original blog post (which I think is super useful), and make the hooks linkable to examples further down the page.  That way readers can learn the order of hooks, and also see real examples.

### DIFF
--- a/source/components/the-component-lifecycle.md
+++ b/source/components/the-component-lifecycle.md
@@ -1,25 +1,122 @@
-Part of what makes components so useful is that they let you take
-complete control of a section of the DOM.  This allows for direct DOM
-manipulation, listening and responding to browser events, and using 3rd
-party JavaScript libraries in your Ember app.
+Part of what makes components so useful is that they let you take complete control of a section of the DOM.
+This allows for direct DOM manipulation, listening and responding to browser events, and using 3rd party JavaScript libraries in your Ember app.
 
-As components are rendered, re-rendered and finally removed, Ember
-provides _lifecycle hooks_ that allow you to run code at specific
-times in a component's life.
+As components are rendered, re-rendered and finally removed, Ember provides _lifecycle hooks_ that allow you to run code at specific times in a component's life.
 
-To get the most use out of a component, it is important to understand
-these lifecycle methods. The following hooks are a few of the most
-useful and commonly used in Ember apps.
+To get the most use out of a component, it is important to understand these lifecycle methods.
 
-## Attaching to the Component Element
+## Order of Lifecycle Hooks Called
+Listed below are the component lifecycle hooks in order of execution according to render scenario.
 
-Suppose you want to integrate your favorite date picker library into an
-Ember project. Typically, 3rd party JS/jQuery libraries require a DOM
-element to bind to. So, where is the best place to initialize and attach
-the library?
+### On Initial Render
 
-After a component successfully renders its backing HTML element into the
-DOM, it will trigger its [`didInsertElement()`][did-insert-element] hook.
+1. `didInitAttrs`
+2. [`didReceiveAttrs`](#toc_formatting-component-attributes-with-code-didreceiveattrs-code)
+3. `willRender`
+4. [`didInsertElement`](#toc_integrating-with-third-party-libraries-with-code-didinsertelement-code)
+5. [`didRender`](#toc_making-updates-to-the-rendered-dom-with-code-didrender-code)
+
+### On Re-Render 
+
+1. [`didUpdateAttrs`](#toc_resetting-presentation-state-on-attribute-change-with-code-didupdateattrs-code)
+2. [`didReceiveAttrs`](#toc_formatting-component-attributes-with-code-didreceiveattrs-code)
+3. `willUpdate`
+4. `willRender`
+5. `didUpdate`
+6. [`didRender`](#toc_making-updates-to-the-rendered-dom-with-code-didrender-code)
+
+### On Component Destroy
+
+1. [`willDestroyElement`](#toc_detaching-and-tearing-down-component-elements-with-code-willdestroyelement-code)
+2. `willClearRender`
+3. `didDestroyElement`
+
+## Lifecycle Hook Examples
+
+Below are some samples of ways to use lifecycle hooks within your components.
+
+### Resetting Presentation State on Attribute Change with `didUpdateAttrs`
+
+`didUpdateAttrs` runs when the attributes of a component have changed, but not when the component is re-rendered, via `component.rerender`,
+`component.set`, or changes in models or services used by the template.
+
+A `didUpdateAttrs` is called prior to rerender, you can use this hook to execute code when specific attributes are changed.
+This hook can be an effective alternative to an observer, as it will run prior to a re-render, but after an attribute has changed.
+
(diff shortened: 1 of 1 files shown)


source/components/the-component-lifecycle.md
```diff
@@ -116,6 +301,7 @@ Let's use this hook to cleanup our date picker and event listener from above:
 
 ```js
 willDestroyElement() {
+  this._super(...arguments);
```
locks: could you actually go through the guides in a separate PR and add the super calls?

michaelrkn:
Awesome @toddjordan! I'll add some comments inline, but one thing right off the bat is that the [build failed](https://travis-ci.org/emberjs/guides/builds/97456364#L1556).

toddjordan:
@michaelrkn yeah, fixed.  I didn't realize `bundle exec middleman build` was finding broken links and spelling now.  Good stuff!

michaelrkn:
@toddjordan Yeah, isn't that nice? Hopefully we'll have [linting for our code snippets](https://github.com/emberjs/guides/issues/613) before long too!

michaelrkn:
Here are a couple ideas for organizing the page.

Have an H2 for _Order of Hooks Called_, and below that H3s for _On Initial Render_, _On Re-Render_, and _On Destroy_.

Move the contents of _Things That Trigger a Component Re-Render_ to the _On Re-Render_ section.

Take out the little blurbs about each hook from the _Order of Hooks_ sections, and just include links. I was a little confused at first, thinking those blurbs were going to be the only explanation. Also there's a lot of duplication between initial and re-render, and switching to just links will eliminate the need for all the _same as above_s.

michaelrkn:
Also, we're trying out a new line break strategy:

Manually break lines:
* at the beginning of every new sentence;
* after the last comma in sentences over 120 characters; or
* after logical statements in sentences over 120 characters with no commas.

Can you give this a shot?

toddjordan:
sure

source/components/the-component-lifecycle.md
```diff
@@ -1,25 +1,137 @@
-Part of what makes components so useful is that they let you take
-complete control of a section of the DOM.  This allows for direct DOM
-manipulation, listening and responding to browser events, and using 3rd
-party JavaScript libraries in your Ember app.
+Part of what makes components so useful is that they let you take complete control of a section of the DOM.
+This allows for direct DOM manipulation, listening and responding to browser events, and using 3rd party JavaScript libraries in your Ember app.
```
locks: "direct DOM manipulation"? cc @tomdale @michaelrkn

source/components/the-component-lifecycle.md
```diff
@@ -1,25 +1,137 @@
...
-times in a component's life.
+As components are rendered, re-rendered and finally removed, Ember provides _lifecycle hooks_ that allow you to run code at specific times in a component's life.
 
-To get the most use out of a component, it is important to understand
-these lifecycle methods. The following hooks are a few of the most
-useful and commonly used in Ember apps.
+To get the most use out of a component, it is important to understand these lifecycle methods.
 
-## Attaching to the Component Element
+To make use of these hooks simply override them within your component.
```
locks: I'm not too keen on this bit, nor the note about `_super`. I think https://guides.emberjs.com/v2.2.0/object-model/classes-and-instances/#toc_overriding-parent-class-methods should cover it sufficiently.

Maybe replace the example with something like "We will know see which hooks exist, what order and when are they fired and then in what situations you'd want to use them"?

locks:
This is a note for a future iteration so I don't forget about it. If there isn't an example covering more than one hook, we should come up with one.

source/components/the-component-lifecycle.md
```diff
@@ -1,25 +1,137 @@
...
+    const profile = this.attrs.data.value;
+    if (typeof profile === 'string') {
+      this.set('profile', JSON.parse(profile));
+    } else {
+      this.set('profile', profile);
+  }
+});
+```
+
+### `didInsertElement`
```
locks: e.g.

```markdown
### Integrating with third-party libraries
```

Rest is :+1:.

source/components/the-component-lifecycle.md
```diff
@@ -29,34 +141,32 @@ Ember guarantees that, by the time `didInsertElement()` is called:
    [`$()`][dollar]
    method.
 
-A component's [`$()`][dollar] method allows you to access the
-component's DOM element via jQuery. For example, you can set an
-attribute using jQuery's `attr()` method:
+A component's [`$()`][dollar] method allows you to access the component's DOM element via jQuery.
```
locks: Technically it returns a jQuery element, this is an occasional source of confusion so precise language is necessary.

source/components/the-component-lifecycle.md
toddjordan: I guess the thing that might be confusing is that most of the lifecycle hooks aren't actually functions on the parent class.  It seems to be more of a template pattern where the rendering code will call these methods if they are there.

So a developer might think (as I initially did) that since I'm not overriding any parent functionality, then I don't have to call `_super`, but there still is a good case to use it because of possible mixins to your components which I mention here.

source/components/the-component-lifecycle.md
toddjordan: yeah, I'll title the sections with summaries of examples.

source/components/the-component-lifecycle.md
toddjordan: k

source/components/the-component-lifecycle.md
```diff
@@ -1,25 +1,129 @@
...
+### On Re-Render 
+
+1. [`didUpdateAttrs`](#toc_resetting-presentation-state-on-attribute-change-with-code-didupdateattrs-code)
+2. [`didReceiveAttrs`](#toc_formatting-component-attributes-with-code-didreceiveattrs-code)
+3. `willUpdate`
+4. `willRender`
+5. `didUpdate`
+6. [`didRender`](#toc_making-updates-to-the-rendered-dom-with-code-didrender-code)
+
+#### Things That Trigger a Component Re-Render
```
toddjordan: I think removing this section might clean things up some.  Maybe a better home would be API doc?

source/components/the-component-lifecycle.md
```diff
@@ -1,25 +1,129 @@
...
+For example, if you have a component that renders based on a json configuration, but you want to provide your component with the option of taking the config as a string,
+you can leverage `didReceiveAttrs` to ensure the incoming config is always parsed.
+
+```javascript
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  didReceiveAttrs() {
+    this._super(...arguments);
+    const profile = this.attrs.data.value;
```
krisselden: This was intentionally not documented

source/components/the-component-lifecycle.md
locks: Maybe put it in the API docs and link to it?

acorncom:
@locks now that we're getting back from vacations, we ready to merge this at this point?

locks:
I think so, give it a review first.

acorncom:
@toddjordan Thanks for putting in all the work on this!

Issue #1036, opened 2015-12-07
[Redesign] Fix responsive design issues on mobile with the new guides


acorncom:
The new design is supposed to collapse as the screen size shrinks (and does on desktop screens when the browser window is shrunk), but when viewed on an iPhone the full page is shown.  We probably just need to add a screen-size meta tag to the header to trigger this

nicholalexander:
@acorncom i was taking a look at this and a simple view port fixes the issue - but it seems like it'd be nicer to have the menu collapse into a hamburger (or slide out from the side).  Is that something you would like me to add?

acorncom:
@nicholalexander Yes please!  Might be worth doing in two pull requests so we get the first one done while you work on the addition ...

acorncom:
Oh, and thanks! :+1:

steveszc:
I'll take a shot at the menu. I'll aim to submit a PR by christmas, if not sooner.

michaelrkn:
Thanks @steveszc!

steveszc:
Just submitted a PR to fix this: https://github.com/emberjs/guides/pull/1090

PR #1090, opened 2015-12-17
Improved nav and table of contents on small screens

steveszc:
Reference Issue #1036 
- Added nav element, wai-aria roles, and aria-labels as appropriate to nav and table of contents
- Added collapsable table of contents on small screens
- Added condensed nav menu on small screens

Screenshot of new version:
new-guides

Screenshot of current version: 
old-guides


rtablada:
@steveszc could you give an example of the top nav being 2 wide instead of 3 up? The spacing looks a bit tight for smaller devices (iPhone 4 and such).

steveszc:
Here's a variation with 2 wide instead of 3 wide on the nav.
Personally I prefer 3w x 2h. The original screengrab for 3w x 2h was a 320px wide viewport, which is about as narrow as you can get on smart phones these days. 

2up-menu

locks:
I prefer the compactness of the three-wide version.

michaelrkn:
I prefer 3 wide too.

locks:
Time to call @homu r+

acorncom:
:+1: Thanks @steveszc

PR #1095, opened 2015-12-17
Adds paragraph tag to make it more consistent

locks:


rtablada:
I think I actually had this earmarked when I was doing my initial pass through the tuts and kept myself from changing multiple things.

PR #1096, opened 2015-12-17
Fix non-trucated file examples

rtablada:
In the component tutorial, there were two areas where the example templates were truncated.

However, this is inconsistent with the rest of the tutorials and could lead to developer confusion when following along.


michaelrkn:
Thanks Ryan!

Issue #1024, opened 2015-12-05
[Tutorial] Upgrading


acorncom:
See https://github.com/emberjs/guides/pull/801 for where this goes in our outline

mellatone:
@acorncom what the scope of upgrading? It could end up covering a _lot_ of ground.

acorncom:
@mellatone True that :-)

Maybe some significantly simplified version of these tutorials?

- http://www.jordanhawker.com/posts/128580938346
- http://www.jordanhawker.com/posts/128657408146
- http://www.jordanhawker.com/posts/128725196501

Or a slightly expanded version of the Ember-CLI upgrade docs?  Given this is a tutorial, the later might be a better fit with the tutorials above going in under Upgrading under the main guides ...

@michaelrkn @locks Thoughts?

michaelrkn:
I would suggest holding off on this until CLI 2.x is released, and/or write it targeting CLI 2.x, since the process for upgrading Ember and ED will be different than 1.x (and easier).

For the tutorial, I think we should cover upgrading Ember, ED, and Ember CLI, and dealing with deprecations.

jrhorn424:
Happy to take this when contingencies are met.

toddjordan:
Issue #1374 is swallowing this issue up.  Tutorial stuff will be tracked there going forward.

Issue #1023, opened 2015-12-05
[Tutorial] Add-ons


acorncom:
See https://github.com/emberjs/guides/pull/801 for where this goes in our outline

- dotenv
- pagefront

toddjordan:
Issue #1374 is swallowing this issue up.  Tutorial stuff will be tracked there going forward.

PR #1097, opened 2015-12-18
Instructions to reboot serve after addon install

rtablada:
I think one possible hang up that people working through the guides might have is that they do not restart the `ember serve` command after installing addons. 

This adds a note after installing mirage, but should also be considered when looking at #1023 and #1024


michaelrkn:
Good thinking!

PR #1099, opened 2015-12-18
Fix missing link-to on index.hbs

rtablada:
On the model hook part of the guides, I noticed that the `link-to` helpers at the bottom of the page went missing.


locks:
@homu r+

PR #1104, opened 2015-12-19
Revive glossary

locks:
Tried to do some thingamabob which clearly didn't work.

Replaces #1001.

### DIFF
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -232,3 +232,9 @@
       url: "adding-new-features"
     - title: "Repositories"
       url: "repositories"
+
+- title: "Glossary"
+  url: "glossary"
+  pages:
+    - title: "Glossary"
+      url: "glossary"
--- a/data/spelling-exceptions.txt
+++ b/data/spelling-exceptions.txt
@@ -256,3 +256,17 @@ wildcard
 workflow
 xhr
 xss
+CDN
+CDNs
+CoffeeScript
+TypeScript
+preprocessor
+Linter
+jslint
+interoperable
+SSR
+Transpile
+transpiles
+transpiling
+internet
+plugin
--- a/source/glossary/glossary.md
+++ b/source/glossary/glossary.md
@@ -1,0 +1,85 @@
+Joining a web development community can be a challenge within itself, especially when all the resources you visit assume you're familiar with other technologies that you're not familiar with.
+
+Our goal is to help you avoid that mess and come up to speed as fast as possible; you can consider us your internet friend.
+
+## CDN
+Content Delivery Network
+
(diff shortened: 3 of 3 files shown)


data/pages.yml
```diff
@@ -232,3 +232,9 @@
       url: "adding-new-features"
     - title: "Repositories"
       url: "repositories"
+
+- title: "Glossary"
```
locks: @michaelrkn @acorncom halp

michaelrkn: - title: "Glossary"
  url: "glossary"
  pages:
    - title: "This Is The Actual Glossary"
      url: "oh-hi-im-the-glossary"

michaelrkn:
```
- title: "Glossary section thingy"
  url: "glossary-section-thingy"
  pages:
    - title: "There Is Only One Page Here"
      url: "this-is-the-glossary"
```

locks: :ok_hand:

locks: One thought is to split the glossary into different "categories" (each a file), which adds a problem instead of fixing the current one.
But it's a good thought.

michaelrkn:
```
- title: "Glossaries"
  url: "glossaries"
  pages:
    - title: "Glossary, Part 1"
      url: "glossary-part-one"
    - title: "Glossary, Part 2: Return of the Glossary"
      url: "episode-two"
    - title: "Glossary, Part 3: The Glossary Lives"
      url: "glossary-hard-or-die-trying"
```

locks:
```
- title: "Glossaries"
  url: "glossaries"
  pages:
    - title: "Glossary, Part 1: The Phantom.js Menace"
      url: "glossary-part-one"
```

locks:
I keep feeling this should be up on MDN somewhere.

data/pages.yml
michaelrkn: hahahaha

michaelrkn:
@locks +1

michaelrkn:
https://developer.mozilla.org/en-US/docs/Glossary

michaelrkn:
What about moving any content to https://developer.mozilla.org/en-US/docs/Glossary#Contribute_to_the_glossary that isn't there already and then on the first page of the guides suggesting that people use that as a resource if they come across an unfamiliar term as they're coming up to speed on Ember and the New Front-End Development world? /cc @tomdale @wycats

data/pages.yml
acorncom: You guys are having *way* too much fun here ;-)

locks:
@homu r+

PR #1109, opened 2015-12-21

prastut:
The exact URL isn't specified when we first try to test the ember application. Since we haven't defined the "index" route, nothing would be displayed to the user at that particular time.
It took me a while to realize, so I guess the link should be edited.


locks:
Appreciated! :)

@homu r+

PR #1114, opened 2015-12-23
I struggled a lot installing through homebrew on mac

techsethi:
I struggled a lot installing through homebrew on mac. Even after install, it didn't install npm & only installed node. The installer at this location was super easy & installed both node & npm with a single click very smoothely.


knownasilya:
Did you forget a link to the installer?

locks:
Hi @techsethi, thanks for the feedback. Like @knownasilya mentions, I think you forgot to include the link? If you meant to link to the Node.js installer, it is already referenced in the [previous bullet point](https://github.com/emberjs/guides/pull/1114/files#diff-0bddccd6a3d7e77ab47715d844a840fbR33).
It is also unnecessary to specify the versions installed (we would have to update that every release), or putting the date it was written.

michaelrkn:
I'm closing this out due to lack of activity.

PR #1115, opened 2015-12-23
Update transitionTo and transitionToRoute links

thijsvdanker:
The links didn't point to the corresponding functions in the API.


rwjblue:
@homu r+

rwjblue:
Thank you!!

PR #1117, opened 2015-12-26
Update Mirage Config entry

Millzor2890:
After looking around a little, I found that Ember's default adapter is http://emberjs.com/api/data/classes/DS.RESTAdapter.html which is very particular about the REST responses it is expecting back.  The current mirage file in the walkthrough doesn't follow the structure detailed in this file here: http://emberjs.com/api/data/classes/DS.RESTAdapter.html  and renders nothing in the template, which was kind of a bummer.  Re-organizing the structure of what mirage returns make it work just fine though.

### DIFF
--- a/source/tutorial/ember-data.md
+++ b/source/tutorial/ember-data.md
@@ -64,43 +64,35 @@ Let's now configure Mirage to send back our rentals that we had defined above by
 
 ```app/mirage/config.js
 export default function() {
+
   this.get('/rentals', function() {
-    return {
-      data: [{
-        type: 'rentals',
+    return ({
+      "rentals": [{
         id: 1,
-        attributes: {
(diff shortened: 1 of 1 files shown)


locks:
Thanks for submitting.

`RESTAdapter` was the default in Ember Data 1.x. In Ember Data 2.x the default is now `JSONAPIAdapter` https://guides.emberjs.com/v2.2.0/models/customizing-adapters/#toc_customizing-the-jsonapiadapter (see bullet points just above header).

As stated on [Getting Started - Ember CLI](https://guides.emberjs.com/v2.2.0/getting-started/ember-cli/#toc_upgrading-ember), the versions of the libraries should match the version of the tutorials you are reading.

PR #1118, opened 2015-12-27
fix broken mirage data

brandocorp:
The mock data provided in the example was returning an error and resulting in nothing being loaded. This fixes the data structure so that it properly returns the data to the rental app


michaelrkn:
Looks like you missed the section on [upgrading to the latest version of Ember Data](https://guides.emberjs.com/v2.2.0/getting-started/ember-cli/#toc_upgrading-ember).

PR #1121, opened 2015-12-30
Changed inline code to shell code

mozeryansky:
Trying to be consistent. Also


mozeryansky:
I'd like to point out this line is a shell code wrapped line in the previous page, but this page is the first page of the tutorial so I think it should have the shell line.

locks:
@homu r+

Issue #1125, opened 2015-12-30
Add (temporary) note to Getting Started about upgrading Ember


toobulkeh:
At the end of [The Getting Started / Installing Ember Guide](https://guides.emberjs.com/v2.2.0/getting-started/) it says "You're good to go", but at this point in onboarding you're using Ember 1.13 and Ember-CLI 1.13. While Ember-CLI 2+ hasn't been production released yet, Ember is on 2.2.1 as of this writing.

It took us, as new users to Ember, way too long to figure out that a few extra steps were needed.

After some discussion on Ember's Slack #needhelp, I believe the preferred Ember installation method is still ember-cli, and the current preferred Ember update procedure is updating bower.json manually, [as referenced here](https://guides.emberjs.com/v2.2.0/getting-started/ember-cli/#toc_upgrading-ember).

Seeing as how Ember 2.0.0 was released 5 months ago and Ember-CLI is pending, (but still pending...), I would suggest adding a detailed (yet temporary) note and steps for overriding those versions without upgrading Ember-CLI.

I totally realize this is temporary work, but this is a large gap to a newbie, which I'm assuming these guides are targeted towards.

Note: I mistakenly commented about this in another [seemingly related task](https://github.com/emberjs/guides/issues/1094), which is not at all related. Sorry about that!

cafreeman:
+1 It totally seems trivial now, but it took several different false starts and a lot of googling to finally find an article that actually mentioned how to get Ember and Ember Data up to the current version.

locks:
This has been previosuly discussed and rejected. It is not trivial to update older versions of the guides (which 2.2.0 would be by then).

michaelrkn:
@locks We have not added anything 2.3-specific yet. Several issues have been opened by people going through the tutorial accidentally using old versions of Ember and especially Ember Data. I think this is a real problem and we should get this in before we add anything 2.3-specific.

locks:
@cafreeman https://guides.emberjs.com/v2.2.0/getting-started/ember-cli/#toc_upgrading-ember shows how to update said versions

locks:
@michaelrkn 2.3 of the guides won't be released until after CLI 2.2 so that is a moot point

michaelrkn:
rwjblue [8:07 AM] 
Most likely going to release 2.3 final next weekend (by the 4th). 

michaelrkn [4:00 PM] 
is cli 2.3 going to be released simultaneously?

rwjblue [4:03 PM] 
[...]

PR #1126, opened 2015-12-30
Move paragraph from Ember-CLI to Getting Started

toobulkeh:
To highlight the fact that Ember may not be currently up to date

Fixes https://github.com/emberjs/guides/issues/1125


michaelrkn:
Good idea. I'd prefer that you simply link to the [existing section on upgrading](https://guides.emberjs.com/v2.2.0/getting-started/ember-cli/#toc_upgrading-ember), so that we keep things DRY.

Also, I think we should add a note to the [beginning of the tutorial](https://guides.emberjs.com/v2.2.0/tutorial/routes-and-templates/) after it says "Let's make sure we have a fresh Ember CLI app..." that also points to the upgrading section.

toobulkeh:
Both sound reasonable. I looked for a call-out design in the current guides but didn't find a decent one, which is why I copy/pasted, as the header was large enough to catch the eye while skimming. Is there a callout / alert format that I missed? I also considered a partial, but that would just keep the backend dry - not the text.

I know some technical writers don't like using call outs, so I didn't want to suggest new inserts that weren't wanted.

michaelrkn:
No, there are no call-outs. I think just put it in the text.

acorncom:
@toobulkeh Thanks for pointing this out.  We're going to adjust this slightly in our latest revisions prior to refreshing Ember 2.2 docs, so will close this now.

Issue #1018, opened 2015-12-05
[Tutorial] Autocomplete component


acorncom:
See https://github.com/emberjs/guides/pull/801 for info on where this goes in an outline

acorncom:
@rtablada were you wanting to work on this?

mattw59:
I'm going to take this one unless somebody has already done so.

rtablada:
@mattw59 go ahead I'm working on helpers I think

mattw59:
@rtablada will do

michaelrkn:
@mattw59 we're shooting to get this stuff out with 2.3, which is scheduled for the end of December. Do you think you'll be able to get to this by then?

mattw59:
Yeah, I've gotten a start. I should have something by Monday, I think.

michaelrkn:
Sweet!

michaelrkn:
@mattw59 How is this coming along?

mattw59:
Sort of slowly with the holidays. Will try to get you something to look at
tomorrow. The code is close to done.

PR #1127, opened 2015-12-30
tutorial entry for an autocomplete component

mattw59:
Entry for #1018

### DIFF
--- a/source/tutorial/autocomplete-component.md
+++ b/source/tutorial/autocomplete-component.md
@@ -1,17 +1,22 @@
-As they search for a rental, a user might also want to narrow their search to a specific city. Let's build a component that will let them search for properties within a city, and also suggest cities to them as they type.
+As they search for a rental, users might also want to narrow their search
+ to a specific city. Let's build a component that will let them search for
+  properties within a city, and also suggest cities to them as they type.
 
-To begin, let's generate our new component. We'll call this component `filter-listing`.
+To begin, let's generate our new component. We'll call this component 
+`filter-listing`.
 
 ```shell
 ember g component filter-listing
 ```
-As before, this creates a Handlebars template (`app\templates\components\filter-listing.hbs`) and a JavaScript file (`app\components\filter-listing.js`).
+As before, this creates a Handlebars template 
+(`app/templates/components/filter-listing.hbs`) and a JavaScript file 
+(`app/components/filter-listing.js`).
 
 The Handlebars template looks like this:
 
-```app\templates\components\filter-listing.hbs
-City: {{input value=filter key-up=(action 'keyUp' filter)}} 
-<button {{action 'click'}}>Search</button>
+```app/templates/components/filter-listing.hbs
+City: {{input value=filter key-up=(action 'autoComplete' filter)}} 
+<button {{action 'search'}}>Search</button>
 
 <ul>
 {{#each filteredList as |item|}}
@@ -19,26 +24,34 @@ City: {{input value=filter key-up=(action 'keyUp' filter)}}
 {{/each}}
 </ul>
 ```
-It contains an `input` component, that renders as a text field that the user can type in to look for properties in a given city. The `value` property of the `input` will be bound to the `filter` property in our backing JavaScript object. The `key-up` property will be bound to a `keyUp` action in our backing object, and passes the `filter` property as a parameter.
+It contains an [`{{input}}`](../../templates/input-helpers) helper, that 
+renders as a text field that the user can type in to look for properties 
+in a given city. The `value` property of the `input` will be bound to the 
+`filter` property in our component. The `key-up` property 
+will be bound to a `autoComplete` action in our backing object, and passes
+ the `filter` property as a parameter.
 
-It also contains a button, whose `action` parameter is bound to the `click` action in our component.
+It also contains a button, whose `action` parameter is bound to the 
+`search` action in our component.
 
-Lastly, it contains an unordered list, that uses the `filteredList` property for data, and displays the `city` property of each item in the list. Clicking the list item will fire the `choose` action, which will populate the `input` field with the name of the `city` in the clicked list item.
+Lastly, it contains an unordered list, that uses the `filteredList` 
+property for data, and displays the `city` property of each item in the 
+list. Clicking the list item will fire the `choose` action, which will 
+populate the `input` field with the name of the `city` in the clicked list
+ item.
 
 Here is what the component's JavaScript looks like:
 
-```app\components\filter-listing.js
-import Ember from 'ember';
-
+```app/components/filter-listing.js
 export default Ember.Component.extend({
   filter: null,
   filteredList: null,
   actions: {
-    keyUp() {
-      this.attrs.autoComplete(this.filter);
+    autoComplete() {
+      this.get('autoComplete')(this.get('filter'));
     },
-    click() {
-      this.attrs.search(this.filter);
+    search() {
+      this.get('search')(this.get('filter'));
     },
     choose(city) {
       this.set('filter',city);
@@ -47,11 +60,17 @@ export default Ember.Component.extend({
 });
 
 ```
-There's a property for each of the `filter` and `filteredList`, and actions as described above. What's interesting is that only the `choose` action is defined by the component. The actual logic of each of the `keyUp` and `click` actions are pulled from the component's `attrs` object, which means that those actions need to be passed in by the calling object, a pattern known as _closure actions_.
+There's a property for each of the `filter` and `filteredList`, and 
+actions as described above. What's interesting is that only the `choose` 
+action is defined by the component. The actual logic of each of the 
+`autoComplete` and `search` actions are pulled from the component's 
+properties, which  means that those actions need to be [passed]
+ (../../components/triggering-changes-with-actions/#toc_passing-the-action-to-the-component) 
+ in by the calling object, a pattern known as _closure actions_.
 
 To see how this works, change your `index.hbs` template to look like this:
 
-```app\templates\index.hbs
+```app/templates/index.hbs
 <h1>Welcome to Super Rentals</h1>
 
 We hope you find exactly what you're looking for in a place to stay.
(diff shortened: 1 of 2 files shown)


source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+City: {{input value=filter key-up=(action 'keyUp' filter)}} 
+<button {{action 'click'}}>Search</button>
+
+<ul>
+{{#each filteredList as |item|}}
+  <li {{action 'choose' item.city}}>{{item.city}}</li>
+{{/each}}
+</ul>
+```
+It contains an `input` component, that renders as a text field that the user can type in to look for properties in a given city. The `value` property of the `input` will be bound to the `filter` property in our backing JavaScript object. The `key-up` property will be bound to a `keyUp` action in our backing object, and passes the `filter` property as a parameter.
```
locks: input helper. you can link to the appropriate section of the guide too

locks: The `value` attribute, I think

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+
+It also contains a button, whose `action` parameter is bound to the `click` action in our component.
```
locks: Using `click` is a bit confusing here.

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+
+Lastly, it contains an unordered list, that uses the `filteredList` property for data, and displays the `city` property of each item in the list. Clicking the list item will fire the `choose` action, which will populate the `input` field with the name of the `city` in the clicked list item.
+
+Here is what the component's JavaScript looks like:
+
+```app\components\filter-listing.js
+import Ember from 'ember';
```
locks: `import`s should not be included

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+
+export default Ember.Component.extend({
+  filter: null,
+  filteredList: null,
+  actions: {
+    keyUp() {
+      this.attrs.autoComplete(this.filter);
```
locks: `this.get('autoComplete')(this.get('filter')`

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+    },
+    click() {
+      this.attrs.search(this.filter);
```
locks: see above

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+    },
+    choose(city) {
+      this.set('filter',city);
```
locks: see above

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+});
+
+```
+There's a property for each of the `filter` and `filteredList`, and actions as described above. What's interesting is that only the `choose` action is defined by the component. The actual logic of each of the `keyUp` and `click` actions are pulled from the component's `attrs` object, which means that those actions need to be passed in by the calling object, a pattern known as _closure actions_.
```
locks: Can you review the newline policy in the guidelines to make diffs/comments easier, please?

We shouldn't refer at all to `attrs`, check the wording on the relevant guide and you can even link there.

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,164 @@
...
+Now, define your new controller like so:
+
+```app\controllers\index.js
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  filteredList: null,
+  actions: {
+    autoComplete(param) {
+      if(param !== undefined && param !== "") {
```
locks: Is this really necessary?
And should you be setting `filteredList` to a promise and not the resolved value?

locks:
@mattw59 the usage of event names for action names worries me a bit, as the distinction is already hard for many people. Is there perhaps more semantic names you could give?

source/tutorial/autocomplete-component.md
mattw59: ok

source/tutorial/autocomplete-component.md
mattw59: ok

source/tutorial/autocomplete-component.md
mattw59: agreed

source/tutorial/autocomplete-component.md
mattw59: done

source/tutorial/autocomplete-component.md
mattw59: done

source/tutorial/autocomplete-component.md
mattw59: not sure exactly what you want to see here, sorry

source/tutorial/autocomplete-component.md
mattw59: Yeah, because otherwise if you don't check for empty/undefined, you get weird behavior when you clear out the text box.

mattw59:
That's good feedback about the action names, thanks. Should be better now.

source/tutorial/autocomplete-component.md
mattw59: done

mattw59:
OK, now I've gone and hosed up my rebase. Do you want me to delete this PR and re-submit? Apologies.

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+      }
+    },
+    search(param) {
+      if(param !== "") {
+        this.store.query('rental', {city: param}).then((result) => {
+          this.set('model',result);
+        });
+      }
+      else {
+        this.set('model',null);
```
locks: wouldn't `this.get('model').clear()` be better here?

mattw59: agreed - updated accordingly

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
+As they search for a rental, a user might also want to narrow their search
+ to a specific city. Let's build a component that will let them search for
+  properties within a city, and also suggest cities to them as they type.
```
acorncom: Let change the second clause to the below to keep singular / plurals consistent

> for a rental, users might also want

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+```shell
+ember g component filter-listing
+```
+As before, this creates a Handlebars template 
+(`app\templates\components\filter-listing.hbs`) and a JavaScript file 
+(`app\components\filter-listing.js`).
+
+The Handlebars template looks like this:
+
+```app\templates\components\filter-listing.hbs
```
acorncom: Our paths run the other direction in all our other examples, can you switch these?

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+<ul>
+{{#each filteredList as |item|}}
+  <li {{action 'choose' item.city}}>{{item.city}}</li>
+{{/each}}
+</ul>
+```
+It contains an [`{{input}}`](../../templates/input-helpers) helper, that 
+renders as a text field that the user can type in to look for properties 
+in a given city. The `value` property of the `input` will be bound to the 
+`filter` property in our backing JavaScript object. The `key-up` property
```
acorncom: "in our **component**" seems simpler ...

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+
+Lastly, it contains an unordered list, that uses the `filteredList` 
+property for data, and displays the `city` property of each item in the 
+list. Clicking the list item will fire the `choose` action, which will 
+populate the `input` field with the name of the `city` in the clicked list
+ item.
+
+Here is what the component's JavaScript looks like:
+
+```app\components\filter-listing.js
```
acorncom: Path again

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+      }
+    },
+    search(param) {
+      if(param !== "") {
+        this.store.query('rental', {city: param}).then((result) => {
+          this.set('model',result);
+        });
+      }
+      else {
+        this.set('model'.clear();
```
acorncom: Missing close paran before `.clear()`

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+{{/each}}
+
+{{#link-to 'about'}}About{{/link-to}}
+{{#link-to 'contact'}}Click here to contact us.{{/link-to}}
+```
+We've added the `filter-listing` component to our `index.hbs` template. We 
+then pass in the functions and properties we want the `filter-listing` 
+component to use, so that the `index` page can define some of how it wants 
+the component to behave, and so the component can have some 
+narrowly-defined access to aspects of the calling scope.
```
acorncom: component to behave, and **so the component can use those specific functions and properties** ?  When we start discussing "calling scope", for folks just learning Ember their eyes are likely to glaze over a bit ;-)

source/tutorial/autocomplete-component.md
```diff
@@ -0,0 +1,202 @@
...
+  filteredList: null,
+  actions: {
+    autoComplete(param) {
+      if(param !== "") {
+        this.store.query('rental', {city: param}).then((result) => {
+          this.set('filteredList',result);
+        });
+      }
+      else {
+        this.set('filteredList',null);
```
acorncom: Probably should be consistent and `clear()` this as well

acorncom:
@mattw59 Had a few more suggestions here, but then I'm :+1: on merging this.  @locks any additional thoughts?

source/tutorial/autocomplete-component.md
mattw59: ok

source/tutorial/autocomplete-component.md
mattw59: Sorry, my Windows is showing :)

source/tutorial/autocomplete-component.md
mattw59: OK

source/tutorial/autocomplete-component.md
mattw59: ok

source/tutorial/autocomplete-component.md
mattw59: ok

source/tutorial/autocomplete-component.md
mattw59: Sounds good.

source/tutorial/autocomplete-component.md
mattw59: OK

mattw59:
@acorncom I think I've addressed your comments, but let me know if there's anything else. Thanks for your comments, and you too, @locks.

acorncom:
@mixonic Not sure if you have time, but we're wondering your thoughts on overarching approach to how this component is setup.  The use of route, controller and component (and where some of the things are put) doesn't strike @locks or me as particularly Ember-like, but thought it'd be useful to have someone from Ember core weigh in on suggested placement of things (given the high-profile aspect of this being in the tutorial).  Thoughts?

acorncom:
@mattw59 I know you were hoping to get this merged, but I think you're going to get a decent amount of scrutiny here given the fact this is the tutorial.  Bear with us :-)

mattw59:
@acorncom No worries, I get how important it is for this to be right. I look forward to the feedback.

locks:
@homu r+

PR #1129, opened 2015-12-31
Add reminder to use latest versions in tutorial

michaelrkn:


michaelrkn:
@locks What do you think about this?

locks:
@tomdale suggested that ember-cli 2.x should install the latest stable version of the libraries, so idk

tomdale:
@rwjblue Can you confirm installing Ember CLI 2.x gets you Ember.js 2.x?

mozeryansky:
@tomdale The current ember-cli will install 1.13. The ember-cli that is 2.x is in beta, but I'm using ember-cli 2.2.0-beta.2 and can confirm it does install ember 2.2.0. And for extra confirmation: https://github.com/ember-cli/ember-cli/blob/master/blueprints/app/files/bower.json

acorncom:
This was fixed by #1153

PR #1133, opened 2016-01-02
Removed erroneous key=

mozeryansky:
With keys="@item", I get an error in the browser console:
`Error: You must provide a string key when calling `yieldItem`; you provided undefined`

This issue emberjs/ember.js#12109 mentions deleting `key=`, which fixed the issue.


acorncom:
Related to #1166

locks:
@homu r+

PR #1134, opened 2016-01-02
Fixed ember template error

mozeryansky:
With the '/' character immediately at the end, reports an error:
`An unquoted attribute value must be a string or a mustache, preceeded by whitespace or a '=' character, and followed by whitespace or a '>' character (on line 8)`

Adding a space fixes the error.

### DIFF
--- a/source/templates/actions.md
+++ b/source/templates/actions.md
@@ -116,7 +116,7 @@ event listeners and enables to work with one-way bindings.
 
 ```handlebars
 <label>What's your favorite band?</label>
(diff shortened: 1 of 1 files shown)


mozeryansky:
This is probably more of an error with ember-template-compiler since the HTML is valid. What's the opinion on this?

locks:
Hm, if you can reproduce the error I'd say file a bug report with the compiler repo and see where to go from there.

locks:
@homu r+

PR #1137, opened 2016-01-04
Add spec to check for unused spelling exceptions, resolves #1105

### LINKED ISSUE #1105: Keep spell checker exceptions file clean
We should have a script that's run, preferably with the tests, to make sure that every word in the spell checker exception file is actually used.

While this may lead to some churn in the future if said word is reintroduced by another commit, I think the benefit of making the writer more aware outweighs it.

stevekinney:
This is a first pass at addressing #1105. It does a quick scan through the documentation for a particular word. If it is not found, the expectation fails and the developer is notified as to which word was not found.

To my surprise, scanning the entire set of guides is relatively quick. I tried it with 1,000 words and there was no performance issue.


stevekinney:
Ugh, I got one of those git issues where it's easier to give it another shot than to try to resolve it. Disregard this PR, I'll open another one momentarily.

PR #1142, opened 2016-01-06
Create Creating a New App

rovo79:
Migrated this content from the Getting Started-> Ember CLI page to a new page under Tutorial.

### DIFF
--- a/source/tutorial/Creating a New App
+++ b/source/tutorial/Creating a New App
@@ -1,0 +1,124 @@
+## Creating a New App
+
+To create a new project using Ember CLI, use the `new` command. In preparation
+for the tutorial in the next section, you can make an app called `super-rentals`.
+
+```shell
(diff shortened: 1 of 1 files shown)


locks:
I'm not sure this is a change we want to make, but at any rate the change should be made in a single PR so we don't lose the text in git.

rovo79:
I agree. Just trying to point out that content seems to belong to the Tutorial section, and not the Getting Started section.

acorncom:
@rovo79 Thanks for the note on this.  We're in the process of cleaning things up there (the tutorial section is still in the early stages of being re-written), so will close this.  Do let us know if you find any other bugs though!

rovo79:
Thanks so much. Overall, I think the guides coupled with CLI did an amazing job onboarding me quickly.

PR #1143, opened 2016-01-06

rovo79:
Will not need to instruct starting Ember Serve twice, if it's indicated already in a new page:
Creating a New App
https://github.com/emberjs/guides/pull/1142


locks:
I don't think it's a problem to have this information repeated, especially because at the top of the guide we tell people to generate a fresh application.
I am also slightly tweaking the tutorial, so this might change.

PR #1145, opened 2016-01-06
Update model-hook.md

rovo79:
Should they be changed to Rentals? I might be misunderstanding this part.


locks:
No, the model itself is an array of rentals. What's yielded in the block param (`|rental|`) is each rental in the array.

rovo79:
Ah, ok. thank you.

PR #1151, opened 2016-01-07
Remove whitespace from conditional on components guide

jsmestad:


locks:
Appreciated!

@homu r+

PR #1152, opened 2016-01-07
Remove `px` from <img> width attribute

jsmestad:
As suggested by HTML Tidy. According to MDN reference, the width attribute assumes pixels without a unit. [ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)


jsmestad:
Fixed.

locks:
@homu r+

jsmestad:
Not sure how to resolve the squashing failure. Tips?

locks:
@jsmestad can you reach me in #documentation on Slack?

PR #1154, opened 2016-01-08
Update customizing-a-components-element.md

Znegl:
Added information about how to remove unused bound attributes from the DOM.

### DIFF
--- a/source/components/customizing-a-components-element.md
+++ b/source/components/customizing-a-components-element.md
@@ -153,7 +153,7 @@ export default Ember.Component.extend({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


locks:
I think the information is useful, but I find the phrasing confusing. Can you change it to be something more like "if the attribute is null, it won't be rendered"?

acorncom:
@Znegl Thanks for the work.  One last request.  Could you squash your commits now that we've got this in good shape?  We generally want to have a single commit for these types of changes.  Thanks! :+1:

locks:
@homu r+

@acorncom fingers crossed

PR #1158, opened 2016-01-11
Update relationships.md

Mak-Di:
Typo fix


locks:
Thanks for your contribution, but the current grammar is the more appropriate one. See consonant sound examples at the bottom of https://www.englishclub.com/pronunciation/a-an.htm.

PR #1164, opened 2016-01-11
Update formatCurrency helper to match Ember-CLI

jsmestad:
When running `ember g helper format-currency` the generated file follows a different format than the guides show. This is using `ember-cli 2.2.0-beta.3`

### DIFF
--- a/source/templates/writing-helpers.md
+++ b/source/templates/writing-helpers.md
@@ -42,7 +42,7 @@ That file should export a function wrapped with [`Ember.Helper.helper()`][1]:
 [1]: http://emberjs.com/api/classes/Ember.Helper.html#method_helper
 
 ```app/helpers/format-currency.js
-export default Ember.Helper.helper(function(params) {
+export function formatCurrency(params) {
   let value = params[0],
       dollars = Math.floor(value / 100),
       cents = value % 100,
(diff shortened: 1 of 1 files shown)


source/templates/writing-helpers.md
```diff
@@ -42,7 +42,9 @@ That file should export a function wrapped with [`Ember.Helper.helper()`][1]:
 [1]: http://emberjs.com/api/classes/Ember.Helper.html#method_helper
 
 ```app/helpers/format-currency.js
-export default Ember.Helper.helper(function(params) {
+import Ember from 'ember';
```
locks: can you omit the import please?

acorncom:
@jsmestad Background on the request above: what you're putting in is valid, but we're not quite ready to start putting in import statements across all the docs due to some future changes that may affect that.  Thanks!

jsmestad:
@acorncom it may be useful in that case to just call out that running that command could have a different looking output. I'm happy to draft that if your ok with accepting something like that.

locks:
We're already omitting `import`s everywhere except the tutorial. Another thorn is that the blueprints are in CLI until Ember gets completely addonified.

tl;dr no call out

jsmestad:
alright better luck next time ;)

locks:
what about the other changes? :x

jsmestad:
@locks doh! your only talking about the import line, not the export. &#^!!

jsmestad:
Removed `import` statement as requested. Rest should be :+1:

acorncom:
@jsmestad Sweet.  Mind squashing your commits?  We'll merge it right after that :+1:

jsmestad:
@locks @acorncom squashed.

locks:
@homu r+

PR #1169, opened 2016-01-14
Don't have an application folder

rovo79:
Up until this point of the guide, I don't have an application folder. I have an app folder.


locks:
The implicit root route is the application route. You can confirm this by opening Ember Inspector and looking at the top of either `View Tree` or `Routes`.

I've submitted #1171 to introduce what I think is a missing word, can you take a look?

PR #1170, opened 2016-01-14
Update defining-your-routes.md

rovo79:
It seems to distract the reader by suggesting to go to the next section when only one paragraph away from being done with the current page.


locks:
@homu r+

locks:
@homu r+

PR #1172, opened 2016-01-15
asynchronous-routing: then -> catch

jamesarosen:
In the section on recovering from rejection, the text describes using `catch`, but the code example uses `then`, which wouldn't recover from the rejected promise.

### DIFF
--- a/source/routing/asynchronous-routing.md
+++ b/source/routing/asynchronous-routing.md
@@ -156,7 +156,7 @@ them into fulfills that won't halt the transition.
 ```app/routes/funky.js
 export default Ember.Route.extend({
   model() {
-    return iHopeThisWorks().then(null, function() {
+    return iHopeThisWorks().catch(function() {
       // Promise rejected, fulfill with some default value to
       // use as the route's model and continue on with the transition
       return { msg: 'Recovered from rejected promise' };


rwjblue:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch

> The catch() method returns a Promise and deals with rejected cases only. It behaves the same as calling Promise.prototype.then(undefined, onRejected).

jamesarosen:
I missed the `null` in my reading. It's fine as-is.

Though I do think `.catch(fn)` is more common than `.then(null, fn)` in the Ember community.

rwjblue:
Totally agreed, but this PR is invalid as is (it would need to be changed to `.catch(function(){})` instead of `.catch(null, function() { })` to work).

jamesarosen:
My brain. It is not at the moment. Will fix tomorrow and comment.

rwjblue:
:+1: - I definitely agree we should use the easier and more idiomatic `.catch` here.

acorncom:
@jamesarosen Could you update here?  Would love to get this merged ...

jamesarosen:
Re-pushed

acorncom:
@homu r+

acorncom:
Thanks!

PR #1179, opened 2016-01-20
Prevent text wrapping on TOC hover animation

### LINKED ISSUE #1174: Sidebar section links may wrap on hover
I'll start with a gif:

emberguides

Some section name lengths may cause wrapping the link to another line when hovered, but are short enough to stay on one line otherwise. This can be disruptive while navigating through the Guides.

kpfefferle:
Resolves #1174 

Old behavior:
old

New behavior:
new


acorncom:
@homu r+

acorncom:
Looks good, thanks!

PR #1180, opened 2016-01-21
Replace the root element of "data" with "rental".

mTorbin:
The way that the tutorial is written results in a non-functional app since the root element name should be the same as the model name. Once this change was made, the model produced the proper results.


mTorbin:
The exact error message I was receiving was:

WARNING: Encountered "data" in payload, but no model was found for model name "datum" (resolved model name using super-rentals@serializer:-rest:.modelNameFromPayloadKey("data"))

locks:
Thank you for contributing. However, his is a duplicate of https://github.com/emberjs/guides/pull/1118.

You probably didn't update your project dependencies (check Ember and Ember Data in `bower.json`, the version should match the guide), so your application is using Ember Data 1.13 which uses the RESTAdapter by default. The tutorial uses the JSONAPIAdapter format. We're fixing the workflow of the tutorial to avoid this problem.

PR #1148, opened 2016-01-07
issue #853 document passing args, currying, return values, target, value

toddjordan:
I expanded the example a bit to take the mystery out of the confirm() function, so that I could show some of the more advanced closure action features.

PR #1181, opened 2016-01-21
add section on action threading

toddjordan:
Meant to go with PR #1148 (expands on examples and refers back).
Part of the the fix for issue #853

### DIFF
--- a/source/components/triggering-changes-with-actions.md
+++ b/source/components/triggering-changes-with-actions.md
@@ -96,14 +96,37 @@ the action from the component:
 
 ```app/components/button-with-confirmation.js
 export default Ember.Component.extend({
-  tagName: 'button',
-  click() {
-    if (confirm(this.get('text'))) {
+
+  actions: {
+    launchConfirmDialog() {
+      this.set('confirmShown', true);
+    },
+
+    submitConfirm() {
       // trigger action on parent component
+      this.set('confirmShown', false);
+    },
+
+    cancelConfirm() {
+      this.set('confirmShown', false);
     }
   }
 });
 ```
+
+The component template will have a button and a div that shows the confirmation dialog
+based on the value of `confirmShown`.
+
+```app/templates/components/button-with-confirmation.hbs
+<button {{action "launchConfirmDialog"}}>{{text}}</button>
+{{#if confirmShown}}
+  <div class="confirm-dialog">
+    <button class="confirm-submit" {{action "submitConfirm"}}>OK</button>
+    <button class="confirm-cancel" {{action "cancelConfirm"}}>Cancel</button>
+  </div>
+{{/if}}
+```
+
 
 ## Passing the Action to the Component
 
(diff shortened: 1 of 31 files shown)


source/components/triggering-changes-with-actions.md
```diff
@@ -156,3 +156,56 @@ a function instead.
 Actions in components allow you to
 decouple an event happening from how it's handled, leading to modular,
 more reusable components.
+
+## Calling Actions Across Component Layers
```
locks: "Across" might be a bit ambiguous as people can read it as siblings instead of nested.

source/components/triggering-changes-with-actions.md
```diff
@@ -156,3 +156,56 @@ a function instead.
...
+
+When your components go multiple layers deep, its common to need to handle an action several layers up the tree.
```
locks: I think it's important to differentiate between being nested in the context/template sense, and using block components and being nested in the markup sense. This section not being applicable to the latter.

source/components/triggering-changes-with-actions.md
```diff
@@ -156,3 +156,56 @@ a function instead.
...
+Using action helpers, its possible to make top level actions available at the bottom layers of your component tree
```
locks: "the action helper"
"it's possible"

I'd suggest rephrasing "top level actions". "top level" is a bit fuzzy for more beginner readers.

source/components/triggering-changes-with-actions.md
```diff
@@ -156,3 +156,56 @@ a function instead.
...
+{{button-with-confirmation onConfirm=(action deleteCurrentUser)
+  text="Click OK to delete your account."}}
+```
+
+Note that `deleteCurrentUser` is not in quotes as was the case [previously](#toc_passing-the-action-to-the-component)
+when the action was local to `user-profile`.
+This is because when you pass an action helper a string, it will attempt to invoke that function from the component's
+local `actions` object.
+
+When you pass an action helper a function reference as in this case, it will invoke the function from the component's
```
locks: Confusing wording. "When you pass a function reference to the `action` helper" maybe?

locks:
@toddjordan I like your explanation of `(action 'bloop')` vs `(action bloop)`, but I wonder if the flow could be improved a little. Can you do another pass and see if any idea pops into your head?
Otherwise, I'll merge when the other comments are addressed.

Appreciated!

toddjordan:
@locks thanks for the comments.  I'll give the wording there some more work.  

Also, do plan on doing the master conflict resolution or should I just include it with my next push?

source/components/triggering-changes-with-actions.md
toddjordan: `Calling Actions Up Multiple Component Levels`?

source/components/triggering-changes-with-actions.md
toddjordan: `When your components go multiple template layers deep, its common to need to handle an action several layers up the tree.`
Difficult to express the difference.  "template layers" hopefully will give the impression of multiple templates in use.  Hopefully the reader also gets it from the example...

locks:
@homu r+

toddjordan:
closing this PR, because I've hosed it with my merge/squash madness

PR #1182, opened 2016-01-21
Fixes SSL mixed-content issue

### LINKED ISSUE #1146: Run loop example has a mixed content error from the iframe src
https://guides.emberjs.com/v2.2.0/applications/run-loop/#toc_an-example-of-the-internals

In source/applications/run-loop.md, the iframe points to a http url:
`http://emberjs.com.s3.amazonaws.com/run-loop-guide/index.html`
But the site is served on a https connection, so the console reports a mixed content error and doesn't load the url. Testing in Firefox and Chrome.

I would submit a PR for https, but the amazon certificate isn't valid for the domain. Chrome reports it as "NET::ERR_CERT_COMMON_NAME_INVALID", with the explanation "This server could not prove that it is emberjs.com.s3.amazonaws.com; its security certificate is from *.s3.amazonaws.com."

Doing some research I found the original app came from `https://github.com/machty/ember-run-loop-visual`, but the version in that repo doesn't look like the one being served through s3.

I believe this is an artifact from when the guides moved from emberjs/website to emberjs/guides. I think a solution to this would be to serve the example through Firebase, like the rest of the guides. But this is not something I can submit a PR for.

acorncom:
Fixes #1146

### DIFF
--- a/source/applications/run-loop.md
+++ b/source/applications/run-loop.md
@@ -154,7 +154,7 @@ Working with this API directly is not common in most Ember apps,
 but understanding this example will help you to understand the run-loops algorithm,
 which will make you a better Ember developer.
 
-<iframe src="http://emberjs.com.s3.amazonaws.com/run-loop-guide/index.html" width="678" height="410" style="border:1px solid rgb(170, 170, 170);margin-bottom:1.5em;"></iframe>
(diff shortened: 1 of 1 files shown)


locks:
@homu r+

PR #1186, opened 2016-01-21
Provide a subsection describing nested helpers

grapho:


source/templates/handlebars-basics.md
```diff
@@ -52,6 +52,44 @@ controllers and components.
...
-Helpers bring a minimum of logic into Ember templating. Ember ships with several
-built-in helpers, which are explained in the following guides, and also allows
-you to [write your own helpers](../writing-helpers/).
+Ember gives you the ability to [write your own helpers](../writing-helpers/), to bring a minimum of logic into Ember templating.
+
+For example, let's say you would like the ability to add a few numbers together, without needing to define a computed property everywhere you would like to do so.
+
+```javascript
+// helpers/sum.js
+export default Ember.Helper.helper(function(params) {
```
locks: This snippet helper does not conform to the ones generated by CLI, can you tweak it please?

source/templates/handlebars-basics.md
```diff
@@ -52,6 +52,44 @@ controllers and components.
...
 ### Helpers
 
-Helpers bring a minimum of logic into Ember templating. Ember ships with several
-built-in helpers, which are explained in the following guides, and also allows
-you to [write your own helpers](../writing-helpers/).
+Ember gives you the ability to [write your own helpers](../writing-helpers/), to bring a minimum of logic into Ember templating.
+
+For example, let's say you would like the ability to add a few numbers together, without needing to define a computed property everywhere you would like to do so.
+
+```javascript
```
locks: this should be

    ```app/helpers/sum.js

source/templates/handlebars-basics.md
```diff
@@ -52,6 +52,44 @@ controllers and components.
...
+// helpers/sum.js
```
locks: remove this, see above

source/templates/handlebars-basics.md
```diff
@@ -52,6 +52,44 @@ controllers and components.
...
+
+Ember ships with several built-in helpers, which you will learn more about in the following guides.
+
+#### Nested Helpers
+
+Helpers have the ability to be nested within other helper invocations and also component invocations.
+
+This gives you the flexibility to compute a value _before_ it is passed in as an argument or an attribute of another.
+
+It is not possible to nest curly braces `{{}}` inside of a helper already using the curly braces, so a "sub-expression" can be created using parentheses `()`:
```
locks: it's not possible to nest mustaches at all

locks: don't use `"sub-expression"`, doesn't explain anything and might confuse beginners.
consider a wording like "since it is not possible to nest curly braces `{{}}`, the correct way to use nested helpers is using parenthesis `()`:"

source/templates/handlebars-basics.md
```diff
@@ -52,6 +52,44 @@ controllers and components.
...
+
+```html
+{{sum (multiply 2 4) 2}}
+```
+
+In this example, we are using a helper to multiply `2` and `4` _before_ passing the value into `{{sum`.
+
+Thus, the output of these combined helpers is `10`.
+
+As you move forward with these template guides, keep in mind that most of Ember's built-in helpers can accept _nested_ helper values as arguments.
```
locks: the thing to focus on here is not that helpers can accept nested helpers (helpers can be used wherever a value can), is that most helpers can be used in the nested form. think inline and block forms of `{{if`.

locks:
@homu r+

PR #1189, opened 2016-01-23
Fixes output example for index route creation

tcanabarro:
Small fix of the cli output creation messages in Tutorial: Routes and Templates.


locks:
@tcanabarro appreciated!

@homu r+

Issue #1193, opened 2016-01-25
Filtering Deprecations example uses underscores


gandalfar:
This example:
https://guides.emberjs.com/v2.3.0/configuring-ember/handling-deprecations/#toc_filtering-deprecations

Doesn't work out of the box, because it uses `_` from underscores.js. It might be better to either mention this dependancy or rewrite it in a way that doesn't need underscores.

acorncom:
@toddjordan Here's that underscore issue that was getting discussed on the Slack `#documentation` channel

acorncom:
Closed by the PR above, thanks for the feedback!

PR #1195, opened 2016-01-26
favor using typeof instead of 3rd party lib

toddjordan:
fix for issue #1193


acorncom:
@homu r+

acorncom:
Thanks @toddjordan

PR #1198, opened 2016-01-28
Fix inconsistency between guide and server output

sotojuan:
The example says the default server displays "Welcome to Ember.js" but actually it's "Welcome to Ember". Just a little inconsistency that was bugging me!


acorncom:
@sotojuan Thanks, was going to merge this but the Travis test failed. I think it's probably a Travis problem, could you re-run the test? Thanks!

sotojuan:
Sure, but I am not entirely sure how to do that! I don't see a button on Travis to re-run the test.

acorncom:
@sotojuan You'll need to change your commit in some way so that Travis notices that it's a new commit and re-runs. Probably the easiest way would be to rebase against `master`, as there have been new changes which would change your Git commit sha.  Let me know if you need help with that.

sotojuan:
Looks like it worked (amended the commit with no change!).

acorncom:
@homu r+

acorncom:
Thanks!

PR #1206, opened 2016-02-01
Correct custom inflector example

mfeckie:
The `Inflector` isn't available on `Ember` at the point which this example is loaded.

This change imports directly from the inflector library and changes to ES6 const syntax.


locks:
Appreciated!

@homu r+

PR #1209, opened 2016-02-02
Remove unnecessary word from Nav Links section

mikeymurph77:
Sentence did not flow. I think the word "to" needs to be omitted.


locks:
@homu r+

PR #1211, opened 2016-02-02
Grammar, wording, punctuation, example, and terminology fixes

zaccharles:
I previously submitted two PRs for fixing guides. I hindsight I should have waited until I finished reading everything. Oh well, I finished reading everything now, and have fixed up the bits and pieces I found.

These are mostly missing commas, missing words, incorrect grammar, or wrong terminology. In a couple of cases, I changed code examples to prevent horizontal scrolling.

### DIFF
--- a/source/testing/testing-components.md
+++ b/source/testing/testing-components.md
@@ -3,7 +3,7 @@
 Components can be tested with integration tests using the `moduleForComponent` helper.
 
 Let's assume we have a component with a `style` property that is updated
-whenever the value for its `name` property changes. The `style` attribute of the
+whenever the value of the `name` property changes. The `style` attribute of the
 component is bound to its `style` property.
 
 > You can follow along by generating your own component with `ember generate
@@ -186,7 +186,7 @@ test('should trigger external action on form submit', function(assert) {
 ### Stubbing Services
 
 In cases where components have dependencies on Ember services, it is possible to stub these
-dependencies for integration tests. Stub Ember services by using the built-in register
+dependencies for integration tests. You stub Ember services by using the built-in `register()`
 function to register your stub service in place of the default.
 
 Imagine you have the following component that uses a location service to display the city
@@ -199,7 +199,7 @@ and country of your current location:
 export default Ember.Component.extend({
   locationService: Ember.inject.service('location-service'),
 
-  //when the coordinates change, call the location service to evaluate what the city and country would be
+  // when the coordinates change, call the location service to get the current city and country
   city: Ember.computed('locationService.currentLocation', function () {
     return this.get('locationService').getCurrentCity();
   }),
--- a/source/testing/testing-controllers.md
+++ b/source/testing/testing-controllers.md
@@ -6,8 +6,8 @@ of the ember-qunit framework.
 
 ### Testing Controller Actions
 
-Here we have a controller `PostsController` with some computed properties and an
-action `setProps`.
+Here we have a controller `PostsController` with two properties, a method that
+sets one of those properties, and an action named `setProps`.
 
 > You can follow along by generating your own controller with `ember generate
 > controller posts`.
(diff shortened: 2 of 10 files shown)


source/testing/testing-controllers.md
```diff
@@ -30,8 +30,8 @@ export default Ember.Controller.extend({
 });
 ```
 
-`setProps` sets a property on the controller and also calls a method. In our 
-generated test, ember-cli already uses the `moduleFor` helper to setup a test 
+The `setProps` action directly sets one property, and calls the method to set the other.
```
locks: calls a method

source/testing/testing-controllers.md
```diff
@@ -30,8 +30,8 @@ export default Ember.Controller.extend({
...
+In our generated test, ember-cli has already used the `moduleFor` helper to set up a test
```
locks: can you revert this one to the original please?

source/testing/testing-components.md
```diff
@@ -258,7 +258,7 @@ is being returned from the service is reflected in the component output.
 ```tests/integration/components/location-indicator-test.js
 test('should reveal current location', function(assert) {
   this.render(hbs`{{location-indicator}}`);
-  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA');
+  assert.equal(this.$().text().trim(), 'Location: New York, USA');
```
locks: these changes feel arbitrary, maybe I'm just missing the context. can you explain them pliz?

locks:
@zaccharles left you a couple notes. can you address them and squishy squash the commits?

source/testing/testing-components.md
zaccharles: Yeah, I can see how it would feel arbitrary. Basically, it was just done to reduce line length so the examples didn't scroll. I felt like it conveyed the same message, but was easier to grok if you can see all of the line at once, without scrolling or wrapping.

locks: could you split it into a different PR? if it's too much hassle I'll merge as is

source/testing/testing-controllers.md
zaccharles: I originally had "a" here instead of "the" (as in the original text). I changed it at the last second because I felt it flowed better in the context of reading the page top to bottom.

What I mean is, before the code snippet of the controller, it says:
> Here we have a controller `PostsController` with two properties, a method that
> sets one of those properties, and an action named `setProps`.

So, assuming you're reading that first, you already know about _the_ method. Saying _a_ makes it sound like maybe it's another method again?

That's my logic anyway. Happy to change it to "a" if you still disagree :)

zaccharles: I believe that in the context of the rest of the page, it makes more sense to say _the_, rather than _a_.

Before the code block, it says:
> Here we have a controller `PostsController` with two properties, a method that
> sets one of those properties, and an action named `setProps`.

Therefore, if you're reading top to bottom, I thought it was clearer that we're talking about that method, _the_ method. Rather than another method, _a_ method.

If you disagree, I'm happy to change it :)

source/testing/testing-controllers.md
zaccharles: I reverted the sentence starting with "In our generated", except change the word "setup" to "set up", because that's just plain wrong. Is that the part you meant to be reverted? It's hard to tell with the way GitHub is showing the diff.

source/testing/testing-components.md
zaccharles: I've reverted this change and will look at putting it in another PR after this one.

zaccharles:
Squished. Also actioned and replied to comments. Will sleep now and do more fixes in response to your reply(ies) tomorrow :)

locks:
:shipit: 

@homu r+

PR #1212, opened 2016-02-03
Fix API doc links in "Redirecting" section

chnn:
When browsing the guides [here](https://guides.emberjs.com/v2.3.0/routing/redirection/), I noticed a few of the links to the API docs are incorrect. These changes fix that.


acorncom:
@homu r+

acorncom:
Thanks!

PR #1217, opened 2016-02-04
Tweaks TOC

locks:
Moves Ember CLI installation instructions to tutorial so they're more
contextualized.
Renames Glossary Glossary to Glossary Web Development in preparation to
addition of Ember.js glossary and more.


locks:
@homu r+

PR #1220, opened 2016-02-04
[WIP] Glossary for Ember.js specific terms

locks:


pixelhandler:
@locks what do you think about adding a "vocab" section to the guides? Perhaps like -> http://kylecoberly.github.io/ember-vocab/

acorncom:
@locks I'm going to close this for now, we've got it tracked in Trello

PR #1227, opened 2016-02-05
Fix reference to a controller from code example

mehulkar:


source/routing/query-params.md
```diff
@@ -82,7 +82,7 @@ The `link-to` helper supports specifying query params using the
 ```
 
 In the above examples, `direction` is presumably a query param property
-on the `controller:post`, but it could also refer to a `direction` property
+on the `controller:posts`, but it could also refer to a `direction` property
```
locks: can you change it to prose instead? "on the `posts` controller"

mehulkar:
done!

locks:
Thank you :dancers: 

@homu r+

PR #1228, opened 2016-02-05
Reference using a serializer not a adapter in JSONSerializer section

andrelandry:
Customizing Serializers referenced creating a adapter instead of a serializer.


acorncom:
@homu r+

acorncom:
Thanks, good catch!

PR #1230, opened 2016-02-06
Update hbs-helper.md

admande:
Updated text to reflect actual boilerplate language ("rentalPropertyType" not "rentalGroup"). Still functional as it was, but perhaps this is more clear for people following along.


locks:
:+1: 

@homu r+

PR #1234, opened 2016-02-09
Corrected the request URL in Sideloaded Data.

kpavankumar:
Corrected the URL mentioned in Sideloaded Data example from `people/1` to `articles/1`


locks:
Thank you!

@homu r+

PR #1235, opened 2016-02-09
Fix duplicate reference-style links

fsoumare:
Links on the page https://guides.emberjs.com/v2.3.0/object-model/classes-and-instances/ point to wrong sections of the api doc.


acorncom:
Thanks!

@homu r+

PR #1238, opened 2016-02-09
change rental-unit to rentalUnit

ChangJoo-Park:
It need to be change for consistency.

Fixes #1238


locks:
@ChangJoo-Park consistency with what? They were consistent with themselves already. Since the variable is only used in one place in the loop I don't see the benefit of the change.

ChangJoo-Park:
@locks It is confuse thing. 
 source/tutorial/autocomplete-component.md Use rental-unit variable for iteration.
But 
https://guides.emberjs.com/v2.3.0/tutorial/simple-component/ Use rentalUnit

Both of them uses different name. I think it is making confuse to beginner like me.

See issue #1238

locks:
I see, thank you!

@homu r+

PR #1240, opened 2016-02-10
Fix data provided by Mirage in tutorial

voutilad:
I noticed when walking through the 2.3.0 tutorial I was getting errors in Chrome when trying to wire up the mirage data and even when fixing the key problem it didn't render properly. Looks like it could just be out of date documentation based on changes in Ember and Mirage.

Note: The version of mirage that `ember install ember-cli-mirage` installed per the tutorial as of time of this PR is `1.13.15`


locks:
Hi, this is a duplicate of #1117. See my comment at #1117 (comment) for the full explanation.

PR #1248, opened 2016-02-11
`cursor:pointer` shouldn't hit `disabled` elements

barneycarroll:
`cursor: pointer` is a mixed bag: on the one hand it can enable otherwise unavailable click handler functionality on certain elements; on the other, it forces a pointer cursor regardless of other factors that might disable this functionality.

Given the blanket `[data-ember-action]` selector, `<button>`s, for instance, will retain a pointer cursor even when they are `disabled`.

This qualification guards against that.

### DIFF
--- a/source/localizable/templates/actions.md
+++ b/source/localizable/templates/actions.md
@@ -154,7 +154,7 @@ elements, a CSS workaround exists to make them clickable, `cursor: pointer`.
 For example:
 
 ```css
-[data-ember-action] {
+[data-ember-action]:not(:disabled) {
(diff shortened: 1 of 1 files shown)


barneycarroll:
[Here's a twiddle illustrating the problem](https://ember-twiddle.com/b5ebfdc20f155f8df2fd?numColumns=1&openFiles=my-component.template.hbs%2C).

acorncom:
@homu r+

acorncom:
Thanks!

PR #1255, opened 2016-02-12
Tweaked the edit link on the guides pages

### LINKED ISSUE #1253: The "edit pencil" icon for suggesting a guides change is broken
post file moves due to our translations work.  We'll want to fix this soon ... <sigh>

acorncom:
Fixes #1253

### DIFF
--- a/source/layout.erb
+++ b/source/layout.erb
@@ -38,7 +38,7 @@
       <article class="chapter">
         <h1>
           <%= page_title %>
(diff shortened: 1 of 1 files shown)


acorncom:
@locks Seems like we might want to back port this change, thoughts?

locks:
@acorncom unnecessary, since we don't fix old versions.

@homu r+

Issue #1251, opened 2016-02-11
[Meta] Shift to Ember as a CMS


acorncom:
This is a placeholder issue for making a shift from Middleman to Ember (with a docs API backend).  It will grow over time, but these are the first thoughts on what we need to tackle:

### Phase 1
- [ ] build an initial Ember app to handle versioned apis (and multiple languages) (see #1257 for todos)
    - [ ] use pre-rendered HTML built using Middleman as part of builds
    - [ ] work through how to handle putting guides in our API system (using our build system)
- [ ] Fastboot integration (handles our SEO issues)
    - [ ] Caching (how to handle things so that pages don't need to render markdown on the server side for each page

### Phase 2
- [ ] shift away from Middleman completely (looks like we mainly need to handle highlight.js, which can be used either in Node or in-browser

### Phase 3
- [ ] offline docs using PouchDB ?

Other ideas / comments / suggestions about things needed here are more than welcome!

rtablada:
In a similar vain, I've built todomvc.embergrep.com as an ember app that is serving markdown generated content.

Right now it is just a set of MD strings that are shoved in to an ENV variable but I'm hoping to actually make this a bit more stable and have an Ember addon that allows for a list of markdown files, publishing over to the final build, and lists of what posts are available as a static JSON feed.

locks:
We could phase this.

* Phase 1, render HTML from `guides.github.com`
* Phase 2, render Markdown from `guides`
* etc

acorncom:
Agreed on phasing. For phase 2, I just checked and highlight.js will run in Node, so we'll be able to do that on the Fastboot side when we get to that point.

rondale-sc:
Had a thought this morning on this.  Would having this as an Ember app without fastboot cause it to be less visible?

acorncom:
Yes, Fastboot is a requirement for this to be feasible. But per our docs meeting yesterday (still need to send in the minutes), Fastboot will be available for this project

rondale-sc:
Ooh, neat!  What API system is currently the plan?

acorncom:
Haven't figured that out yet ;-)  I've wondered about using Firebase (we already have an account), but haven't thought through it.  We'd need it to work with both Fastboot and in-browser, want to take a look at that?  I spent some time last night working up a prototype and it's working pretty nicely so far.  Will link in the PR in a minute

acorncom:
[...]

PR #1257, opened 2016-02-12
[CMS] WIP: Guides viewer

acorncom:
Helps with #1251

To do:
- [x] shift basic page design over for initial spike
- [x] prototype getting data from builds using Express
- [ ] fix remaining issues visible at https://github.com/acorncom/guides/issues
- [ ] integrate with the API setup


locks:
@acorncom can we close for now and revisit in the future?

acorncom:
@locks Yep.

For future reference (if anyone looks at this), we decided that this Ember spike was too much effort for not enough return.  It was fun to experiment with, but wasn't worth pursuing at this time ...

PR #1264, opened 2016-02-16

mikzat:
There was an issue with calling `clear()` on the result of `this.set(...)`. It's possible the intention here was to use `this.get(...).clear()`, but for me getting rid of `clear()` resulted in better behavior in the example.

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -109,7 +109,7 @@ export default Ember.Controller.extend({
           this.set('filteredList',result);
         });
       } else {
-        this.set('filteredList');
+        this.get('filteredList').clear();
       }
     },
     search(param) {
@@ -118,7 +118,7 @@ export default Ember.Controller.extend({
           this.set('model',result);
         });
       } else {
-        this.set('model');
+        this.get('model').clear();
       }
     }
   }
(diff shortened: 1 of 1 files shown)


locks:
Setting the property to undefined is too much of a semantic change from the intended one. Replacing the array vs emptying it.
I would prefer the typo to be corrected to `.get` instead

mikzat:
Done.

locks:
Thank you for spotting this! :) I'm trying to fix the test suite before merging, trying to do it ASAP.

fedekau:
Although this corrects the error of using `clear()` on a undefined object you will still get `Uncaught Error: The result of a server query (on super-rentals@model:rental:) is immutable.` 

May be using the `.set(...)` is not a bad idea, correct me if i am wrong :)

locks:
Hm, we'll review the whole guide, thanks for bringing it to our attention @fedekau.

amamaenko:
Having started learning Ember a couple of days ago, I also struggled with understanding this part of tutorial. Eventually, I went for a dumb solution, replacing "clear" with another request getting full data from the server:
```
search(param) {
            if (param != "") {
                this.store.query('rental', {city:param}).then((result) => {
                   this.set('model', result) 
                });
            } else {
                 this.store.query('rental', {}).then((result) => {
                   this.set('model', result)
                 });
            }
        }
```

toddjordan:
Created #1359 to address the issue here, as well is the immutable message.

toddjordan:
Thanks for the contribution!  Closing this PR in favor of broader fixes around autocomplete in #1359

PR #1270, opened 2016-02-19
Ignores translated files during build

locks:
Translated files have a problem with code blocks. They lose the closing
backticks of code blocks, rendering the project unbuildable.
Ignoring those files is a necessary temporary measure while we sort the
actual problem and continue working on the translations themselves.


locks:
@homu r+

PR #1272, opened 2016-02-19
store.find is private should be store.findRecord

frunjik:
store.find is private should be store.findRecord


locks:
Appreciated!

@homu r+

PR #1274, opened 2016-02-19
Ember CLI is what is installed when you do `npm install -g ember-cli@2.3`

spastorino:


locks:
Appreciated!

@homu r+

PR #1275, opened 2016-02-19
Remove the open a directory with the editor part

spastorino:
Mentioning the editor part here is confusing given that the guide is trying to explain how to start the development server in console


rwjblue:
I agree

locks:
@homu r+

PR #1276, opened 2016-02-20
Update repositories.md

kentaro0919:
fix 
**Ember Data** - A data persistence library for Ember.js.
to
**Ember Data** - A data persistence library for Ember.


locks:
@kentaro0919 I'll keep this open in case you find more of them to fix, if that's ok with you.

rwjblue:
`packages/metamorph` is gone

rwjblue:
We should also list tildeio/glimmer, tildeio/htmlbars, https://github.com/krisselden/dag-map, maybe others?

kentaro0919:
@locks It is fine by me. I will add them if I find

locks:
@homu r+

PR #1225, opened 2016-02-04
removed diff style

GavinJoyce:
fixes: https://github.com/emberjs/ember.js/issues/12913
slack chat: https://embercommunity.slack.com/archives/documentation/p1454619920000419

**before:**

image

**after:**

image

PR #1277, opened 2016-02-20
Diff style is not being displayed

spastorino:
I've edited the title to describe the real issue.

### DIFF
--- a/source/localizable/getting-started/quick-start.md
+++ b/source/localizable/getting-started/quick-start.md
@@ -210,14 +210,7 @@ curly braces (`{{component}}`). We're going to tell our component:
 2. What array of people to use, via the `people` attribute. We'll
    provide this route's `model` as the list of people.
 
-```app/templates/scientists.hbs{-1,-2,-3,-4,-5,-6,-7,+8}
-<h2>List of Scientists</h2>
-
-<ul>
-  {{#each model as |scientist|}}
-    <li>{{scientist}}</li>
-  {{/each}}
-</ul>
+```app/templates/scientists.hbs
 {{people-list title="List of Scientists" people=model}}
 ```
 


source/localizable/getting-started/quick-start.md
```diff
@@ -211,13 +211,6 @@ curly braces (`{{component}}`). We're going to tell our component:
    provide this route's `model` as the list of people.
 
 ```app/templates/scientists.hbs{-1,-2,-3,-4,-5,-6,-7,+8}
```
locks: this markup is now incorrect, since you removed the mentioned lines

spastorino: Ahh now I see what the intention is ... it tries to show that lines from 1 to 7 should be removed and line 8 is added. But that's not showing properly here https://guides.emberjs.com/v2.3.0/getting-started/quick-start/

acorncom: Hmm, looks like I merged this PR to fix that issue a while back (https://github.com/emberjs/guides/pull/1225) but instead of us fixing the display of the CSS issues we just eliminated `added` / `removed` styles.  <sigh>

spastorino: @acorncom yep, I ended tracking exactly that commit which is wrong as you figured.
I'd let others with better styling skills do the job but just doing something easy, what about ...

diff

spastorino: This https://gist.github.com/spastorino/810b47638232affdcca8 is what makes styles look like 👆

spastorino: Also, all this started happening since https://github.com/emberjs/guides/commit/d2b2c49f4828fcfec3bfcf0d025755d53a27976e, before the background was white so those green/red applied correctly. See for instance how was the style before the mentioned commit and how looks in the blog http://emberjs.com/blog/2016/01/12/ember-data-2-3-released.html

locks: @acorncom that was on purpose

locks:
@spastorino remove both the code and the diff notation please

spastorino:
@locks done, but if that's the solution the same procedure should be applied in a lot of different places like https://github.com/spastorino/guides/blob/02c6b634ef694555637701b60ddea7aaa640bc18/source/localizable/getting-started/quick-start.md#L130 and https://github.com/spastorino/guides/blob/02c6b634ef694555637701b60ddea7aaa640bc18/source/localizable/getting-started/quick-start.md#L157 to mention only two more places in just this file.

Probably there are more places where this should be done.

locks:
Hm. We get to get a designer to fix the styling then.

acorncom:
I think it's just used on the Getting Started page, but would be nice to have it styled.  Let me see what I can come up with ...

acorncom:
Just resolved this on #1283, will close this for now.  @spastorino thanks for the help in tracking down what was going wrong here ...

PR #1278, opened 2016-02-20
There's no need to reload due to live reload server

spastorino:


acorncom:
Thanks, agreed! :-)
@homu r+

PR #1279, opened 2016-02-20
Update ember-cli.md with `const` instead of `var`

timcmiller:
Teaching myself Ember and wanted to update the example with `const` instead of `var`. Added a quick description on `const` in case the reader was unfamiliar with ES2015.


locks:
Thank you for contributing!

@homu r+

PR #1282, opened 2016-02-22
Example code using component contains old (duplicate) code

frunjik:
The text says: 
Replace all our old code with our new componentized version.

But the example code contains the old (original) code and the component, which would show up duplicated.


locks:
Duplicate of #1277.
Thanks for reporting!

PR #1283, opened 2016-02-23
Putting back our added/removed styling

acorncom:
Supersedes #1277 and fully deals with issues initial fixed in #1225


acorncom:
screen shot 2016-02-22 at 6 21 07 pm

@locks @spastorino

locks:
So ugly :C

toddjordan:
@homu r+

PR #1286, opened 2016-02-24
Fix reference-style links

fsoumare:
Links on the following pages point to wrong sections of the api doc:
https://guides.emberjs.com/v2.3.0/object-model/bindings/
https://guides.emberjs.com/v2.3.0/object-model/enumerables/
https://guides.emberjs.com/v2.3.0/object-model/observers/
https://guides.emberjs.com/v2.3.0/routing/asynchronous-routing/

### DIFF
--- a/source/localizable/object-model/bindings.md
+++ b/source/localizable/object-model/bindings.md
@@ -3,10 +3,8 @@ bindings in Ember.js can be used with any object. That said, bindings are most
 often used within the Ember framework itself, and for most problems Ember app
 developers face, computed properties are the appropriate solution.
 
-The easiest way to create a two-way binding is to use a [`computed.alias()`][1],
+The easiest way to create a two-way binding is to use a [`computed.alias()`](http://emberjs.com/api/classes/Ember.computed.html#method_alias),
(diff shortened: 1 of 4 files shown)


locks:
@fsoumare can you do me a favour and inline the links instead? I've found that using footnote-style links has been leading to problems with the references getting out of sync or wrong.

fsoumare:
Done. I think we should use inline links throughout the docs. I can give a hand if need be.

acorncom:
@fsoumare if you wouldn't mind helping run through to adjust those, it would be much appreciated. I tried the other approach because it looked cleaner but it's been a headache :-/

For this PR, could you squash your commits? Once you have we'll get it merged :-)

fsoumare:
Looks like I didn't squash my commits properly. Shall I create a new PR with a single commit?

locks:
@fsoumare it's probably the more straightforward way, so you don't have to much too much with git. go ahead :)

fsoumare:
How about now? :)

acorncom:
@homu r+

Looks good!

PR #1288, opened 2016-02-24

milworm:
fixed an issue: When you click on a "search"-button and the field is empty nothing happens, but seems it's expected to show all items.


locks:
@milworm thank you for your contribution, but we already have a meta-discussion going at https://github.com/emberjs/guides/pull/1264. We might modify the tutorial so the component works slightly differently.

milworm:
ok, no problem :)

PR #1290, opened 2016-02-26
Clarify action closure walkthrough

averydev:
I found this to be a bit confusing and it took a good bit of trial and error before the distinction between `(action passedAction)` and `(action 'localAction')` made sense. Hopefully this clears things up for other users.

### DIFF
--- a/source/localizable/components/triggering-changes-with-actions.md
+++ b/source/localizable/components/triggering-changes-with-actions.md
@@ -134,8 +134,8 @@ Now we need to make it so that the `onConfirm()` event in the
 `button-with-confirmation()` component triggers the
 `userDidDeleteAccount()` action in the `user-profile` component.
 One important thing to know about actions is that they're functions
-you can call, like any other method on your component.
-So they can be passed from one component to another like this:
+you can call like any other method on your component.
+They can be passed from one component to another like this:
 
 ```app/components/user-profile.hbs
 {{button-with-confirmation text="Click here to delete your account." onConfirm=(action "userDidDeleteAccount")}}
@@ -186,8 +186,7 @@ That makes it easy to remember how to add an action to a component. It's
 like passing an attribute, but you use the `action` helper to pass
 a function instead.
 
-Actions in components allow you to
-decouple an event happening from how it's handled, leading to modular,
+Actions in components allow you to decouple an event happening from how it's handled, leading to modular,
 more reusable components.
 
 ## Handling Action Completion
@@ -386,7 +385,7 @@ export default Ember.Component.extend({
 ## Calling Actions Up Multiple Component Layers
 
 When your components go multiple template layers deep, it is common to need to handle an action several layers up the tree. 
-Using the action helper, it is possible to make actions which are defined in parent components available to their children through templates alone without adding JavaScript code to child components in between.
+Using the action helper, parent components can pass actions to child components through templates alone without adding JavaScript code to those child components.
 
 For example, say we want move account deletion from the `user-profile` component to its parent `system-preferences-editor` to be handled.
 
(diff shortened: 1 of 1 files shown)


source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -426,3 +403,22 @@ export default Ember.Component.extend({
   }
 });
 ```
+Then our `system-preferences-editor` template passes its local `deleteUser` action into the `user-profile` as that
+component's `deleteCurrentUser` property.
+
+```app/templates/components/system-preferences-editor.hbs
+{{user-profile deleteCurrentUser=(action 'deleteUser' login.currentUser.id)}}
+```
+The action `deleteUser` is in quotes, since `system-preferences-editor` is where the action is defined now. Quotes simply indicate that the action should be looked for in that comoponent's `actions:{}` object, rather than in the `attrs` that have been passed from the parent.
```
toddjordan: We aren't generally documenting the `attrs` object, rather using the action helper with an unquoted argument will call any function on the component's scope.  So I think it would be more accurate to say something like "... rather than in the scope of the component"

toddjordan:
@averydev thanks for the contribution.  I think what you propose is clearer, but I did have one comment.

locks:
See https://locks.svbtle.com/to-attrs-or-not-to-attrs for more context on `this.attrs`.

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -385,35 +385,12 @@ export default Ember.Component.extend({
 
 ## Calling Actions Up Multiple Component Layers
 
-When your components go multiple template layers deep, its common to need to handle an action several layers up the tree.
-Using the action helper, it is possible to make actions defined in parent components available at the bottom layers of
-your component tree without adding JavaScript code to the components in between.
+When your components go multiple template layers deep, its common to need to handle an action several layers up the tree.
```
locks: `it's`

source/localizable/components/triggering-changes-with-actions.md
locks: See https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md for guidelines on how to break up the text.

locks: `actions`, instead of `actions:{}`

locks: "Quotes simply indicate" -> "Quotes indicate"

teddyzeenny: s/comoponent/component

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -426,3 +403,22 @@ export default Ember.Component.extend({
...
+The action `deleteUser` is in quotes, since `system-preferences-editor` is where the action is defined now. Quotes simply indicate that the action should be looked for in that comoponent's `actions:{}` object, rather than in the `attrs` that have been passed from the parent.
+
+In our `user-profile.hbs` template we change our action to call `deleteCurrentUser` as passed above.
+
+```app/templates/components/user-profile.hbs
+{{button-with-confirmation onConfirm=(action deleteCurrentUser)
+  text="Click OK to delete your account."}}
+```
+
+Note that `deleteCurrentUser` is no longer in quotes here as opposed to [previously](#toc_passing-the-action-to-the-component) . Quotes are used to initially pass the action down the component tree, but at every subsequent level you are instead passing the actual function reference (without quotes) in the action helper.
```
acorncom: Remove the space before the period in this text:

> ) . Quotes are used

locks:
Hi @averydev, can you give this PR some TLC?

averydev:
Sorry for the delay fixing this up.  Hopefully this improves things. I haven't contributed to the guides before, so it's a... learning experience  :)

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -385,35 +384,12 @@ export default Ember.Component.extend({
...
-component's local `deleteCurrentUser` property.
-
-```app/templates/components/system-preferences-editor.hbs
-{{user-profile deleteCurrentUser=(action 'deleteUser' login.currentUser.id)}}
-```
-
-Now when you confirm deletion, the action goes straight to the `system-preferences-editor` to handle.
+For example, say we want move account deletion from the `user-profile` component to its parent `system-preferences-editor` to be handled.
 
+First we would move the `deleteUser` action from `user-profile.js` to the actions object on `system-preferences-editor.`
```
toddjordan: bring the . out of the quote here.  Also, give a line break before and after the code sample.  I think this is causing spellcheck in our ci to be confused.

toddjordan:
@averydev once we get the CI passing we'll be good to go!

locks:
Manually merged. Thanks everyone :)

PR #1292, opened 2016-02-26
Minor markup change

andrewvida:
After creating a new application and opening `application.hbs`, "Welcome to Ember" is actually inside of a `h2` instead of a `h1`.


acorncom:
@homu r+  Thanks!

It's about to go totally haywire once we put in https://github.com/locks/ember-welcome-page ;-)

PR #1295, opened 2016-03-01
[WIP] Remove negative margin causing glitch on hover

miguelmota:
Bug
sidebar-glitch

### DIFF
--- a/source/stylesheets/components/_sidebar.scss
+++ b/source/stylesheets/components/_sidebar.scss
@@ -111,7 +111,7 @@ ol.toc-level-1 {
 
   li {
     border-left: 3px solid transparent;
-    transition: border-width $duration, margin-right $duration;
+    transition: border-width $duration;
 
     &.selected,
     &:hover {
@@ -120,7 +120,6 @@ ol.toc-level-1 {
 
     &:hover {
       border-left-width: 6px;
-      margin-right: -3px;
     }
   }
 


acorncom:
I'm pretty sure I remember that being added deliberately ;-)  See https://github.com/emberjs/guides/pull/1179 for details.  Any suggestions on how to fix that more permanently?  Seems like we might be stuck with it being a tad glitchy somewhere ...

locks:
@ryanlabouve can you give this a review?

kpfefferle:
(I made the previous PR #1179)

The addition of the negative margin fixes the unselected case, but I see why it's breaking the selected case. I will spin up an adjustment that can handle both cases. The key thing is that we don't want the item width to change in either case. The proposal to remove the negative margin returns us to both cases potentially suffering from this problem.

kpfefferle:
I've been playing around with this for a bit, and the behavior on that element is even glitchier in my browser:

old

This seems to be happening because we are animating both the border-width on the left (from 3px to 6px) and the margin on the right (from 0 to -3px to offset the increased border-width). Depending on how the browser handles the animation of these two properties, the width of the element (and thus the text wrapping) varies during animation and causes this glitchy behavior.

I'm going to keep investigating how we might resolve this.

ryanlabouve:
@locks sure thing!

I'm wondering if we might use a psuedo element and animate the width on that instead of using margins at all. That way the text should not be affected at all.

@kpfefferle let me know if you'd like to take run at it! I'm happy to work on this.

kpfefferle:
@ryanlabouve If you have time, feel free. I'm not sure when I'd be able to loop back around and look at it.

acorncom:
Marking this WIP for now

acorncom:
I'm going to close this for now due to lack of activity, but happy to open it again if folks want to work on this more ...

PR #1298, opened 2016-03-02

iambenjohnson:
The second code snippet for the contact.hbs file is inconsistent with the first. The latter does not have the email address as a link.


locks:
Thanks for your contribution!

@homu r+

PR #1304, opened 2016-03-04
Fixed typo in Services guide

greyhwndz:
In the Services section of Application Concerns - Accessing Services, there seems to be a typo wherein instead of referring to a component named _**cart-contents**_ it refers to _**cart-contacts**_...


locks:
@homu r+

PR #1309, opened 2016-03-06
Update index.md

kentaro0919:


locks:
@kentaro0919 thank you for these

@homu r+

PR #1316, opened 2016-03-08
Update index.md

sheriffderek:
I think that "comprehensive" sounds much more positive than "exhaustive."


acorncom:
@homu r+ yep, agreed, thanks!

PR #1319, opened 2016-03-09
Correct route name in documentation

caugusty:
The route is favoritePosts and not posts


caugusty:
Corrected route name in documentation

locks:
Thank you :)

@homu r+

PR #1322, opened 2016-03-10
[Proposal] WIP: Flip "Quick Start" to "index.md" for "Getting Started"

rtablada:
Right now the url https://guides.emberjs.com/v2.4.0/getting-started/ leads to "Installing Ember".

I think this is a bit confusing since the logical first step within getting started (and the ordering in navigation) leads to the "Quick Start".

The reason this is occurring is that the "Installing Ember" markdown file is `getting-started/index.md` and middleman is setting this as the root of the `getting-started` URL.

I think this will clear up some confusion as to why the first thing you go to isn't the root? http://quick.as/9Q6QSObz0
Also, I personally as a maintainer went into `getting-started` knowing the first thing is the quick start, instinctively open an `index` file before other named files `</gripe>`.

This may (mostly speculation) also improve indexing of the "Quick Start" with a shorter URL?

So the steps to accomplish this would mean:

* Change `index.md` and translations to `installing.md`
* Change `quick-start.md` and translations to `index.md`
* Update `pages.yml`

## Blocking

* Wait on translation freeze for "Getting Started" section
* Cleaner Git Diff

Since the files are renamed back into `index.md` Git sees this as one huge swath of removes and inserts rather than moves.
This could be :dragon:s when it comes to merge conflicts.

Another thing we could do (although it leaves a 404) root is to leave `quick-start.md` and just rename `index.md` to `installing.md` and call it a day?


acorncom:
I'm inclined to close this PR for now to clean up our PR list to focus on actionable PRs.  @rtablada are you interested in continuing to pursue this?  I don't mind seeing it happen, but would also like to not be breaking links to old pages.  So we'd need to be updating the `firebase.json` file in https://github.com/emberjs/guides.emberjs.com and doing other changes to ensure this goes through.

For the level of work it will take to make it happen well, I'm not sure it's worth it, despite it being cleaner and more accurate.  ;-)  Thoughts?

acorncom:
Closing this for now, happy to re-open if someone wants to champion it ...

PR #1324, opened 2016-03-13
Update fontello to include new google plus icon

ryanlabouve:
In https://github.com/emberjs/website/pull/2496 we talked about updating the google plus icon on guides.

Here's what the new icon looks like:
ember_js_-_guides_and_tutorials 2


locks:
@homu r+

PR #1325, opened 2016-03-14

twify93:
Fix simple type and missing attribute in the code for index.js autocomplete

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -102,7 +102,6 @@ Now, define your new controller like so:
 ```app/controllers/index.js
 export default Ember.Controller.extend({
   filteredList: null,
-  model: null,
   actions: {
     autoComplete(param) {
       if(param !== "") {
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -28,7 +28,7 @@ It contains an [`{{input}}`](../../templates/input-helpers) helper, that
 renders as a text field that the user can type in to look for properties 
 in a given city. The `value` property of the `input` will be bound to the 
 `filter` property in our component. The `key-up` property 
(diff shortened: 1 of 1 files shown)


source/localizable/tutorial/autocomplete-component.md
```diff
@@ -102,6 +102,7 @@ Now, define your new controller like so:
 ```app/controllers/index.js
 export default Ember.Controller.extend({
   filteredList: null,
+  model: null,
```
locks: Can you amend the commit to remove this line? We generally don't declare the model property explicitly for routable controllers.

locks:
Thanks for the contribution! I left a small note inline, I'll merge once you address it :)

source/localizable/tutorial/autocomplete-component.md
twify93: But running the script without this line, makes the code throwing error when the code reaches this line:
this.set('model').clear(); on line 21
The error thrown:
> index.js:25 Uncaught TypeError: Cannot read property 'clear' of undefined

The error occures when you search for empty string.

Clone my repo and test it:
[https://github.com/twify93/emberjs-rentals](url)

locks: The error is related with [another problem](https://github.com/emberjs/guides/pull/1264), we're currently reviewing the whole guide.

twify93: Ok, so I am removing this line and will fix it with .set :)

acorncom:
@homu r+  Thanks!

toddjordan:
@homu r+

toddjordan:
@homu retry

toddjordan:
@twify93 the tests on this timed out for some reason causing the build to fail.  Since then there have been other merges, so if you could rebase with the latest we can try the merge again.

twify93:
OK. I will review the documentation and make additional changes where they are needed.

PR #1326, opened 2016-03-15
Modified to match the HTML specified in templates

morhook:
The guide specifies a different HTML (in the screenshot there's only a `<ul><li>`). Replaced with real output

PR #1328, opened 2016-03-16
Update code snippet to reflect screenshot

andreavaghi:
Based on the screenshot for this step of the guide, i've updated the code snippet.
You could in alternative change the screenshot, but it's quite long and i've preferred to suggest the change in the code


locks:
Thanks for your contribution, but this issue is already being addressed in another PR (https://github.com/emberjs/guides/pull/1326) that updates the screenshot instead.

PR #1329, opened 2016-03-16
Replaced second observer

morhook:
According to my testing, looks like the `processFullName` was called three-times (on modification of `firstname`, `lastName` and by `Ember.run.once(this, 'processFullName')`), not as expected by the surrounding text. 

Maybe author meant something like what I propose?


toddjordan:
You're right, thanks for the catch!

toddjordan:
@homu r+

morhook:
No problem! Thanks @toddjordan for looking into this one

PR #1332, opened 2016-03-17
Replace usages of testem.json with testem.js

Aevin1387:
The ember 2.4 guides still have references to testem.json, even though this file was replaced with testem.js. This replaces the usages of testem.json with testem.js.


locks:
Thanks for spotting this :)

@homu r+

PR #1336, opened 2016-03-18
Multi-dependency callout

ctataryn:
Calling out the fact your initializer ordering may depend on **multiple** initializers

### DIFF
--- a/source/localizable/applications/initializers.md
+++ b/source/localizable/applications/initializers.md
@@ -89,5 +89,17 @@ export default {
 };
 ```
 
+```app/initializers/asset-init.js
+export function initialize(application) {
+  // ... your code ...
+};
+
+export default {
+  name: 'assetInit',
+  after: ['configReader', 'websocketInit'],
+  initialize: initialize
+};
+``
+
 Note that ordering only applies to initializers of the same type (i.e. application or application instance).
 Application initializers will always run before application instance initializers.


stefanpenner:
sweet

source/localizable/applications/initializers.md
```diff
@@ -89,5 +89,17 @@ export default {
...
+export function initialize(application) {
+  // ... your code ...
+};
+
+export default {
+  name: 'assetInit',
+  after: ['configReader', 'websocketInit'],
+  initialize: initialize
+};
+``
```
acorncom: Instead of adding another code block, let's just tweak the existing code block that uses `after`.  That should be clear enough.

And thanks!

acorncom:
@ctataryn Would you have a chance to update this?  Would love to get it merged ...

acorncom:
@ctataryn Any change you can make the change here?

ctataryn:
Oh boy, sorry @acorncom I missed this notification the first time around. Let me take a look

ctataryn:
@acorncom I think it's appropriate to have the code blocks as they stand because they form a cohesive example (i.e. two initializers, and then a third that is dependent on those other two). Do you agree?

locks:
Let me just note that this explained in the API documentation, which we should probably be linking to.

ctataryn:
@locks where abouts?

locks:
Hm, good question, I wrote it back in 1.x days before the initializer/instance initializer split.

locks:
@homu r+

locks:
Thanks!

PR #1342, opened 2016-03-24
Typo: Use get instead of set

AlexEdgcomb:
Disclaimer: I'm new to Ember.

In the line `this.set('filteredList').clear();`, `this.set('filteredList')` returns `undefined`, so `clear()` crashes.

Similarly, I've updated `this.set('model').clear();`.


locks:
Thank you for your contribution @AlexEdgcomb but this is a duplicate of https://github.com/emberjs/guides/pull/1264.
We are currently reviewing the guide to improve the logic and fix this problem.

PR #1343, opened 2016-03-24
adding clarification about removing the ((outlet}} handlbars

specialorange:


toddjordan:
Thanks @specialorange ! Instead of the sidenote, can you be more direct and say something like:

`For our `about` template, we'll replace the default `{{outlet}}` helper with some HTML that has a bit of information about the site.`

acorncom:
I'm going to close this for now, due to a lack of response.  @specialorange if you'd still like to get this in, let us know and we can re-open this once you've put up the requested update ...

PR #1344, opened 2016-03-24
Fix example to not use leaky state

nucleartide:


source/localizable/components/the-component-lifecycle.md
```diff
@@ -67,7 +67,12 @@ you can use `didUpdateAttrs` to clear any error state that was built up from edi
 import Ember from 'ember';
 
 export default Ember.Component.extend({
-  errors: [],
+  errors: null,
```
krisselden: this isn't needed in addition to the one in `init`

nucleartide: oh ok, just thought it would be good to list all variables like in a C# or Java class. will remove :)

locks:
@homu r+

PR #1346, opened 2016-03-28
Discuss resetNamespace option in router

tsmacdonald:
See also https://github.com/emberjs/ember.js/issues/11251

The full documentation is found in-source [here](https://github.com/emberjs/ember.js/blob/1ce6fbf352c5c8810709e9fbbb662f9ac02ad37f/packages/ember-routing/lib/system/router.js#L1008-L1014)

### DIFF
--- a/source/localizable/routing/defining-your-routes.md
+++ b/source/localizable/routing/defining-your-routes.md
@@ -84,7 +84,18 @@ the `{{outlet}}` of the `posts` template.
 A nested route's names includes the names of its ancestors.
 If you want to transition to a route (either
 via `transitionTo` or `{{#link-to}}`), make sure to use the full route
-name (`posts.new`, not `new`).
+name (`posts.new`, not `new`). If you don't want to use the full route
+name (e.g., because a resource can be accessed in multiple locations), use the
+`resetNamespace` option. This will not change the structure of the URL, it
+merely shortens the route name.
+
+```app/router.js
+Router.map(function() {
+  this.route('posts', function() {
+    this.route('post', { path: '/:post_id', resetNamespace: true});
+  });
+});
+```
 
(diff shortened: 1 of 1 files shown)


source/localizable/routing/defining-your-routes.md
```diff
@@ -84,7 +84,18 @@ the `{{outlet}}` of the `posts` template.
 A nested route's names includes the names of its ancestors.
 If you want to transition to a route (either
 via `transitionTo` or `{{#link-to}}`), make sure to use the full route
-name (`posts.new`, not `new`).
+name (`posts.new`, not `new`). If you don't want to use the full route
+name (e.g., because a resource can be accessed in multiple locations), use the
```
rwjblue: This is exactly why this section was removed. You CANNOT use the same route name for two routes. The only thing that restNamespace gives you is a shorter name.

tsmacdonald: Should I change it to something like `If you don't want to use that full name, you can refer to it just by the name of the child with the resetNamespace option (but note that this does not change the URL structure).`?

Or if you really think that `resetNamespace` should be used rarely, then it should be removed from the final example on emberjs.com.

locks: Hm, I would personally prefer it were removed, as I don't want to encourage it's use because the current router<->file system mapping starts breaking down. I'll bring it up for discussion.

locks:
I am closing this issue due to the factual errors pointed out above, as well as thinking that `resetNamespace` is an advanced, and possibly unnecessary. It should be unnecessary for most applications, and there are other ways, like a wrapping componnet.
Thanks for your contribution @tsmacdonald!

Issue #1318, opened 2016-03-08
Incorrect example in writing helpers


stocksr:
in source/localizable/templates/writing-helpers.md line 263 there is meant to be an example of a class based helper format, for the is-authenticated helper instead there is a copy of a previous example format-currency.

acorncom:
Confirmed, things are currently worded a bit confusingly.  Looks like that's due to the insertion of the `is-authenticated` helper example **above** the `format-currency` instead of after it.  If we shift the example that was introduced in this https://github.com/emberjs/guides/commit/9e392679485d7cb412e7c62f706bdc873281210f to after the `format-currency` helper (and then introduce the aspect of using services) this should get much clearer ...

PR #1347, opened 2016-03-28
Clarify class-based helper format documentation.

cantsin:
Fixes https://github.com/emberjs/guides/issues/1318

Also-by: Dig <dig.vargas@locusenergy.com>

### DIFF
--- a/source/localizable/templates/writing-helpers.md
+++ b/source/localizable/templates/writing-helpers.md
@@ -239,26 +239,8 @@ should export a subclass of [`Ember.Helper`][1]. Helper classes must contain a
 into the class-based helper.  Once added, you can call the service's methods or
 access its properties from within the `compute()` method.
 
-To exemplify, let's make a helper utilizing an authentication service that
-welcomes users by their name if they're logged in:
-
-```app/helpers/is-authenticated.js
-export default Ember.Helper.extend({
-  authentication: Ember.inject.service(),
-  compute() {
-    let authentication = this.get('authentication');
-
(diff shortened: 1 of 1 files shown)


source/localizable/templates/writing-helpers.md
```diff
@@ -239,26 +239,9 @@ should export a subclass of [`Ember.Helper`][1]. Helper classes must contain a
...
-    } else {
-      return 'Not logged in';
-    }
-  }
-});
-```
-
-In fact, we could also refactor the above stateless helper into a class-based
-helper by making the function into a `compute` method on the class:
+In fact, we could also refactor the above `format-currency` helper
```
mixonic: How about:

> An an exercise, here is the above `format-concurrency` helper refactored into a class-based helper:

source/localizable/templates/writing-helpers.md
```diff
@@ -274,9 +257,27 @@ export default Ember.Helper.extend({
 });
 ```
 
-This is exactly equivalent to the example above. You can think of the
-function version as a shorthand for the longer class form if it does not
-require any state.
+This is exactly equivalent to the `format-currency` example above. You
+can think of the function version as a shorthand for the longer class
+form if it does not require any state.
```
mixonic: ~~state~~. "if the helper does not require dependency injection."

cantsin:
@mixonic updated, thank you for the suggestions!

acorncom:
@homu r+  Thanks, @cantsin!

Issue #1340, opened 2016-03-23
Property brace expansion in computed properties


adam-knights:
I only found out about property brace expansion by luck. The only official link to it that I can see is in the 1.4 blog post (http://emberjs.com/blog/2014/02/12/ember-1-4-0-and-ember-1-5-0-beta-released.html#toc_property-brace-expansion).

Suggest that a section is added to computed properties in the guide to show this handy shorthand syntax.

acorncom:
Sounds good to me.  @adam-knights if you'd like to send in a PR, we'd be happy to take it :smile:

PR #1348, opened 2016-03-28
Add brace expansion to computed properties

codeNgJon:
Fixes https://github.com/emberjs/guides/issues/1340
Takes [description of property brace expansion](http://emberjs.com/blog/2014/02/12/ember-1-4-0-and-ember-1-5-0-beta-released.html#toc_property-brace-expansion) and adds to the guides.

screen shot 2016-03-28 at 12 29 47 pm

### DIFF
--- a/source/localizable/object-model/computed-properties.md
+++ b/source/localizable/object-model/computed-properties.md
@@ -31,6 +31,20 @@ This declares the function to be a computed property, and the arguments tell Emb
 
 Whenever you access the `fullName` property, this function gets called, and it returns the value of the function, which simply calls `firstName` + `lastName`.
 
+When you want to depend on a property which belongs to an object, you can setup multiple dependent keys by using brace expansion:
+
+```javascript
+var obj = Ember.Object.extend({
+  baz: {foo: 'BLAMMO', bar: 'BLAZORZ'},
+
+  something: Ember.computed('baz.{foo, bar}') {
+    return this.get('baz.foo') + ' ' + this.get('baz.bar');
+  }
+```
+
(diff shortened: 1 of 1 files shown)


source/localizable/object-model/computed-properties.md
```diff
@@ -31,6 +31,20 @@ This declares the function to be a computed property, and the arguments tell Emb
...
 
+When you want to depend on a property which belongs to an object, you can setup multiple dependent keys by using brace expansion:
+
+```javascript
+var obj = Ember.Object.extend({
+  baz: {foo: 'BLAMMO', bar: 'BLAZORZ'},
+
+  something: function(){
+    return this.get('baz.foo') + ' ' + this.get('baz.bar');
+  }.property('baz.{foo,bar}')
```
acorncom: @codeNgJon Looks great!  One thing we've been shifting toward is an updated syntax that would look more like this:

```
something: Ember.computed('baz.{foo,bar}') {
    return this.get('baz.foo') + ' ' + this.get('baz.bar');
}
```

Would you be able to test that that works (possibly on Ember Twiddle) and update this PR?

acorncom:
Also, we tend to squash commits, so once you've got things worked out nicely, would you do a force push so we only have one commit here?  Thanks!

source/localizable/object-model/computed-properties.md
codeNgJon: Sounds good. I'll give that a try. Thankss

codeNgJon:
Thanks @acorncom. Added the fiddle link to the description up top.

locks:
@homu r+

PR #1350, opened 2016-03-29
Fix missing header for "View all registered templates"

AkeemMcLennon:
This header was present in in the documentation for 1.10 but was somehow lost along the way.


locks:
@AkeemMcLennon thank you!
@homu r+

PR #1355, opened 2016-04-04
Add sidenote that there is a computed.filterBy macro

alexspeller:
There is a computed.filterBy macro that does the same thing as the more verbose example, add a note explaining this.

### DIFF
--- a/source/localizable/object-model/computed-properties-and-aggregate-data.md
+++ b/source/localizable/object-model/computed-properties-and-aggregate-data.md
@@ -27,7 +27,23 @@ and fire observers when any of the following events occurs:
 3. An item is removed from the `todos` array.
 4. The `todos` property of the component is changed to a different array.
 
-In the example above, the `remaining` count is `1`:
+Ember also provides a computed property macro
+[`computed.filterBy`](http://emberjs.com/api/classes/Ember.computed.html#method_filterBy),
+which is a shorter way of expressing the above computed property:
+
+```app/components/todos.js
+export default Ember.Component.extend({
+  todos: [
+    Ember.Object.create({ isDone: true }),
+    Ember.Object.create({ isDone: false }),
+    Ember.Object.create({ isDone: true })
+  ],
+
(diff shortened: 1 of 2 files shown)


toddjordan:
build failing because `sidenote` is failing spellcheck.  Added it to exceptions in pr #1360

source/localizable/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -27,7 +27,22 @@ and fire observers when any of the following events occurs:
 3. An item is removed from the `todos` array.
 4. The `todos` property of the component is changed to a different array.
 
-In the example above, the `remaining` count is `1`:
+Sidenote: ember provides the built-in macro `computed.filterBy`, which is syntax
```
locks: Ember should be capitalized

locks: can you rephrase "built-in macro" and "syntax sugar"? at least while we don't have a good glossary with linkage. "provides a built-in computed property short hand" or something along those lines.

source/localizable/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -27,7 +27,22 @@ and fire observers when any of the following events occurs:
...
+    Ember.Object.create({ isDone: true }),
+    Ember.Object.create({ isDone: false }),
+    Ember.Object.create({ isDone: true })
+  ],
+
+  remaining: Ember.computed.filterBy('todos', 'isDone', false)
+});
+```
+
+In both of the the examples above, the `remaining` count is `1`:
```
locks: repeated `the`

alexspeller:
@locks it was bugging me that this was the only mention of macros so I added a brief section to the computed property guide. I am calling them computed property macros still because that's their name, but each time I've expanded on that for people who don't know what a macro is already (although now they're introduced in an early section this might be redundant)

locks:
@homu r+

PR #1356, opened 2016-04-05
Update ember-data.md

locks:


source/localizable/tutorial/ember-data.md
```diff
@@ -123,5 +123,7 @@ export default Ember.Route.extend({
 ```
 
 When we call `this.store.findAll('rental')`, Ember Data will make a GET request to `/rentals`.
+You can read more about Ember Data in the [Models section](https://guides.emberjs.com/v2.4.0/models/).
```
rwjblue: Should this be a relative path instead of absolute?

locks: Oops, relative!

locks:
@homu r+

Issue #1236, opened 2016-02-09
[Tutorial] Autocomplete after delete input


ChangJoo-Park:
I am following ember.js 2.3.0 guide.

now I read and code autocomplete method.
https://guides.emberjs.com/v2.3.0/tutorial/autocomplete-component/

In this page, **app/controller/index.js** make undefined error when empty input string.

I don't want to leave it.

So.

is it right code to solve?
I find Ember.isEmpty().

```app/controllers/index.js
export default Ember.Controller.extend({
  filteredList: null,
  actions: {
    autoComplete(param) {
      if(param !== "") {
        this.store.query('rental', {city: param}).then((result) => {
          this.set('filteredList',result);
        });
      }
      else {
        let filteredList = this.get('filteredList');
        if(Ember.isEmpty(filteredList)) {
          return;
        }
        console.log();
        this.set('filteredList').clear();
      }
    },
    search(param) {
      if(param !== "") {
        this.store.query('rental', {city: param}).then((result) => {
          this.set('model',result);
        });
      }
      else {
        let model = this.get('model');
        if(Ember.isEmpty(model)){
          return;
        }
        this.set('model').clear();
      }
    }
  }
});

```
I want to know someone who have elegant solution.

**add 2016-02-09**
I solved problem but I don't trust my code is ember.js way.

```javascript
export default Ember.Controller.extend({
  filteredList: null,
  actions: {
    autoComplete(param) {
      if(param !== "") {
        this.store.query('rental', {city: param}).then((result) => {
          this.set('filteredList',result);
        });
      }
      else {
        this.set('filteredList', null);
      }
    },
    search(param) {
      if(param !== "") {
        this.store.query('rental', {city: param}).then((result) => {
          this.set('model',result);
        });
      }
      else {
        this.set('model', this.store.findAll('rental'));
      }
    }
  }
});
```

acorncom:
@toddjordan Could you take a look at this?

estani:
2.4 tutorial still has this issue.

`this.set('filteredList').clear();` makes no sense to me and I think it requires some explanation 
(calling a set method with no value and expecting it to return something upon which `clear()` may be called is everything but self describing, furthermore it doesn't seem to work at least in this case).

locks:
More discussion on this https://github.com/emberjs/guides/pull/1264

vtambourine:
[...]

PR #1359, opened 2016-04-06
Fix autocomplete logic to avoid throwing errors

toddjordan:
This is to fix issues surrounding autocomplete that have prompted some issues and comments
See 
#1236 
#1264

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -99,27 +99,23 @@ ember g controller index
 Now, define your new controller like so:
 
 ```app/controllers/index.js
+import Ember from 'ember';
+
 export default Ember.Controller.extend({
   filteredList: null,
   actions: {
     autoComplete(param) {
       if (param !== '') {
-        this.store.query('rental', { city: param }).then((result) => {
-          this.set('filteredList', result);
-        });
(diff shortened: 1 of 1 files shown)


toddjordan:
I merged/rebased another pr that addresses the issues, so this one just makes a few minor updates now.

source/localizable/tutorial/autocomplete-component.md
```diff
@@ -144,11 +140,11 @@ We also define a `search` action here that is passed in to the component,
 For these actions to work, we need to modify the Mirage `config.js` file 
 to look like this, so that it can respond to our queries.
 
-```app/mirage/config.js
+```mirage/config.js
 export default function() {
   this.get('/rentals', function(db, request) {
     let rentals = [{
-        type: 'rentals',
```
locks: type is pluralized in JSONAPI, iirc

toddjordan: I know this works from the super rentals repo. I'll try plural and look in the spec to see

locks: I know it's either for ED, and json:api doesn't specifically say it has to be plural, but all their [examples have it that way](http://jsonapi.org/format/#document-resource-objects). It's just a matter of consistency.

toddjordan:
@locks good to go?

locks:
:shipit: 

@homu r+

PR #1363, opened 2016-04-07
Makes TOC links Easier to Click

rtablada:
Currently the TOC links are position absolute and only appear when the parent is hovered.
While this is fine, the issue is that the link is hard to trigger showing up and quite easy to miss and hover away from.

This PR fixes this by making the surrounding header element have padding on the left under the anchor tag, so that the full height of the element is available.


acorncom:
Nice, thanks!  @homu r+

PR #1366, opened 2016-04-10
Update unit-testing-basics.md

daltonamitchell:
Spotted a small typo. Given the sentence context, it looks like there was a missing word.

Also, it feels like the word *subject* should be wrapped with backticks for consistency, since we're referring to a function name, similar to *moduleFor*.


locks:
Well spotted, thanks for the contribution :)

@homu r+

PR #1368, opened 2016-04-13
Fix typo in computed properties doc

James1x0:


locks:
Thanks for spotting it!

@homu r+

PR #1369, opened 2016-04-13
Improve code highlighting typography

chriskrycho:
- Don't let user agent stylesheet `pre` override fonts set in `.highlight`
- Don't use ligatures in code samples.

### Before
Note the "fi" and "ff" ligatures in "first-name" and "Jeff Atwood" respectively, and the small type size.

before

### After
The text is sized properly, and without ligatures.

after


locks:
Thanks for the improvement!

@homu r+

chriskrycho:
My pleasure! 👍🏼

PR #1370, opened 2016-04-14
Clarifies `Controller` use

trek:
This PR comes out of a conversation a the last core team face to face about better messaging Controllers, whose death has been widely exaggerated.


locks:
:heart: thanks for getting to it

@homu r+

gabrielgrant:
@trek Do you mind explaining a little about why you removed `dependencies-between-controllers.md` with this commit?

Is there a different recommended method for accessing a parent controller's model?

(I'm wondering if it may have just been an accident, since the translated versions are still there?)

locks:
Seems like CrowdIn doesn't handle this specific case, thanks for pointing it out @gabrielgrant. It was removed as a service should generally be used instead.
Injecting controllers is fraught with edge cases, especially if the injected controller isn't associated with a parent route. At least with services you have to address the service not being initialized directly.

gabrielgrant:
Ok, good to know it was intentional. And yes, I was a bit surprised there was no real distinction in the inter-controller-dependency page between accessing a parent controller (fairly common, and seems reasonable) vs injecting any other controller randomly (almost certainly a bad idea).

For the specific case of accessing the parent controller's model, is the recommendation to still use `Ember.inject.controller()` or to just re-fetch the model in the child route's `model()` hook? (Creating a service for that seems like overkill, no?)

locks:
> (Creating a service for that seems like overkill, no?)

If you know what you're doing, it should be fine ;P And anyone else that comes along afterwards and reads the code.

PR #1371, opened 2016-04-14
Disable ligatures inside code tag

chriskrycho:
Similar to #1369, but applied to `<code>` rather than `<pre>` (I didn't realize this issue affected both, as the hierarchy is slightly different).

### Before
Note the spacing issue around the code block caused by the "fl" ligatures in "inflector":

before

### After
Note the non-ligature "fl" in "inflector":

after


locks:
Thanks for your contribution!

@homu r+

PR #1378, opened 2016-04-16
Update repositories.md

kentaro0919:


source/localizable/contributing/repositories.md
```diff
@@ -1,7 +1,7 @@
-Ember is made up of several libraries. If you wish to add a feature or fix a bug please file a pull request against the appropriate repository. Be sure to check the libraries listed below before making changes in the Ember.js repository.
+Ember is made up of several libraries. If you wish to add a feature or fix a bug please file a pull request against the appropriate repository. Be sure to check the libraries listed below before making changes in the Ember repository.
 
 # Main Repositories
-**Ember.js** - The main repository for Ember.
```
locks: I think this is the one place where we would want to make an exception, since the repository is actually named ember.js. What do you think?

acorncom: Agreed

toddjordan:
@kentaro0919 do you mind restoring line 4 back to `**ember.js**` to reflect the actual name of the git repository?  Once that happens the PR should be good to go!

kentaro0919:
It is good to me ;>

locks:
Perfect, thanks :D

@homu r+

PR #1379, opened 2016-04-17
Update quick-start.md

mintymac:
Unless I'm not understanding something, the version should reflect the docs version.


locks:
Thank you for bringing this to our attention @mintymac!
I am only going to close your PR because we need to fix our build process so this doesn't happen again, and it also means that 2.3 and 2.4 probably have the wrong version as well.

mintymac:
np, happy to!

PR #1380, opened 2016-04-17
Update documentation to include route-serializer info

trentmwillis:
As requested in emberjs/ember.js#13016 and discussed in the RFC: [Route Serializers](https://github.com/emberjs/rfcs/blob/master/text/0120-route-serializers.md). Let me know if anything else is needed as I haven't contributed to the guides before.

### DIFF
--- a/source/localizable/routing/defining-your-routes.md
+++ b/source/localizable/routing/defining-your-routes.md
@@ -183,6 +183,35 @@ If the user navigates to `/post/5`, the route will then have the `post_id` of
 `5` to use to load the correct post. In the next section, [Specifying a Route's
 Model](../specifying-a-routes-model), you will learn more about how to load a model.
 
+### Serializing data into dynamic segments
+
+By default, dynamic segments of a route's url will be populated by the properties of the same name on a passed in data model.
+For example, given `postModel={ post_id: 10 }` and the same route map from above, the following:
+
+```templates/application.hbs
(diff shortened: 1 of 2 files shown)


rwjblue:
Awesome, thank you @trentmwillis.  We will likely hold this PR and merge closer to the release version of the feature, but I defer to the @emberjs/guides-managers on that...

locks:
I confirm, it will get merged when we do the review period freeze so we can get updates to the current content in as much as possible. 
Appreciate the work! :)

source/localizable/routing/defining-your-routes.md
```diff
@@ -183,6 +183,26 @@ If the user navigates to `/post/5`, the route will then have the `post_id` of
...
+
+Router.map(function() {
+  this.route('post', {
+    path: '/post/:post_id',
+    serialize: serializePost
+  });
+});
+```
+
+If no `serialize` function is defined, then a straight property lookup on the model will be used for each segment instead.
```
locks: Can you give an example? Or maybe inverse it, first explain how the default works, and then explain how you can customize. I believe you can check the `model()` hook documentation for a similar flow.

trentmwillis: Makes sense. Will update.

trentmwillis:
@locks updated. Let me know what you think.

PR #1382, opened 2016-04-18
Fix ambiguous references to old methods

andrewhavens:
In the guides, on [customizing the adapter](https://guides.emberjs.com/v2.5.0/models/customizing-adapters/#toc_url-conventions), under URL conventions. It references old methods, in a way that might still apply, but could be confusing. I have updated this documentation to reference specific methods which are the current methods you would use to perform these requests.

I'm still pretty new to Ember, so maybe the original documentation was intended.

### DIFF
--- a/source/localizable/models/customizing-adapters.md
+++ b/source/localizable/models/customizing-adapters.md
@@ -93,11 +93,11 @@ JSON API adapter:
     <tr><th>Action</th><th>HTTP Verb</th><th>URL</th></tr>
   </thead>
   <tbody>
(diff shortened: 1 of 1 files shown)


locks:
@andrewhavens they're not methods, they're operations that just happen to sort of map to methods. @bmac can you confirm?

bmac:
That is correct @locks.

locks:
I am closing this issue because it was indeed the intended original documentation. Hm, not sure how to make it clearer.
Feel free to open a new issue if you have an idea how to improve that, thank you!

PR #1383, opened 2016-04-19
Update name of slack room for to #-learning

ryanlabouve:
Received a complaint in Slack #general about not being able to find the `#documentation` room in slack.

(Figured it would be just as easy to open up a PR to fix this as it would be an issue)


locks:
Thanks for your contribution!

@homu r+

Issue #1372, opened 2016-04-15
Add example of using `{{some-component class="some-class"}}`


HeroicEric:
I think it would be useful to show an example of using `{{some-component class="some-class"}}` style of specifying classes for components [Customizing a Component's Element](https://guides.emberjs.com/v2.5.0/components/customizing-a-components-element/) section of the guides.

There are examples for customizing via `classNames` prop inside the component class but not from the template.

I'll have some time to do this tomorrow if this is wanted.

acorncom:
Let's do it

PR #1388, opened 2016-04-21
Add basic component class example

HeroicEric:
Fixes https://github.com/emberjs/guides/issues/1372

This adds an example of setting the class for a component's element
like:

```hbs
{{navigation-bar class="primary"}}
```

### DIFF
--- a/source/localizable/components/customizing-a-components-element.md
+++ b/source/localizable/components/customizing-a-components-element.md
@@ -31,7 +31,14 @@ export default Ember.Component.extend({
 </ul>
 ```
 
-### Customizing Class Names
+### Customizing the Element's Class
+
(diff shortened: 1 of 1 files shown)


source/localizable/components/customizing-a-components-element.md
```diff
@@ -31,7 +31,14 @@ export default Ember.Component.extend({
 </ul>
 ```
 
-### Customizing Class Names
+### Customizing the Element's Class
+
+You can specify the class of a component's element the same way you would for a
```
rwjblue: Minor suggestion:

> You can specify the class of a component's element at invocation time .......

toddjordan:
Looks good, thanks @HeroicEric

toddjordan:
@homu r+

PR #1389, opened 2016-04-21
Change ember version for npm install

aliaporci:
I could be crazy, but I don't think the command-line instructions were updated for Ember 2.5.0.


locks:
Thank you for your contribution @aliaporci but this is something that we have to fix in our build process because it's not the only place were the version is wrong, nor the only version where it's wrong.
See https://github.com/emberjs/guides/pull/1379 for more context.

PR #1396, opened 2016-04-24
Update quick-start.md

kentaro0919:

PR #1397, opened 2016-04-24
Adds a bit of JavaScript to show the correct version on ember-cli install

locks:
Defaults to installing latest (npm install -g ember-cli) if JavaScript
does not run.
Fixes #1396.


locks:
@homu r+

locks:
@homu r+

PR #1399, opened 2016-04-24
Update declarations in classes-and-instances.md

hankfanchiu:
- Replace `var` with `const`.
- Elaborate on `get()` vs. `set()` with examples.

### DIFF
--- a/source/localizable/object-model/classes-and-instances.md
+++ b/source/localizable/object-model/classes-and-instances.md
@@ -11,7 +11,7 @@ To define a new Ember _class_, call the [`extend()`][1] method on
 [2]: http://emberjs.com/api/classes/Ember.Object.html
 
 ```javascript
-Person = Ember.Object.extend({
+const Person = Ember.Object.extend({
   say(thing) {
     alert(thing);
   }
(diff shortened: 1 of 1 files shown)


source/localizable/object-model/classes-and-instances.md
```diff
@@ -212,10 +211,17 @@ and [`set()`][8] accessor methods:
 [8]: http://emberjs.com/api/classes/Ember.Object.html#method_set
 
 ```js
-var person = Person.create();
+Person = Ember.Object.extend({
+  name: 'Robert Jackson'
```
locks: why was this expanded?

locks: poke @hankfanchiu

toddjordan:
In general (mostly for consistency) going forward we want to use let for anything other than native constants.  

What I'd suggest for this PR is to use `const` for class definitions: `const Person = Ember.Object.extend({...});` and native constants `const TIMEOUT_PERIOD = 5000;`

and use `let` for everthing else: `let myPerson = Person.create();`

source/localizable/object-model/classes-and-instances.md
hankfanchiu: @locks In every section, before each instance of `Person` is created, the class is first defined (e.g. [line 135](https://github.com/emberjs/guides/pull/1399/files#diff-ef8f883b306b4c47672331305e5d5da6R135) and [line 181](https://github.com/emberjs/guides/pull/1399/files#diff-ef8f883b306b4c47672331305e5d5da6R181)). To maintain consistency I thought it would make sense to expand this code block as well.

hankfanchiu:
@toddjordan You're right. I'll update my PR.

source/localizable/object-model/classes-and-instances.md
locks: gotcha! I'll re-read the full thing to get a better sense

locks:
Thanks for your contribution!

@homu r+

PR #1400, opened 2016-04-24
you => your and add comma for 'For example'

BenMorganIO:
Noticed this while reading the docs :)


source/localizable/models/customizing-serializers.md
```diff
@@ -129,7 +129,7 @@ import JSONSerializer from 'ember-data/serializers/json';
 export default JSONSerializer.extend({});
 ```
 
-You can also define serializer for a specific model. For example if
+You can also define serializer for a specific model. For example, if
```
locks: a serializer

Issue #1387, opened 2016-04-20
Make names of variables and actions different between the component and the controller.


joetidee:
From reading the tutorial, it was difficullt to discern which name related to either the component or the controller:

[https://guides.emberjs.com/v2.5.0/tutorial/autocomplete-component/](https://guides.emberjs.com/v2.5.0/tutorial/autocomplete-component/)

I had some assitance understanding this on the Slack forum, however, I think it would be advatangeous to newbies like myself if some naming adjustments could be made.

For example, here I understood that the component was being inserted into the template file:

`{{filter-listing filteredList=filteredList autoComplete=(action 'autoComplete') search=(action 'search')}}`

However, what wasn't clear was where the values where coming from because the flow of the explanation in the text was broken up by the text explaingin how to generate the controller file.

What would help, is to do something like this:

`{{filter-listing filteredList=filteredList controllerAutoComplete=(action 'controllerAutoComplete') controllerSearch=(action 'controllerSearch')}}`

> filter-listing.js:

```
export default Ember.Component.extend({
  filter: null,
  filteredList: null,
  actions: {
    componentAutoComplete() {
      this.get('controllerAutoComplete')(this.get('filter'));
    },
    componentSearch() {
      this.get('controllerSearch')(this.get('filter'));
    },
    componentChoose(city) {
      this.set('filter', city);
    }
  }
});
```

> /controllers/index.js:

```
export default Ember.Controller.extend({
  filteredList: null,
  actions: {
    controllerAutoComplete(param) {
      if (param !== '') {
        this.store.query('rental', { city: param }).then((result) => {
          this.set('filteredList', result);
        });
      } else {
        this.set('filteredList', null);
      }
    },
    controllerSearch(param) {
      if (param !== '') {
        this.store.query('rental', { city: param }).then((result) => {
          this.set('model', result);
        });
      } else {
        this.store.findAll('rental').then((result) => {
          this.set('model', result);
        });
      }
    }
  }
});
```

> /templates/components/filter-listing.js:

```
City: {{input value=filter key-up=(action 'componentAutoComplete')}}
<button {{action 'componentSearch'}}>Search</button>
<ul>
  {{#each filteredList as |item|}}
      <li {{action 'componentChoose' item.city}}>{{item.city}}</li>
  {{/each}}
</ul>
```

acorncom:
@joetidee Those sound like good suggestions, would you be able to send in a PR with your changes?

joetidee:
No idea how to do this.

[...]

PR #1401, opened 2016-04-24
Distinguish controller vs. component actions

hankfanchiu:
In response to: https://github.com/emberjs/guides/issues/1387

- Update autocomplete-component.md.
- Add distinction between the `filter-listing` component's actions and the `index` controller's actions,  by prefixing the controller's actions with `controller`- to reduce the chance of any confusion.
- Add safety check for `filter` to be not `null` nor empty string, before querying.
- Set `model` to the query result's `data` property (adhering to the JSON:API)

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -1,43 +1,49 @@
 As they search for a rental, users might also want to narrow their search
- to a specific city. Let's build a component that will let them search for
-  properties within a city, and also suggest cities to them as they type.
+to a specific city. Let's build a component that will let them search for
+properties within a city, and also suggest cities to them as they type.
 
-To begin, let's generate our new component. We'll call this component 
+To begin, let's generate our new component. We'll call this component
 `filter-listing`.
 
 ```shell
 ember g component filter-listing
 ```
-As before, this creates a Handlebars template 
-(`app/templates/components/filter-listing.hbs`) and a JavaScript file 
+
+As before, this creates a Handlebars template
+(`app/templates/components/filter-listing.hbs`) and a JavaScript file
 (`app/components/filter-listing.js`).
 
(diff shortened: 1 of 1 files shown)


locks:
@hankfanchiu could you separate the action renaming from the rest please? It would make it easier to review.

toddjordan:
Hey @hankfanchiu thanks for the pr.

I'm against referring to a controller from a component.  I don't think our filter-listing component should care whether its forwarding its action to a controller, or just a parent controller, or even a route (using the route helper addon).

What I would prefer to see is the filter-listing not even define actions and just forward its actions to its parent.  You can do this by using the action helper with an unquoted argument in the filter-listing template.  That way there is only one thing handling the action, so there is not need to make a distinction for clarity.

source/localizable/tutorial/autocomplete-component.md
```diff
@@ -15,7 +15,8 @@ As before, this creates a Handlebars template
 The Handlebars template looks like this:
 
 ```app/templates/components/filter-listing.hbs
-City: {{input value=filter key-up=(action 'autoComplete')}} 
+City: {{input value=filter key-up=(action 'autoComplete')}}
```
toddjordan: Need to try this but I think you can just call it like this and remove the need for handling the action within the component.

```hbs
{{input value=filter key-up=(action autoComplete filter)}}
```

hankfanchiu: How about this change in the actual app? https://github.com/emberjs/super-rentals/pull/17

If that's good, then I'll update this PR for the guide to reflect the app.

toddjordan:
Thanks @hankfanchiu looks good.  go ahead and update the guides

toddjordan:
Looks great @hankfanchiu, thanks for contributing!

toddjordan:
@homu r+

Issue #1374, opened 2016-04-15
Ember Tutorial


toddjordan:
This is a continuation of what started in issue https://github.com/emberjs/guides/issues/445

Goal: to create guided tutorial, geared towards people new to Ember.  This will be some people's first experience with the techology, so we want to make it show off the basics of creating an application from scratch.  At the end of the tutorial the developer should have an understanding of the basic concepts of ember and ember-cli and be able to create and deploy a basic ember application from scratch.

The tutorial will lead the user through the process of creating a vacation rental app called "Super Rentals".  MVP of the app will be to allow perspective renters to browse available properties.
- [x] ~~Creating a new app~~
- [x] ~~Routes and Templates~~
- [x] ~~Using Addons (CSS addon and mirage)~~
- [x] ~~Writing an Acceptance Test~~
- [x] ~~The Model Hook~~
- [x] ~~A Simple Component~~
- [x] ~~Handlebars Helper~~
- [x] ~~A Complex Component~~
- [x] ~~Service~~
- [x] ~~Nested Routes~~
- [x] ~~Integrate testing into existing tutorials~~
- [ ] Creating Forms
- [x] Links to the super-rentals repo
- [ ] Make super-rentals accessible, add a section?
## Standards for Each Tutorial Page
- Clear language adhering to guides contributing standards: https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md
- Each Tutorial Page should be written in a step-by-step format, guiding the user through how they would build it themselves
- Examples tailored to someone new to ember, showing off basic concepts while not diving into complex approaches
- Testing should be incorporated into each functionality described.
- The super-rentals repo should be updated with working code (and tests): https://github.com/emberjs/super-rentals
## Using addons

Our addon examples will use Mirage and a custom CSS addon we've written to decorate our website.  This section should occur early in the tutorial so as pages are create they are instantly styled.  Also we want to employ Mirage as our datasource.

An alternative is creating an addon for adding styles into the addon (suggested by tom dale)  This would show how to install the addon, plus give the app a nice look and feel without having to include css in the tutorials. 
## Writing an acceptance test

As a good basic intro to acceptance testing, we want to show how we can write a test to exercise the basic initial load of the page and test one of the routes.  We'll do this early, after we've installed mirage, but before creating most of the page content.
## Complex Component

[...]

PR #1403, opened 2016-04-25
Google Maps service tutorial

toddjordan:
Part of issue #1374 

This is our services tutorial page.  features the following
* incorporate testing (plan to do this for all applicable tutorial pages eventually)
* doing some simple caching of maps to show the benefit of using a service
* using a utility to abstract the details of a 3rd party api

I took an outside in approach, starting with the display and going back to the actual interfacing with the 3rd party api.  I also start sections with tests.

### DIFF
--- a/source/localizable/tutorial/service.md
+++ b/source/localizable/tutorial/service.md
@@ -1,0 +1,303 @@
+For Super Rentals, we want to be able to display a map showing where each rental is.  To implement this feature, we will take advantage of several Ember concepts:
+
+1. A component to display a map on each rental listing.
+2. A service to keep a cache of rendered maps to use in different places in the application.
+3. A utility function to create a map from the Google Maps API.
+
+We'll start by displaying the map and work our way back to using the Google Map API.
+
+
+### Display Maps With a Component
+
+We'll start by adding a component that shows the rental's city on a map.
+
+```app/templates/components/rental-listing.hbs{+14}
+<h2>{{rental.title}}</h2>
+<p>Owner: {{rental.owner}}</p>
+<p>Type: {{rental-property-type rental.type}} - {{rental.type}}</p>
+<p>Location: {{rental.city}}</p>
+<p>Number of bedrooms: {{rental.bedrooms}}</p>
+{{#if isImageShowing}}
+  <p>
+    <img alt={{rental.type}} src={{rental.image}} width="500"/>
+  </p>
+  <button {{action "imageHide"}}>Hide image</button>
+{{else}}
+  <button {{action "imageShow"}}>Show image</button>
+{{/if}}
+{{location-map location=rental.city}}
+```
+
+Next, generate the map component using Ember-CLI.
+
+```shell
+ember g component location-map
+```
+
+Running this command generates three files: a component JavaScript file, a template, and a test file.
+To help think through what we want our component to do, we'll implement a test first.
+
+In this case, we plan on having our Google Maps service handle map display.
+Our component's job will be to take the results from the map service (which is a map element) and append it to an element in the component template.
+
+To limit the test to validating just this behavior, we'll take advantage of the registration API to provide a stub maps service.
+A stub stands in place of the real object in your application and simulates its behavior.
+In the stub service, define a method that will fetch the map based on location, called `getMapElement`.
+
+```tests/integration/components/location-map-test.js
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import Ember from 'ember';
+
+let StubMapsService = Ember.Service.extend({
+  getMapElement(location) {
+    this.set('calledWithLocation', location);
+    // We create a div here to simulate our maps service,
+    // which will create and then cache the map element
+    return document.createElement('div');
+  }
+});
+
+moduleForComponent('location-map', 'Integration | Component | location map', {
+  integration: true,
+  beforeEach() {
+    this.register('service:maps', StubMapsService);
+    this.inject.service('maps', { as: 'mapsService' });
+  }
+});
+
+test('should append map element to container element', function(assert) {
+  this.set('myLocation', 'New York');
+  this.render(hbs`{{location-map location=myLocation}}`);
+  assert.equal(this.$('.map-container').children().length, 1, 'the map element should be put onscreen');
+  assert.equal(this.get('mapsService.calledWithLocation'), 'New York', 'a map of New York should be requested');
+});
+```
+
+In the `beforeEach` function that runs before each test, we use the implicit function `this.register` to register our stub service in place of the maps service.
+Registration makes an object available to your Ember application for things like loading components from templates and injecting services in this case.
+
+The call to the function `this.inject.service` injects the service we just registered into the context of the tests, so each test may access it through `this.get('mapsService')`.
+In the example we assert that `calledWithLocation` in our stub is set to the location we passed to the component.
+
+To get the test to pass, add the container element to the component template.
+
+```app/templates/components/location-map.hbs
+<div class="map-container"></div>
+```
+
+Then update the component to append the map output to its inner container element.
+We'll add a maps service injection, and call the `getMapElement` function with the provided location.
+
+We then append the map element we get back from the service by implementing `didInsertElement`, 
+which is a [component lifecycle hook](../../components/the-component-lifecycle/#toc_integrating-with-third-party-libraries-with-code-didinsertelement-code).
+This function gets executed at render time after the component's markup gets inserted into the DOM.
+
+```app/components/location-map.js
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  maps: Ember.inject.service(),
+
+  didInsertElement() {
+    this._super(...arguments);
+    let location = this.get('location');
+    let mapElement = this.get('maps').getMapElement(location)
+    this.$('.map-container').append(mapElement);
+  }
+});
+```
+
+### Fetching Maps With a Service
+
+Now we have a passing component integration test, but no maps showing up when we view our web page.
+To generate the maps, we'll implement the maps service.
+
+Accessing our maps API through a service will give us several benefits
+
+* It is injected with a [service locator pattern](https://en.wikipedia.org/wiki/Service_locator_pattern), meaning it will abstract the maps API from the code that uses it, allowing for easier refactoring and maintenance
+* It is lazy-loaded, meaning it won't be initialized until it is called the first time.
+In some cases this can reduce your app's processor load and memory consumption.
+* It is a singleton, which will allow us cache map data.
+* It follows a lifecycle, meaning we have hooks to execute cleanup code when the service stops, preventing things like memory leaks and unnecessary processing.
+
+Let's get started creating our service by generating it through Ember-CLI, which will create the service file, as well as a unit test for it.
+
+```shell
+ember g service maps
+```
+
+The service will keep a cache of map elements based on location.
+If the map element exists in the cache, the service will return it, otherwise it will create a new one and add it to the cache.
+
+To test our service, we'll want to assert that locations that have been previously loaded are fetched from cache, while new locations are created using the utility.
+
+```tests/unit/services/maps-test.js
+import { moduleFor, test } from 'ember-qunit';
+import Ember from 'ember';
+
+const DUMMY_ELEMENT = {};
+
+let MapUtilStub = Ember.Object.extend({
+  createMap(element, location) {
+    this.assert.ok(element, 'createMap called with element');
+    this.assert.ok(location, 'createMap called with location');
+    return DUMMY_ELEMENT;
+  }
+});
+
+moduleFor('service:maps', 'Unit | Service | maps', {
+  needs: ['util:google-maps']
+});
+
+test('should create a new map if one isnt cached for location', function (assert) {
+  assert.expect(4);
+  let stubMapUtil = MapUtilStub.create({ assert });
+  let mapService = this.subject({ mapUtil: stubMapUtil });
+  let element = mapService.getMapElement('San Francisco');
+  assert.ok(element, 'element exists');
+  assert.equal(element.className, 'map', 'element has class name of map');
+});
+
+test('should use existing map if one is cached for location', function (assert) {
+  assert.expect(1);
+  let stubCachedMaps = Ember.Object.create({
+    sanFrancisco: DUMMY_ELEMENT
+  });
+  let mapService = this.subject({ cachedMaps: stubCachedMaps });
+  let element = mapService.getMapElement('San Francisco');
+  assert.equal(element, DUMMY_ELEMENT, 'element fetched from cache');
+});
+```
+
+Note that the test uses a dummy object as the returned map element.  This can be any object because it is only used to assert that the cache has been accessed.
+Also note that the location has been `camelized` in the cache object, so that it may be used as a key.
+
+Now implement the service as follows.  Note that we check if a map already exists for the given location and use that one, otherwise we call a Google Maps utility to create one.
+We abstract our interaction with the maps API behind an Ember utility so that we can test our service without making network requests to Google.
+
+```app/services/maps.js
+import Ember from 'ember';
+import MapUtil from '../utils/google-maps';
+
+export default Ember.Service.extend({
+
+  init() {
+    if (!this.get('cachedMaps')) {
+      this.set('cachedMaps', Ember.Object.create());
+    }
+    if (!this.get('mapUtil')) {
+      this.set('mapUtil', MapUtil.create());
+    }
+  },
+
+  getMapElement(location) {
+    let camelizedLocation = location.camelize();
+    let element = this.get(`cachedMaps.${camelizedLocation}`);
+    if (!element) {
+      element = this.createMapElement();
+      this.get('mapUtil').createMap(element, location);
+      this.set(`cachedMaps.${camelizedLocation}`, element);
+    }
+    return element;
+  },
+
+  createMapElement() {
+    let element = document.createElement('div');
+    element.className = 'map';
+    element.setAttribute('style', 'width:300px;height:300px;');
+    return element;
+  }
+
+});
+```
+
+### Making Google Maps Available
+
+Before implementing the map utility, we need to make the 3rd party map API available to our Ember app.
+There are several ways to include 3rd party libraries in Ember. See the guides section on [managing dependencies](../../addons-and-dependencies/managing-dependencies/) as a starting point when you need to add one.
+
+Since Google provides its map API as a remote script, we'll use curl to download it into our project's vendor directory.
+
+From your project's root directory, run the following command to put the Google maps script in your projects vendor folder as `gmaps.js`.  
+`Curl` is a UNIX command, so if you are on windows you should take advantage of [Windows bash support](https://msdn.microsoft.com/en-us/commandline/wsl/about), or use an alternate method to download the script into the vendor directory.
+
+```shell
+curl -o vendor/gmaps.js https://maps.googleapis.com/maps/api/js?v=3.22
+```
+
+Once in the vendor directory, the script can be built into the app.
+We just need to tell Ember-CLI to import it using our build file:
+
+```ember-cli-build.js{+22}
+/*jshint node:true*/
+/* global require, module */
+var EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function(defaults) {
+  var app = new EmberApp(defaults, {
+    // Add options here
+  });
+
+  // Use `app.import` to add additional libraries to the generated
+  // output files.
+  //
+  // If you need to use different assets in different
+  // environments, specify an object as the first parameter. That
+  // object's keys should be the environment name and the values
+  // should be the asset to use in that environment.
+  //
+  // If the library that you are including contains AMD or ES6
+  // modules that you would like to import into your application
+  // please specify an object with the list of modules as keys
+  // along with the exports of each module as its value.
+  app.import('vendor/gmaps.js');
+
+  return app.toTree();
+};
+```
+
+### Accessing the Google Maps API
+
+Now that we have the maps API available to the application, we can create our map utility.
+Utility files can be generated using Ember CLI.
+
+```shell
+ember g util google-maps
+```
+
+The CLI `generate util` command will create a utility file and a unit test.
+We'll delete the unit test since we don't want to test Google code.
+Our app needs a single function, `createMapElement`, which makes use of `google.maps.Map` to create our map element, `google.maps.Geocoder` to lookup the coordinates of our location, and `google.maps.Marker` to pin our map based on the resolved location.
+
+```app/utils/google-maps.js
+import Ember from 'ember';
+
+const google = window.google;
+
+export default Ember.Object.extend({
+
+  init() {
+    this.set('geocoder', new google.maps.Geocoder());
+  },
+
+  createMap(element, location) {
+    let map = new google.maps.Map(element, { scrollwheel: false, zoom: 10 });
+    this.pinLocation(location, map);
+    return map;
+  },
+
+  pinLocation(location, map) {
+    this.get('geocoder').geocode({address: location}, (result, status) => {
+      if (status === google.maps.GeocoderStatus.OK) {
+        let location = result[0].geometry.location;
+        let position = { lat: location.lat(), lng: location.lng() };
+        map.setCenter(position);
+        new google.maps.Marker({ position, map, title: location });
+      }
+    });
+  }
+
+});
+
+```
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -32,6 +32,8 @@
       url: "hbs-helper"
     - title: "An Autocomplete Component"
       url: "autocomplete-component"
+    - title: "A Service and Utility"
+      url: "service"
     - title: "Deploying"
       url: "deploying"
 
--- a/data/spelling-exceptions.txt
+++ b/data/spelling-exceptions.txt
@@ -138,6 +138,7 @@ laravel
 li
 lifecycle
 localhost
+locator
 login
 lookup
 lookups
@@ -192,6 +193,7 @@ readme
 readonly
 recordarrays
 refactor
+refactoring
 relayout
 repo
 rerender


source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+
+Now that we have the maps API available to the application, we can create our map utility.
+Utility files can be generated using Ember CLI.
+
+```shell
+ember g util google-maps
+```
+
+The CLI generate util command will create a utility file and a unit test.
+We'll delete the unit test since we don't want to test Google's code.
```
toddjordan: I can add a unit test, but didn't feel like constantly hitting goole maps would be as helpful.  We've already create a unit test with the maps service.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+  // If you need to use different assets in different
+  // environments, specify an object as the first parameter. That
+  // object's keys should be the environment name and the values
+  // should be the asset to use in that environment.
+  //
+  // If the library that you are including contains AMD or ES6
+  // modules that you would like to import into your application
+  // please specify an object with the list of modules as keys
+  // along with the exports of each module as its value.
+  app.import('vendor/gmaps.js');
```
toddjordan: Will utilize the diff color highlighting here to show this line need to be added. My first pass at it was causing an error.  Will look at it some more to figure out why.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+
+Before implmenting the map utility we need to make the 3rd party map API available to our Ember app.
+There are several ways to include 3rd party libraries in Ember. See the guides section on [managing dependencies](../../addons-and-dependencies/managing-dependencies/) as a starting point when you need to add one.
+
+Since Google provides its map API as a remote script, we'll use curl to download it into our project's vendor directory.
+
+From your project's root directory, run the following command to put the Google maps script in your projects vendor folder as `gmaps.js`.
+
+```shell
+curl -o vendor/gmaps.js https://maps.googleapis.com/maps/api/js?v=3.22
```
toddjordan: I know this is unix-centric, but is this ok now that windows has a shell? 😀

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+      this.get('mapUtil').createMap(element, location);
+      this.set(`cachedMaps.${camelizedLocation}`, element);
+    }
+    return element;
+  },
+
+  createMapElement() {
+    let element = document.createElement('div');
+    element.className = 'map';
+    element.setAttribute('style', 'width:300px;height:300px;');
```
toddjordan: Not ok with hardcoding this.  Alternatives are setting a classname here the dev can define in app css, or passing dimensions in thru the component.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+<p>Number of bedrooms: {{rental.bedrooms}}</p>
+{{#if isImageShowing}}
+  <p>
+    <img alt={{rental.type}} src={{rental.image}} width="500"/>
+  </p>
+  <button {{action "imageHide"}}>Hide image</button>
+{{else}}
+  <button {{action "imageShow"}}>Show image</button>
+{{/if}}
+{{location-map location=rental.city}}
```
toddjordan: Also would like to get the added line syntax coloring here to indicate a new line added to the already existing rental-listing

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
+For Super Rentals, we want to add the ability to display maps of where each rental is located.  To implement this feature, we will take advantage of several Ember concepts:
+
+1. A component to display a map on each rental listing.
+2. A service to keep a cache of rendered maps to use in different places in the application.
+3. A utility function to create a map from Google's map API
```
HeroicEric: The period is missing from this list item

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+```shell
+ember g component location-map
+```
+
+The Ember CLI component generator will create for us a test file, a component JavaScript file, and a template.
+To think about what we want our component to do, will write a testcase.
+In this case we want our Google Maps service to provide us with the map display, so all we want our component to do is take the output of our maps service, which is a map element, and append it to an element in the component's template.
+
+All we want to do in our component is make sure its appending the output of the maps service to an inner map container element.
+To limit the test to validating just this behavior we'll take advantage of the registration api to provide a mock amps service.
```
HeroicEric: "api" -> "API"
"amps" -> "maps"

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+  assert.equal(this.get('mapsService.calledWithLocation'), 'New York', 'should call service with New York');
+});
+```
+
+Notice that our mock service above sets a variable called `calledWithLocation`.  We later assert on this variable to ensure that the service is being called with the same location provided to the component.
+
+To get the test to pass, add the container element to the component template.
+
+```app/templates/components/location-map.hbs
+<div class="map-container"/>
```
HeroicEric: I don't think self-closing `<div>` tags are valid. As I understand it, only [void elements](https://www.w3.org/TR/html-markup/syntax.html#void-element) can omit the end-tag.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+
+```tests/integration/components/location-map-test.js
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+import Ember from 'ember';
+
+let MockMapsService = Ember.Service.extend({
+  getMapElement(location) {
+    this.set('calledWithLocation', location);
+    return document.createElement('div');
```
HeroicEric: It's not clear to me why this function creates and returns a `<div>` element.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,289 @@
...
+
+```app/components/location-map.js
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  maps: Ember.inject.service(),
+
+  didInsertElement() {
+    this._super(...arguments);
+    this.$('.map-container').append(this.get('maps').getMapElement(this.get('location')));
```
HeroicEric: This line might be easier to read if it were broken up a little bit:

```
let location = this.get('location');
let mapElement = this.get('maps').getMapElement(location);

this.$('.map-container').append(mapElement);
```

HeroicEric:
@toddjordan lots of good stuff here 😄

Have you thought about building a more simple version of the component without the service and bringing in the service to improve it after? I think it might be easier to focus on understanding how the service is helpful if you've already already gotten the basic version of the component working. It would also let you more gradually step into some of the tests before needing to create and inject the mock service.

source/localizable/tutorial/service.md
toddjordan: because the maps service returns an element that gets appended to the dom in the component.  I can add some text around it.

toddjordan:
@HeroicEric Thank you for reviewing.  I'm making updates based on your changes. RE: starting with a simpler component,  I could create a component that just shows an image of a map based on location and then swap that impl out with the maps service. I like the idea and will consider it.  It'll be a matter of still keeping the tutorial section relatively short.

toddjordan:
@locks @acorncom can you guys take a look when you get a chance?  Moving onto autocomplete

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
+For Super Rentals, we want to add the ability to display maps of where each rental is located.  To implement this feature, we will take advantage of several Ember concepts:
+
+1. A component to display a map on each rental listing.
+2. A service to keep a cache of rendered maps to use in different places in the application.
+3. A utility function to create a map from the Google map API.
```
acorncom: Google Maps API

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
+For Super Rentals, we want to add the ability to display maps of where each rental is located.  To implement this feature, we will take advantage of several Ember concepts:
```
acorncom: For Super Rentals, we want to be able to display a map showing where each rental is.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+1. A component to display a map on each rental listing.
+2. A service to keep a cache of rendered maps to use in different places in the application.
+3. A utility function to create a map from the Google map API.
+
+We'll work on this feature outside-in, meaning we'll start with the display of the map and work our way back to the actual access of the Google Maps API.
```
acorncom: Might simplify here by saying something like:

> We'll start by displaying the map and work our way back to using the Google Maps API

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+### Display Maps With a Component
+
+For our maps component, all we'll require is a simple component interface that we can simply drop into a template.
+The component will render a map pinned to whatever location we provide it.
+Lets start by appending the component as we want it to the end of the `rental-listing` template, and provide it with the city of the current rental.
```
acorncom: Instead of the above paragraphs, I went to edit things a bit and simplified down to:

> We'll start by adding a map component that shows the rental's city on a map

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+    <img alt={{rental.type}} src={{rental.image}} width="500"/>
+  </p>
+  <button {{action "imageHide"}}>Hide image</button>
+{{else}}
+  <button {{action "imageShow"}}>Show image</button>
+{{/if}}
+{{location-map location=rental.city}}
+```
+
+Next step is to get our page compiling by using Ember CLI to generate this component.
```
acorncom: > Next, generate the map component using Ember-CLI

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+```shell
+ember g component location-map
+```
+
+The Ember CLI component generator will create for us a test file, a component JavaScript file, and a template.
+To think about what we want our component to do, will write a test case.
+In this case we want our Google Maps service to provide us with the map display, so all we want our component to do is take the output of our maps service, which is a map element, and append it to an element in the component's template.
```
acorncom: > Running this command generates three files: a component Javascript file, a template and a test file. To help think through what we want our component to do, we'll write a test case.

> In this case, we plan on having our Google Maps service handle map display. So our component's job will be to take the results from the map service (which is a map element) and append it to an element in the component template.

acorncom:
@toddjordan I've got to shift to some other work, but will come back to this.  Remember our caveats about tutorials getting more attention?  I'm going to be pretty hard on your prose 😄 but I also really appreciate you putting this together

toddjordan:
@acorncom thanks for reviewing.  Be as brutal as you need to be ;-)

locks:
Changed to a more appropriate name :trollface:

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+To generate the maps we'll implement a maps service.
+
+Accessing our maps API through a service will give us several benefits
+
+* It is injected with a [service locator pattern](https://en.wikipedia.org/wiki/Service_locator_pattern), meaning it will abstract the maps API from the code that uses it, allowing for easier refactoring and maintenance
+* It is lazy-loaded, meaning it won't be initialized until its called the first time.
+* It is a singleton, which will allow us cache map data.
+* It follows a lifecycle, meaning we have hooks to execute cleanup code when the app stopped.
+
+Lets get started creating our service by generating it though Ember CLI, which will create the service file, as well as a unit test for it.
```
Turbo87: though -> through ?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+});
+```
+
+### Making Google Maps Available
+
+Before implementing the map utility we need to make the 3rd party map API available to our Ember app.
+There are several ways to include 3rd party libraries in Ember. See the guides section on [managing dependencies](../../addons-and-dependencies/managing-dependencies/) as a starting point when you need to add one.
+
+Since Google provides its map API as a remote script, we'll use curl to download it into our project's vendor directory.
```
Turbo87: Is that a good idea? it seems like the default use case is loading the script in the browser from their server including the API key. if we download it we also won't get any updates and I'm not sure if their data API is actually stable.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+```
+
+Next, generate the map component using Ember-CLI.
+
+```shell
+ember g component location-map
+```
+
+Running this command generates three files: a component JavaScript file, a template, and a test file.
+To help think through what we want our component to do, we'll implement a test case first.
```
acorncom: might be able to remove `case` from "a test case"

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+In this case we plan on having our Google Maps service handle map display.
```
acorncom: Comma after 'case'

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+ember g component location-map
+```
+
+Running this command generates three files: a component JavaScript file, a template, and a test file.
+To help think through what we want our component to do, we'll implement a test case first.
+
+In this case we plan on having our Google Maps service handle map display.  
+Our component's job will be to take the results from the map service (which is a map element) and append it to an element in the component template.
+
+All we want to do in our component is make sure its appending the output of the maps service to an inner map container element.
```
acorncom: > is make sure **it's** appending

acorncom: I think we can drop this entire sentence, as it's almost a duplicate of the sentence above

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+To limit the test to validating just this behavior we'll take advantage of the registration API to provide a mock maps service.
```
acorncom: Comma after behavior

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+import Ember from 'ember';
+
+let MockMapsService = Ember.Service.extend({
+  getMapElement(location) {
+    this.set('calledWithLocation', location);
+    // We create a div here to simulate our maps service,
+    // which will create and then cache the map element
+    return document.createElement('div');
+  }
+});
```
acorncom: Are we assuming folks are doing this with prior testing experience or not?  This is a great example of a mock service, but if we're assuming little to no knowledge of testing prior to this (which might be the case in cases of jQuery converts), we should probably spend a bit more time explaining what we're doing with a mock service ...  /cc @locks ?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+moduleForComponent('location-map', 'Integration | Component | location map', {
+  integration: true,
+  beforeEach() {
+    this.register('service:maps', MockMapsService);
+    this.inject.service('maps', { as: 'mapsService' });
+  }
```
acorncom: And here again, we start throwing **magic** at folks ;-)  Perfectly valid (and a great example of using services in a test), but do we want to soften the blow of this chunk of code by breaking it into chunks were we discuss what we're doing?  Seems like we should segment it up and then show the finished product once we've walked people through what we're going to do ...

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+});
+
+test('should append map element to container element', function(assert) {
+  this.set('myLocation', 'New York');
+  this.render(hbs`{{location-map location=myLocation}}`);
+  assert.equal(this.$('.map-container').children().length, 1, 'container should have one child');
```
acorncom: Can we have better explanatory text for the failure text?  Possibly "our component should be showing at the map"?  container and child are accurate, but also somewhat obscure to a new user ...

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+  assert.equal(this.get('mapsService.calledWithLocation'), 'New York', 'should call service with New York');
```
acorncom: Same goes for this explanatory text.  Maybe:

> the service calledWithLocation variable should be set to New York

?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+});
+```
+
+Notice that our mock service above sets a variable called `calledWithLocation`.  We later assert on this variable to ensure that the service is being called with the same location provided to the component.
```
acorncom: With the explanatory text in our test, think we can drop this sentence?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+To get the test to pass, add the container element to the component template.
+
+```app/templates/components/location-map.hbs
+<div class="map-container"></div>
+```
+
+Then update the component to append the map output to our component.
```
acorncom: to our **container** ?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+We'll add a maps service injection, and call the `getMapElement` function with the provided location.
```
acorncom: Can we link to info about the service injection so folks can go off and read about this?  Feels like we've got more magic here too ;-)

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+export default Ember.Component.extend({
+  maps: Ember.inject.service(),
+
+  didInsertElement() {
+    this._super(...arguments);
+    let location = this.get('location');
+    let mapElement = this.get('maps').getMapElement(location)
+    this.$('.map-container').append(mapElement);
+  }
```
acorncom: Have we explained about `didInsertElement` and where that comes from?  Maybe we could link to the guides as part of our discussion?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+});
+```
+
+### Fetching Maps With a Service
+
+Now we have a passing component integration test, but no maps generating onscreen.
+To generate the maps we'll implement a maps service.
```
acorncom: Comma after "the maps"

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+    let location = this.get('location');
+    let mapElement = this.get('maps').getMapElement(location)
+    this.$('.map-container').append(mapElement);
+  }
+});
+```
+
+### Fetching Maps With a Service
+
+Now we have a passing component integration test, but no maps generating onscreen.
```
acorncom: no maps **showing up when we view our web page.**

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+To generate the maps we'll implement a maps service.
+
+Accessing our maps API through a service will give us several benefits
+
+* It is injected with a [service locator pattern](https://en.wikipedia.org/wiki/Service_locator_pattern), meaning it will abstract the maps API from the code that uses it, allowing for easier refactoring and maintenance
```
acorncom: Can we say this in less 'tech'?  If I'm considering converting to Ember from a jQuery background, my eyes may have just glazed over ;-)  I guess it's worth considering whether simple jQuery users are really our target audience here though ...

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+* It is lazy-loaded, meaning it won't be initialized until its called the first time.
```
acorncom: until **it's** called

And why would our newcomer think this is an advantage?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+* It is a singleton, which will allow us cache map data.
+* It follows a lifecycle, meaning we have hooks to execute cleanup code when the app stopped.
```
acorncom: And why is this an advantage?

source/localizable/tutorial/service.md
acorncom: Let's

acorncom:
@toddjordan I took you at your word ;-)  You've got some great material here, appreciate you putting this up so we can work on it.  I need to take one more pass at this to look through the final section, but just ran out of time on this end.  Will be back later this week :-)

source/localizable/tutorial/service.md
toddjordan: I'm pretty sure its stable by version.  They up the version in the url when they make updates.  For the case of the tutorial I think its fine, since I or the typical tutorial user is not going to acquire a key from them just to do the tutorial sample.  Its worth considering the alternative though, which is linking to it in the index.html.  That way it would be easer to update an existing app (by just changing the url)

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+
+```app/utils/google-maps.js
+import Ember from 'ember';
+
+const google = window.google;
+
+export default Ember.Object.extend({
+
+  init() {
+    this.set('geocoder', new google.maps.Geocoder());
```
chadhietala: There is no need to use `this.set` here instead you just assign it.

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+The CLI `generate util` command will create a utility file and a unit test.
+We'll delete the unit test since we don't want to test Google code.
+Our app needs a single function, `createMapElement`, which makes use of `google.maps.Map` to create our map element, `google.maps.Geocoder` to lookup the coordinates of our location, and `google.maps.Marker` to pin our map based on the resolved location.
+
+```app/utils/google-maps.js
+import Ember from 'ember';
+
+const google = window.google;
+
+export default Ember.Object.extend({
```
chadhietala: Is there a reason to use an `Ember.Object` here?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,293 @@
...
+```app/components/location-map.js
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  maps: Ember.inject.service(),
+
+  didInsertElement() {
+    this._super(...arguments);
+    let location = this.get('location');
+    let mapElement = this.get('maps').getMapElement(location)
```
chadhietala: Why use the lazy-injection version of a service, if all you are doing is immediately causing the lookup to occur?

toddjordan: Thanks for providing input, but I'm not sure what you are asking.  Generally with services, you use the inject helper to make the service available for an object, but that service isn't available until it initially gets called with get.  I agree in this case that the component is getting the service pretty soon after inject, but this is just showing a simple case.  If you really wanted to go further and show off lazy loading, you could take the map lookup out of the didInsertElement and base it off of the component showing up in the viewport, but that would be beyond what we want to show in the tutorial.

source/localizable/tutorial/service.md
toddjordan: Generally, for the tutorial, I'm just using ember objects when there's a need for an object.  I'm using an object in general here so that it can more easily be stubbed in tests.  I could be swayed away from an ember object, but would need a good reason to not use it.

source/localizable/tutorial/service.md
toddjordan: Agree that it doesn't buy me anything special, I just generally use set when setting properties for ember objects, and I think for the sake tutorial its fine to use set for setting object properties the get the reader in the mindset that in most cases they will need or want to.

source/localizable/tutorial/service.md
toddjordan: Added a link to the lifecycle rules page

source/localizable/tutorial/service.md
toddjordan: Added a sentence on what a stub is

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+  assert.equal(this.$('.map-container').children().length, 1, 'the map element should be put onscreen');
+  assert.equal(this.get('mapsService.calledWithLocation'), 'New York', 'a map of New York should be requested');
+});
+```
+
+In the `beforeEach` function that runs before each test, we use the implicit function `this.register` to register our stub service in place of the maps service.
+Registration makes an object available to your Ember application for things like loading components from templates and injecting services in this case.
+
+The call to the function `this.inject.service` injects the service we just registered into the context of the tests, so each test may access it through `this.mapsService`.
+In the example we assert that `calledWithLocation` in our stub is set to the location we passed to the component.
```
acorncom: Nicely done here, that's much clearer ...

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+Now we have a passing component integration test, but no maps showing up when we view our web page.
+To generate the maps, we'll implement the maps service.
+
+Accessing our maps API through a service will give us several benefits
+
+* It is injected with a [service locator pattern](https://en.wikipedia.org/wiki/Service_locator_pattern), meaning it will abstract the maps API from the code that uses it, allowing for easier refactoring and maintenance
+* It is lazy-loaded, meaning it won't be initialized until it is called the first time.
+In some cases this can reduce your app's processor load and memory consumption.
+* It is a singleton, which will allow us cache map data.
+* It follows a lifecycle, meaning we have hooks to execute cleanup code when the app stopped, preventing things like memory leaks and unnecessary processing.
```
acorncom: when the app *stops*

Nicely done on explaining things here, I think that makes it far clearer to folks new to these ideas ...

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+
+Lets get started creating our service by generating it through Ember CLI, which will create the service file, as well as a unit test for it.
```
acorncom: _Let's_ get started

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+  let stubCachedMaps = Ember.Object.create({
+    sanFrancisco: DUMMY_ELEMENT
+  });
+  let mapService = this.subject({ cachedMaps: stubCachedMaps });
+  let element = mapService.getMapElement('San Francisco');
+  assert.equal(element, DUMMY_ELEMENT, 'element fetched from cache');
+});
+```
+
+Note that the test uses a dummy object as the returned map element.  This can be any object because its only used to assert that the cache has been accessed.
```
acorncom: because *it's* only used

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+    element.setAttribute('style', 'width:300px;height:300px;');
+    return element;
+  }
+
+});
+```
+
+### Making Google Maps Available
+
+Before implementing the map utility we need to make the 3rd party map API available to our Ember app.
```
acorncom: map utility **comma**

source/localizable/tutorial/service.md
acorncom: Maybe note it above where I added a note?

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+There are several ways to include 3rd party libraries in Ember. See the guides section on [managing dependencies](../../addons-and-dependencies/managing-dependencies/) as a starting point when you need to add one.
+
+Since Google provides its map API as a remote script, we'll use curl to download it into our project's vendor directory.
+
+From your project's root directory, run the following command to put the Google maps script in your projects vendor folder as `gmaps.js`.
```
acorncom: `gmaps.js`. (or if you're on Windows download the JS file and put it in `vendor/gmaps.js`)

source/localizable/tutorial/service.md
```diff
@@ -0,0 +1,302 @@
...
+
+```shell
+curl -o vendor/gmaps.js https://maps.googleapis.com/maps/api/js?v=3.22
+```
+
+Once in the vendor directory, the script can be built into the app.
+We just need to import it into our build file.
```
acorncom: We just need to tell Ember-CLI to import it using our build file:

source/localizable/tutorial/service.md
acorncom: Also, are we standardizing on calling it Ember-CLI?

acorncom:
@toddjordan Looks good to me, thanks for all the work!  @locks any additional feedback here?

I'll plan on merging this after the weekend unless have other major changes to make.  At some point, we need to ship it! ;-)

source/localizable/tutorial/service.md
toddjordan: Am going to keep for now until we add the style addon section.

acorncom:
Merging here, nice work!

PR #1405, opened 2016-04-26
mentioning multiword route hyphenating convention

### LINKED ISSUE #1404: Mention multi-word routes
Related: https://github.com/emberjs/ember.js/issues/11487

maprules1000:
closes #1404


source/localizable/routing/defining-your-routes.md
```diff
@@ -28,6 +28,7 @@ Router.map(function() {
   this.route('favorites', { path: '/favs' });
 });
 ```
+If your route name has multiple works then your route should follow convention and be hyphenated, not underscored. Routes with underscores in them will not be rendered.
```
locks: Can you tweak an existing example, or add a new one? Also, I think it suffices to say that Ember expects the names to be dasherized.

source/localizable/routing/defining-your-routes.md
```diff
@@ -29,6 +29,16 @@ Router.map(function() {
 });
 ```
 
+If your route name has multiple works then your route should follow convention and be hyphenated. Routes with underscores in them will not be rendered.
```
maprules1000: However, the path does allow underscores, despite convention. Should I mention this, or should we lead users to believe the route and path should both use hyphens?

locks: @maprules1000 the path is entirely unrelated to Ember's resolver/container.

maprules1000: @locks ok 😄

locks: It's hard to explain in a way that doesn't talk about the resolver/container, which we want to avoid :P but the shorter explanation is that all of the classes registered with Ember are dasherized, so that's how it expects the lookups to be. `path` is what shows up in the URL, which can be anything (as long as it's a valid URL :P).

maprules1000: ok so is this explanation and example sufficient?

eccegordo: Shouldn't it be "multiple words" not "multiple works"?

maprules1000: Whelp. Yes.

rwjblue: > It's hard to explain in a way that doesn't talk about the resolver/container

I think the easiest explanation is just about files on disk. ember-cli expects all files on disk to be dasherized.  The first argument to `this.route` represents the path on disk so it should be dasherized.  If however, you need something else in your URL's you can use `path` option to customize...

maprules1000: @rwjblue should we be mentioning ember-cli here and using it as an explanation? I mean, if the route uses an underscore is ember-cli the reason a route will fail to render?

locks:
Thank you for your contribution @maprules1000, but we are going with #1410 instead as we feel it approximates the Guides style more closely.
Hope to see more PRs from you soon :)

PR #1406, opened 2016-04-26
Fixed broken link for `ember-deploy`

jacobq:
Looks like the repository name/path for this changed.


locks:
Appreciated! This had been previously fixed, but a bad merge reverted it :x

@homu r+

PR #1407, opened 2016-04-26
clarify preventDefault=false

caseywatts:
(minor text change)

Previously, `with` vs `without` was unclear. Even after a few glances I almost didn't notice this difference between the sentences. Using `omitted` vs `present` makes the difference here clearer.


locks:
Thanks for your contribution!

@homu r+

PR #1409, opened 2016-05-03
Change parameter name to match description

jacobq:
Wouldn't it make more sense to call this controller `my-controller-name` instead of `my-component-name` since it is not a component?


locks:
Thanks for your contribution!

@homu r+

PR #1410, opened 2016-05-03
Describe multiword route naming conventions.

### LINKED ISSUE #1404: Mention multi-word routes

dpdawson:
closes #1404

### DIFF
--- a/source/localizable/routing/defining-your-routes.md
+++ b/source/localizable/routing/defining-your-routes.md
@@ -46,6 +46,30 @@ routes, using the name that you provided to the `route` method.
 The `{{link-to}}` helper will also add an `active` class to the link that
 points to the currently active route.
 
+Multi-word route names are conventionally dasherized, such as:
+
+```app/router.js
+Router.map(function() {
+  this.route('blog-post', { path: '/blog-post' });
+});
+```
+
+The route defined above will by default use the `blog-post.js` route handler,
+the `blog-post.hbs` template, and be referred to as `blog-post` in any
+`{{link-to}}` helpers.
+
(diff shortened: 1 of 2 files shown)


source/localizable/routing/defining-your-routes.md
```diff
@@ -46,6 +46,30 @@ routes, using the name that you provided to the `route` method.
...
+
+The route defined above will by default use the `blog-post.js` route handler,
+the `blog-post.hbs` template, and be referred to as `blog-post` in any
+`{{link-to}}` helpers.
+
+Multiword route names that break this convention, such as:
+
+```app/router.js
+Router.map(function() {
+  this.route('blog_post', { path: '/blog-post' });
```
locks: didn't you mean it the other way around?

alexspeller:
👍 this looks correct for me, although frankly I think it would be a good idea to just deprecate non-dasherized route names. Also, route names with a `/` in break stuff real bad (i.e. `this.route('foo/bar')`), it would be good to mention that somewhere, or deprecate and warn about it. But that's not a comment on this PR necessarily.

source/localizable/routing/defining-your-routes.md
alexspeller: I _think_ it's correct as written - module lookup for js file will be normalized, but not link-to

rwjblue:
We should probably just deprecate if the first arg to `this.route` isn't already dasherized, but that is a story for another day 😝

locks:
👌 

@homu r+

PR #1408, opened 2016-04-30
revamp autocomplete tutorial to be a filter

toddjordan:
Part of #1374

PR #1412, opened 2016-05-04
remove unneeded attributes: filteredList

williamherry:
filteredList is passed in from the caller


toddjordan:
Thanks @williamherry you are right, though autocomplete is going through a revamp (see #1408 ) making this eventually moot.  I'll go ahead and merge to clear the pr and update the tutorial page until then...

toddjordan:
@homu r+

PR #1413, opened 2016-05-05
Fix input example name attribute in the-component-lifecycle

remino:


toddjordan:
good catch @remino and thanks for the contribution!

toddjordan:
@homu r+

PR #1415, opened 2016-05-06
Add favorites/index default in routing example

remino:
I can't tell if this is too verbose, but I thought it was important, or at least more consistant, to point out that adding `this.route('favorites');` in the router will also automatically add its own index route.


locks:
It won't @remino. What automatically adds an index route is passing a function to `this.route`, which signals that the route might have children, and thus an implicit index. So, an implicit index per `function`.

remino:
Alright. My bad, then. Please discard.

-- 
Rémi

RÉMINO
https://remino.net/

locks:
Thank you for your contribution @remino, I have opened https://github.com/emberjs/guides/pull/1416 to hopefully clarify this issue. Comment there to let us know if it's useful.

PR #1416, opened 2016-05-06
Update defining-your-routes.md

locks:
Clarify when a new level of nesting appears.


remino:
That's good, but I wonder if the example should actually show it. To illustrate:

- - -

A nested router like this:

```app/router.js
Router.map(function() {
  this.route('posts', function() {
    this.route('comments', function() {});
    this.route('favorites');
  });
});
```

Is the equivalent of:

```app/router.js
Router.map(function(){
  this.route('index', { path: '/' });
  this.route('posts', function() {
    this.route('index', { path: '/' });
    this.route('comments', function() {
        this.route('index', { path: '/' });
    });
    this.route('favorites');
  });
});
```

locks:
@homu r+

PR #1425, opened 2016-05-10
Missing Text in code navigation info

samiell:
changed 'app/components/user-profile.hbs' to 'app/templates/components/user-profile.hbs'


locks:
Thanks for your contribution!

@homu r+

PR #1427, opened 2016-05-11
Update quick-start.md

gdomingu:
Fixed possible typo. I think command should be consistent with https://guides.emberjs.com/v2.5.0/tutorial/ember-cli/#toc_the-development-server.


locks:
Either will actually work, but I like the consistency with this change. Thank you :)

@homu r+

PR #1428, opened 2016-05-11
Fix typo: "Web Sockets" -> "WebSockets"

HeroicEric:
WebSocket is how the protocol is spelled in https://tools.ietf.org/html/rfc6455 and is also used [elsewhere in the guides](https://guides.emberjs.com/v2.5.0/models/).

### DIFF
--- a/source/localizable/applications/services.md
+++ b/source/localizable/applications/services.md
@@ -7,7 +7,7 @@ include:
 
 * User/session authentication.
 * Geolocation.
-* Web Sockets.
+* WebSockets.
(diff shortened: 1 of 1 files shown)


locks:
Now I'm confused as to how this goes by the spellchecker. Huh. :P Thanks!

@homu r+

HeroicEric:
@locks looks like there's already an entry in [spelling-exceptions](https://github.com/emberjs/guides/blob/master/data/spelling-exceptions.txt#L268-L269)

PR #1432, opened 2016-05-18
Install addons using ember-cli instead of npm

jacobq:
@alexspeller, is this better or should the `npm` part be removed completely?


locks:
@jacobq it should be removed. We should consider adding an npm section above the bower on later on, but for now removing it is the best course.

jacobq:
@locks Is this better?

locks:
👌  Thanks for your contribution!

@homu r+

PR #1202, opened 2016-01-28
Change search provider to DocSearch

jeradg:
(Reopening https://github.com/emberjs/guides/pull/1188)

This PR changes the Guides search provider from Swiftkey to DocSearch.

See issues https://github.com/emberjs/guides/issues/1184 and https://github.com/emberjs/guides/issues/1183 for discussion.

Functionality checked in Chrome and Firefox latest.

## Todo:

- [ ] More browser testing

### Screenshots

*Small width:*
docsearch-small-width

*Large width:*
docsearch-large

*Typo correction:*
docsearch-spellcheck

PR #1199, opened 2016-01-28
CSS tweaks to search menu

acorncom:
Mainly making the menu a bit smaller and more in keeping with the previous search menu design

Previously:
screen shot 2016-01-28 at 9 05 54 am

With these changes:
screen shot 2016-01-28 at 9 06 41 am

PR #1433, opened 2016-05-19
Search improvements

acorncom:
- [x] Pending latest version of the guides being indexed on Algolia's end


acorncom:
Supersedes #1202 and #1199

acorncom:
@locks @toddjordan This is ready for review.  One thing to note (that differs from our current setup and that we may want to handle before we merge this).

Algolia does search indexing *post*-deploy.  We've got a few options for handling that:

- per their suggestion: "if you could provide us a public beta version of this new doc, we could index it before the official release".  This would only work if we're deploying at the appropriate location (i.e. `/v2.6/`) but don't enable it for all visitors to see (primarily our `firebase.json` file and our `versions.json` file).  We can't have them index the beta site, as the index contains the page url and if that's not accurate, we'll end up with very confused visitors ;-)
- leave search broken for the 1 day window (or so) that it takes for them to re-index
- jury-rig something else?

Thoughts here?

locks:
> leave search broken for the 1 day window (or so) that it takes for them to re-index

acorncom:
Then we are ready to merge here.  Would suggest we deploy soon so we can iron out the bugs prior to 2.6 release

toddjordan:
@acorncom LGTM 👍 I'll let @locks merge 💥  🔫

locks:
@toddjordan go ahead, I don't mind :trollface:

locks:
@toddjordan @acorncom 2.6 is releasing today

acorncom:
This is back to pending while I wait to hear from the Algolia folks about a slight tweak they've made on their end.  Once that's in, I'll merge this

acorncom:
@homu r+

PR #1444, opened 2016-05-30
Remove extra space between dependent keys for brace expansion example

rafael-paiva:
* To avoid confusion; If there's a space between dependent keys we get the following error `Brace expanded properties cannot contain spaces, e.g. "user.{firstName, lastName}" should be "user.{firstName,lastName}"`.


locks:
Appreciated @rafael-paiva!

@homu r+

PR #1446, opened 2016-05-31
Update deploy to use ember-cli-surge

kiwiupover:
The ember-cli-surge addon has been updated so that the surge cli is not longer needed. 
https://github.com/kiwiupover/ember-cli-surge

### DIFF
--- a/source/localizable/tutorial/deploying.md
+++ b/source/localizable/tutorial/deploying.md
@@ -13,35 +13,26 @@ scp -r dist/* myserver.com:/var/www/public/
 
 ## Deploying to surge.sh
 
-[Surge.sh](http://surge.sh/) allows you to publish any folder to the web for free.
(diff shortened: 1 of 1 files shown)


kiwiupover:
What should I do about the translated files? It looks only the japanese file has been translated. I can update all the others if you want.

Cheers

locks:
Hi @kiwiupover, thank you for letting us know about the addon, but we purposefully do it the manual way in the tutorial for a couple of reasons: 

* External dependency: This is something that we have to be careful with. We need some level of control over the dependency, or a maintenance/quality guarantee, otherwise we risk having a broken Guides;
* "Magic": The intention is to show that after you build an Ember project, that they're static files that you can deploy as you do with other static files.

I'm closing the PR due to the constraints above, but we'll keep this in mind for a potential deployment section. Thanks once again!

P.S. For future contributions, you only need to modify the original English file, the rest are synchronized through Crowdin.

PR #1448, opened 2016-06-02
didInitAttrs hook is added to the guide.

ykaragol:
According to [this blog post](http://emberjs.com/blog/2015/06/12/ember-1-13-0-released.html#toc_component-lifecycle-hooks) didInitAttrs hook is available since 1.13.


toddjordan:
Its deprecated in 2.6 so we don't want to advertise it. http://emberjs.com/blog/2016/04/11/ember-2-5-released.html#toc_ember-js-2-6-beta

PR #1453, opened 2016-06-04
Tutorial revisions

acorncom:
### Page names
- [x] Creating a new app, rename to Creating Your App ?
- [x] rename acceptance tests page -> Setting Up Tests page?

### Creating a new app 
- [x] welcome page note at bottom of /tutorial/ember-cli/ will be slightly out-of-date with the new welcome page add-on (2.6), do we care?  not sure we do ...
- [x] when generating the tests, do I do this in a separate terminal window or do I have to stop my server, add them and re-run?

### Acceptance test page
- [x] not sure I’m a fan of using the red / green for short snippets with a lot of minor changes ... (first sample acceptance test), thoughts here?
- [x] when I run the test suite for the first time, do I do this in a new command line window or the same one?  seems like we should be gentle and walk folks through it ...
- [x] after initial test run, I see 10/10 instead of one successful test (JSHint results)


acorncom:
Needing to knock off on this for now, will be chipping away at it over time.  Put in some thoughts for us to discuss in the main description section above ...

toddjordan:
For Title name changes, I think we should delineate that its "acceptance" testing, though I'm ok with changing the title and doing it in the text.  I'm also ok with changing to "Creating your app", though in both cases I'm ok as-is as well.

toddjordan:
I think its useful to highlight small changes because it points the user to what's changed from what's already there.

toddjordan:
@acorncom I like the changes in the PR, we just need to resolve conflicts

toddjordan:
Thanks!

toddjordan:
@homu r+

PR #1454, opened 2016-06-04
Correct `filterBy` example

GavinJoyce:
```js
remaining: Ember.computed('todos.@each.isDone', function() {
  var todos = this.get('todos');
  return todos.filterBy('isDone', false).get('length');
})
```

isn't the same as:

```js
remaining: Ember.computed.filterBy('todos', 'isDone', false)
```

as one returns a count and the other returns an array.

I've modified them both to return an `incomplete` array and have updated the examples to use `incomplete.length`


locks:
Your usage of `todos` is making the spell checking fail, could you add it to the exception list? Or if possible, avoid referring to them by that word someone, though that seems needlessly constraining.

GavinJoyce:
I've updated it to `todo's` the be consistent with the rest of the guides

locks:
@homu r+

PR #1458, opened 2016-06-05
Set links apart from content.

### LINKED ISSUE #1450: Not enough contrast between links and other text
I'm very, very mildly color-blind, but I can't distinguish links at all. It all looks the same to me. A higher contrast color combination or a simple underline would help.

nathanhammond:
Closes #1450.


locks:
@homu r+

PR #1461, opened 2016-06-05
Change where to check for Ember Data version

yawboakye:
Ember Data dependency moved from `bower.json` to `package.json`.


locks:
Thanks for your contribution!

@homu r+

locks:
@yawboakye it's great as it is ❤️

PR #1462, opened 2016-06-05
Use `ember serve` to be consistent

GavinJoyce:
It would be nice to be consistent with our use of `ember server` and `ember serve`. I think we should use `ember serve` everywhere as that is what ember-cli promotes as the main command:

screen shot 2016-06-05 at 12 36 20


GavinJoyce:
screen shot 2016-06-05 at 12 44 28

yawboakye:
@GavinJoyce I found that the command's file is called `serve`, and the name of the command is `serve`. `server` and `s` are rather listed as aliases. Please consider these arguments and if they make any sense maybe you can reopen this PR.

PR #1464, opened 2016-06-05
Update Bower dependencies of a newly generated app

yawboakye:
Remove jQuery and Ember Data. Add Ember CLI Shims and Ember CLI Test Loader.


locks:
😁 
@homu r+

PR #1466, opened 2016-06-06
Fix JSON example

franzliedke:
As far as I understand, this is meant to be a hypermedia relationship, with the `related` key containing the URL of the related resources.

See the `comments` example on http://jsonapi.org/.


bmac:
👍

locks:
@homu r+

PR #1467, opened 2016-06-06
[BUGFIX] Remove references to Ember.Logger.assert

Serabe:
As explained in [this comment](https://github.com/emberjs/ember.js/issues/13597#issuecomment-223951392):

> Ember.Logger is largely a polyfill from the dark ages when we couldn't rely
on console methods to exist or function properly.

This PR removes references to `Ember.Logger`

Related to emberjs/ember.js#13597

_Note:_ The extra line at the end of file is just my editor complying with the [editorconfig](https://github.com/emberjs/guides/blob/master/.editorconfig#L10)

### DIFF
--- a/source/localizable/configuring-ember/debugging.de-DE.md
+++ b/source/localizable/configuring-ember/debugging.de-DE.md
@@ -103,7 +103,7 @@ You can provide an `onerror` function that will be called with the error details
 
 ```app/app.js import Ember from 'ember'; import RSVP from 'rsvp';
 
-RSVP.on('error', function(error) { Ember.Logger.assert(false, error); });
+RSVP.on('error', function(error) { Ember.assert(false, error); });
 
     <br />#### Errors within `Ember.run.later` ([Backburner.js](https://github.com/ebryn/backburner.js))
     
--- a/source/localizable/configuring-ember/debugging.es-ES.md
+++ b/source/localizable/configuring-ember/debugging.es-ES.md
@@ -103,7 +103,7 @@ You can provide an `onerror` function that will be called with the error details
 
 ```app/app.js import Ember from 'ember'; import RSVP from 'rsvp';
 
-RSVP.on('error', function(error) { Ember.Logger.assert(false, error); });
+RSVP.on('error', function(error) { Ember.assert(false, error); });
 
     <br />#### Errors within `Ember.run.later` ([Backburner.js](https://github.com/ebryn/backburner.js))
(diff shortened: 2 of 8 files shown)


mixonic:
Some thoughts @Serabe:

* `info` is a private API: http://emberjs.com/api/#method_info we should not be documenting it like public API. If we want to make those APIs public, I kind of want to suggest we do it holistically with an RFC that coordinates a direction around the logging APIs.
* Though @rwjblue suggested `Ember.Logger` is no longer needed, IE9 appears to still require it http://caniuse.com/#feat=console-basic. If the debugging tools are not open, calls to `console` would end up throwing an error as `console` is `undefined`. This suggests `Ember.Logger` is *not* ready for retirement.

Based on this I think we're stuck supporting `Ember.Logger` as long as we support IE9.

rwjblue:
I agree with @mixonic that we cannot remove Ember.Logger from the Ember code base. Regardless, this change seems good to me.

mixonic:
I would merge they demonstrated use of assert without hesitation.

@rwjblue Ember.info is private but this PR adds it to docs demoing it in app code. You suggest you would like to make Ember.info public?

rwjblue:
I missed that part of your initial comment, I was mostly focusing on the need for Ember.Logger as long as we support IE9.

I'd suggest it was privatized during the great privatization of all things back in the 1.12 timeframe, and not necessarily intended to be private, but I'm happy to discuss with the larger group to confirm. 

---

Regardless, I don't think the guides are fundamentally wrong right now (while using Ember.Logger), as far as I know we don't suggest these logs are silent in production builds.

Serabe:
The problem with current usages is that, though is not suggested that `Ember.Logger` is silent in production, it has been suggested to move in the past to use `Ember.x` instead of `Ember.Logger.x` cause they are noops in production.

Having a logging API that can be silent (it would be awesome if it could be conditionally silent) would be great, not only for its own uses, but for current addons that want to log to provide useful information to the developers while using it.

nathanhammond:
ember-cli-defeatureify is the current method for removing, code, but it's not run by default against apps. We want to support this inside of ember-cli: https://github.com/ember-cli/rfcs/pull/50

Serabe:
I've updated this PR leaving alone `Ember.Logger.info` and only removing `Ember.Logger.assert`

locks:
Thanks for your contribution!

@homu r+

PR #1471, opened 2016-06-08
Fix JSON-API serializer extending example

franzliedke:
Took me quite a while to find out all the problems in this snippet that was causing my Ember app to crash. :(

I believe there are other examples on this page where using the `json-api` serializer instead of `json` would be more appropriate, given the surrounding text. But I'd prefer if one of the maintainers would just go through these, as I don't want to make all these decisions...

Thanks.

### DIFF
--- a/source/localizable/models/customizing-serializers.md
+++ b/source/localizable/models/customizing-serializers.md
@@ -318,9 +318,10 @@ payload. For example, if your backend returned attributes that are
 method like this.
 
 ```app/serializers/application.js
-import JSONSerializer from 'ember-data/serializers/json';
+import Ember from 'ember';
+import JSONAPISerializer from 'ember-data/serializers/json-api';
 
-export default JSONSerializer.extend({
+export default JSONAPISerializer.extend({
   keyForAttribute: function(attr) {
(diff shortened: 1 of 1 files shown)


source/localizable/models/customizing-serializers.md
```diff
@@ -318,9 +318,10 @@ payload. For example, if your backend returned attributes that are
 method like this.
 
 ```app/serializers/application.js
-import JSONSerializer from 'ember-data/serializers/json';
+import Ember from 'ember';
```
locks: remove this line please

locks:
Hm, I think you can safely change the code samples to JSONAPIAdapter/Serializer, up to the JSONSerializer-specific section.

source/localizable/models/customizing-serializers.md
franzliedke: Why? It's clearly needed in the implementation...

locks: oh, I missed the `Ember.String`

franzliedke:
@locks Okay, I changed the other ones, too.

locks:
@homu r+

PR #1473, opened 2016-06-09
Keep using the app created at start of tutorial

yawboakye:
Let's keep using the app we created at the start of the tutorial, and
wrote acceptance tests for.


toddjordan:
Thanks @yawboakye!

toddjordan:
@homu r+

PR #1478, opened 2016-06-09
Make tests match the scaffold provided earlier

StevenNunez:
Doc references a list of goals, then deviates. This brings them back in alignment.


locks:
Thanks for your contribution!

@homu r+

PR #1480, opened 2016-06-10
Setup continuous delivery to staging

acorncom:
With these changes, the latest changes to `master` will always be visible at https://ember-guides-staging.firebaseapp.com/v2.6.0/

I'm not updating https://github.com/emberjs/guides.emberjs.com on every commit (that seems excessive), but we'll end up with functionally the same setup when we do more major deploys.

When we do any changes on older versions (by manually working with the repo above) we'll need to make sure to update the latest version with the latest build from `master` here as well ...

/cc @locks @toddjordan

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ addons:
   apt:
     packages:
     - aspell-en
+cache: bundler
 script: bundle exec rspec spec && bundle exec middleman build --verbose
 deploy:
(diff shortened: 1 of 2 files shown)


toddjordan:
just need to squash

toddjordan:
@homu r+

acorncom:
Ditto :-)

Issue #1482, opened 2016-06-10
Tutorial: Building a Complex Compoment not working for me


gr2m:
After I followed the instructions to add the complex component, no rentals are shown any longer. Here is the diff (together with two steps before, bu they were minor and worked): https://github.com/gr2m/super-rentals/commit/7b03bd67811287c996bd5832964232c5660c7927

![dafuq](https://cloud.githubusercontent.com/assets/39992/15965514/68244602-2eec-11e6-95fe-a533a76cc3ab.gif)

toddjordan:
@gr2m we made some fixes to the tutorial yesterday afternoon and I think one should address that.  Can you verify?

toddjordan:
I'll try the tutorial steps out for this and make sure too.

courajs:
https://github.com/gr2m/super-rentals/commit/7b03bd67811287c996bd5832964232c5660c7927#commitcomment-17822592

toddjordan:
yeap, that's it.  Good find.  That fix didn't make it back.

toddjordan:
@courajs @gr2m can you review the pr and make sure it works for you?

gr2m:
@toddjordan yes, @courajs’ fix makes it work again

toddjordan:
sorry, I meant my PR https://github.com/emberjs/guides/issues/1483

gr2m:
done. I applied the exact change to my local code and it fixed it: https://github.com/gr2m/super-rentals/commit/f71b880275d1c6d62b99e7de57d6e73d840ccbb1

toddjordan:
Thank you so much for reporting this.  We'll try to get it back on the site asap

toddjordan:
fix merged... closing issue.

PR #1483, opened 2016-06-10
change list-filter attr to match yield

toddjordan:
addresses #1482


gr2m:
👍 fixes it for me

toddjordan:
awesome

toddjordan:
@acorncom could you take a look and merge?  We should try to do a deploy today with this

toddjordan:
cc @locks

acorncom:
@homu r+

PR #1486, opened 2016-06-10
Shifting back to recommending Ember-CLI 2.6

acorncom:
For install instructions

### DIFF
--- a/source/localizable/getting-started/index.md
+++ b/source/localizable/getting-started/index.md
@@ -55,7 +55,7 @@ need for a browser to be open. Consult the [PhantomJS download instructions](htt
 Install Ember using npm:
 
 ```bash
-npm install -g ember-cli@2.5.1
+npm install -g ember-cli
 ```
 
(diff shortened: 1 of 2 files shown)


source/javascripts/app/version-select.js.erb
```diff
@@ -14,8 +14,8 @@ window.GUIDE_VERSIONS.onReady(function (versions) {
   });
 
   var npmInstallVersion = $('pre:contains("install -g ember-cli")');
-  //only append the version if there's one specified and not one already in the content
-  if (currentRevision.slice(0,1) !== "C" && npmInstallVersion.text().indexOf("@") === -1) {
```
toddjordan: @acorncom should we leave this code so that when we want to force a version we just update the markdown to have the `@version`?

acorncom: Good point, I like it

acorncom:
@homu r+

PR #1487, opened 2016-06-10
Tutorial: clarify build before deploy

gr2m:
I think the instructions are good as they are, I don’t know why I built with `--env=production`, I didn’t think about it. But I tested it with just `ember build` and it works. But if you want to be 100% clear, we can add an extra explanation to it

@toddjordan

### DIFF
--- a/source/localizable/tutorial/deploying.md
+++ b/source/localizable/tutorial/deploying.md
@@ -1,4 +1,4 @@
-To deploy an Ember application simply transfer the output from `ember build` to a web server.
+To deploy an Ember application simply transfer the output from `ember build` to a web server (make sure to not build with `--enviroment=production`, as the data mocked with Mirage will only work with a development build).
 This can be done with standard Unix file transfer tools such as `rsync` or `scp`.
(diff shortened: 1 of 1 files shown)


toddjordan:
Thanks, @gr2m

toddjordan:
@homu r+

source/localizable/tutorial/deploying.md
```diff
@@ -1,4 +1,4 @@
-To deploy an Ember application simply transfer the output from `ember build` to a web server.
+To deploy an Ember application simply transfer the output from `ember build` to a web server (make sure to not build with `--enviroment=production`, as the data mocked with Mirage will only work with a development build).
```
acorncom: Could we tweak this a bit?

> simply transfer the output from `ember build --environment=development` to a web server (this ensures that Mirage continues to work instead of turning off as happens with `--environment=production`)

Thanks!

acorncom:
Ah, got beaten here, I'll tweak separately ;-)

toddjordan:
doh! sorry bout that @acorncom

PR #1488, opened 2016-06-10
Updated tweaks to deploy instructions

acorncom:
@toddjordan The reason folks were having problems is that we specifically stated that they should use `ember build --environment=production` further down the page ... 😄


toddjordan:
@homu r+

PR #1489, opened 2016-06-10
Setting up continuous deploys to live setup

acorncom:
Here we go, living in style! Shipping now so we can try it out when it's slow


acorncom:
@toddjordan FYI, merged without asking for your review so I can test while it's slow.  I can fix this if things break

PR #1492, opened 2016-06-11
Update reference in {{#link-to }} helper.

vjrngn:
The {{#link-to}} helper is supposed to point to the 'contact' route and not a named route of 'index' which is not defined.


locks:
Thanks for your contribution!

@homu r+

PR #1494, opened 2016-06-13
Long code samples font shouldn't be auto-adjusted

### LINKED ISSUE #1445: Code block text does not match line number spacing at smaller resolutions
When viewed on an iPhone 6 (not in a browser at smaller browser widths), code block text does not match up with the line numbers, leading to it being quite distracting ...
large-text

jkarsrud:
This fixes #1445 by setting the `text-size-adjust` property on `.highlight pre`-elements. Mobile Safari would auto-adjust the font-size of wide elements to make sure they weren't too small to read, which resulted in code samples with long lines being larger than the line numbers.  

I added a mixin with browser-prefixes for Mozilla and IE/Edge too, but I'm lacking devices to test that on, but it should be consistently fixed (if there were any issues there in the first place).

### DIFF
--- a/source/stylesheets/base/_typography.scss
+++ b/source/stylesheets/base/_typography.scss
@@ -1,4 +1,5 @@
 @import "mixins/hidpi";
+@import "mixins/text-adjust";
 
 body {
   @include font-feature-settings("kern", "liga", "pnum");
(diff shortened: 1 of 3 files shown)


locks:
@homu r+

acorncom:
Thanks!

PR #1496, opened 2016-06-14
Update quick-start.md

Cody-Cooper:
Added steps to remove the ember welcome page and to create the application.hbs file


acorncom:
Any reason you closed this?  We don't necessarily need to remove the `ember-welcome-page` add-on from `package.json`, as generating a new template should be enough, but would be happy to get these instructions improved ...

PR #1503, opened 2016-06-16
Fix misspelling of "vendor"

jordan-king:


locks:
Thanks for your contribution!

@homu r+

PR #1508, opened 2016-06-18
fix link to keyForAttribute()

HusseinMorsy:
the original link to keyForAttribute() has return character, because of a line break inside the link


locks:
Thanks for your contribution!

@homu r+

PR #1509, opened 2016-06-18
 Added missing classes to getting started

seansellek:
If this code were to be copied/pasted, several previously established classes would be missing. Specifically, the `location` class is necessary for the `list-filter` integration tests to pass.


acorncom:
@homu r+  Thank you!

PR #1515, opened 2016-06-21
Miscellaneous edits

CodingItWrong:
Word choice, formatting, and accuracy.

PR #1518, opened 2016-06-21
Localhost link standardization

### LINKED ISSUE #1514: Inconsistent localhost links
I've noticed some inconsistencies with the way `localhost` links are presented in the guides, and I'd like to do a quick PR to make them consistent. Just wanted to see if anyone had a preference first. The two styles I've seen are:

1. Not linked, `http://` not shown, backticks, e.g. `localhost:4200/about`
2. Linked, `http://` shown, no backticks, e.g. [http://localhost:4200/about](http://localhost:4200/about)

I could do either of those, or a third option:

3. Linked, `http://` shown, with backticks, e.g. [`http://localhost:4200/about`](http://localhost:4200/about)

Preferences?

CodingItWrong:
Split off from #1515.

Makes the format of links to localhost throughout the docs consistent. Previously not all places `localhost` was shown were links, not all showed `http://`, and not all were in backticks; now all instances of `localhost` have all three of those attributes.

[Fixes #1514]

### DIFF
--- a/source/localizable/configuring-ember/handling-deprecations.md
+++ b/source/localizable/configuring-ember/handling-deprecations.md
@@ -42,7 +42,7 @@ Once installed, the addon works in 3 steps:
 The ember-cli-deprecation-workflow addon provides a command that will collect deprecations from your console and generate JavaScript code listing
 its findings.
 
-To collect deprecations, first run your in-browser test suite by starting your development server and navigating to `http://localhost:4200/tests`.  If your test suite isn't fully covering your app's functionality, you may also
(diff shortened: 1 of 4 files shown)


locks:
@homu r+

acorncom:
Glad to see this getting standardized.  My one concern with linking to the `http://localhost:4200` (as we're now doing) is that it may impact our SEO rankings to be linking to what (from Google's point of view) may be considered a nonexistent page.  I *like* the user-friendliness of it, just don't want to see us get penalized for this change.  @locks thoughts?

PR #1520, opened 2016-06-21
Remove "pattern" because the pattern isn't acting

CodingItWrong:
Split off from #1515.

The "service locator pattern" is in use, but a pattern doesn't do anything: a pattern is a description of a type of code. It's more correct to say that the "service locator" does something.

An alternative could be to say "It is injected using the service locator pattern". The pattern *is* used, but the pattern isn't the thing that acts.


locks:
@homu r+

PR #1521, opened 2016-06-21
Fix incorrect reference to `rentalUnit` variable.

CodingItWrong:
Split off from #1515.

This paragraph references a `rentalUnit` variable that's passed into the component, but the correct name of the variable is `rental`. This updates the paragraph to match the code.

A clearer option could be to leave the paragraph as-is, and *actually* name the variable `rentalUnit` in the code here and in `model-hook.md`.

### DIFF
--- a/source/localizable/tutorial/simple-component.md
+++ b/source/localizable/tutorial/simple-component.md
@@ -86,7 +86,7 @@ To start, let's move the rental display details for a single rental from the `in
 In our `index.hbs` template, let's replace the old HTML markup within our `{{#each}}` loop
 with our new `rental-listing` component:
 
-```app/templates/index.hbs{+14,-15,-16,-17,-18,-19,-20,-21,-22,-23,-24,-25,-26,-27,-28,-29}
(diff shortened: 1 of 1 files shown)


toddjordan:
@CodingItWrong I think changing the code back to rentalUnit would be easier for the reader to understand.  Do you mind updating the PR with what you suggest?

CodingItWrong:
Will do!

CodingItWrong:
@toddjordan When I looked into making that change, the simplest way to do it seemed to be to rename the loop variable, and keep the component attribute the same. Because the page says we simply _move_ the rental display details from `index.hbs` to `rental-listing.hbs`, it seems simplest to keep the variable names the same in both of those places. So the smallest change is just to change the loop variable that occurs in `index.hbs` to be different. Let me know your thoughts!

toddjordan:
Sounds good, now that you've changed the code, you can change the text back...
"... and assign each `rentalUnit` as the `rental` ..."

CodingItWrong:
@toddjordan Sorry about that; it should be correct now.

toddjordan:
LGTM. Thanks for contribution!

toddjordan:
@homu r+

PR #1522, opened 2016-06-21
Add comments in generated model

CodingItWrong:
Split off from #1515 

When Ember Data 2.6 generates a model, it adds commented-out imports for `ember-data/attr` and `ember-data/relationships`. This commit adds these to the docs showing a generated model, to avoid confusion.

The contributing guidelines say to "Omit the boilerplate that Ember CLI generates, especially the `import Ember from 'ember'` at the top of every file." But in this case `import Model from 'ember-data/model';` was already being shown, which makes sense because this is the tutorial orienting new developers to what they'll see. It seems that including these comments would avoid some confusion.


toddjordan:
@locks @acorncom I'm ok with this, since its what gets generated from Ember CLI.  

I guess another option is to show them as removed in the code (with red indicators).  
What do you guys think/prefer?

toddjordan:
@locks, where's this at these days? Should we go forward and merge?

CodingItWrong:
@toddjordan @locks After #1578, I don't believe this PR is relevant anymore. Now that Ember Data generates models with `import DS from 'ember-data'` style imports, there are no commented-out imports, so there's no question about whether or not to include commented-out imports in the docs.

locks:
Thank you for the follow up @CodingItWrong. This seemed stalled but there was actually some internal discussion resulting in Ember Data's blueprints going back to the old imports, like you mention.
I hope to see further contributions from you :)

CodingItWrong:
No problem. Any contributions are of course always y'all's discretion, and y'all accepted a bunch of my contributions right away =]

PR #1523, opened 2016-06-21
Fixes #1512 - Links attr to API documentation

### LINKED ISSUE #1512: Suggestion: Example showing defaultValue as a function in defining-models
In:
https://github.com/emberjs/guides/blob/master/source/localizable/models/defining-models.md

there should be an example showing that default value can also be a function.  As in: http://emberjs.com/api/data/classes/DS.html#method_attr

locks:


rwjblue:
@homu r+

acorncom:
Homu appears to be broken these days ...

Issue #1185, opened 2016-01-21
Improve search results on Google


acorncom:
Currently, all Google links point to `/v1.10.0` guides instead of more recent guides. In order to fix that, we need to work out how to have a guides page that Google ends up looking at as more "authoritative" (probably by keeping the link the same).  Additional context and discussion is available here: #604

Once we've worked through that, we'll also need to inform Google that the v1.10.0 guides are not the most current guides (probably with a `<canonical>` tag put in the header of the old files).

We need to decide what our `<canonical>` url for guides should be.  We seem to have two options:
1. Have a `/current` link where the `master` version of guides always lives.
2. Have current guides live under `guides.emberjs.com` with no version number specified. This is the approach taken by the Ruby on Rails folks.

In either of these cases, we will also want to publish the guides at a static version number so that blog post authors can link to versioned guide pages without having to worry about those pages changing later.

Feedback about pros / cons to these approaches would be quite welcome.  What issues / problems might we run into from an SEO point of view?

robbyrussell:
Perhaps we can lean on what the [Read the Docs](https://readthedocs.org/) folks are doing. It appears that they're doing something to option 1 on previous versions always referencing the latest as the canonical.

For example, looking at [Sphinx 1.3.5 docs](http://www.sphinx-doc.org/en/1.3.5/), I see the following:

```
<!-- 
Always link to the latest version, as canonical.
http://docs.readthedocs.org/en/latest/canonical.html
-->
<link rel="canonical" href="http://www.sphinx-doc.org/en/stable/" />
```

Found this after looking through http://docs.readthedocs.org/en/latest/versions.html

tonycoco:
Agree that option #1 seems the most legit. Plus, in reading http://read-the-docs.readthedocs.org/en/latest/versions.html#redirects-on-root-urls they recommend that a redirect is placed in the root to the "latest" doc branch.

tomdale:
@robbyrussell One potential problem with the `<link rel="canonical">` approach is that it makes it difficult for users of an older version.

I'll give you a concrete example from my own experience. I've been writing Node.js apps that target AWS Lambda, which only supports Node 0.10. The latest release of Node is v5, so if I search for e.g., `node child_process spawn` on Google, it gives me the Node v5 docs.

[...]

PR #1530, opened 2016-06-23
[ON HOLD] Add auto-transition to newest docs on initial load

### LINKED ISSUE #1505: Request: cookie to know I always want the newest docs
I know this is a known problem, but whenever I get to guides from google, they are always the out-of-date ones.

Could you guys add a cookie that remembers that I always want to be transtionedTo the newest docs?

djsegal:
Resolve #1505, if not a little more aggressively than described before.

`AutoTransition` works as follows:
- Only happens on initial page entries (i.e. from google and not from internal redirects)
- Transitions the docs to the newest version (if it would normally have the red redirect button shown)

---

To make it a little less aggressive, 
- there could be a cookie that remembers if a user wants the newest docs 
- some people might want to default to old docs, like one's with them in their bookmarks

### DIFF
--- a/source/javascripts/app/auto-transitions.js
+++ b/source/javascripts/app/auto-transitions.js
@@ -1,0 +1,81 @@
+$(document).ready(function() {
+
+  window.localLinkClicked = false;
+
+  var cookieName = 'isReadingGuides';
+
+  createCookie = function(name, value, days) {
+    if ( !days ) { days = 100; }
+
+    var date = new Date();
+    date.setTime(date.getTime() + (days*24*60*60*1000));
+
+    var expires = '; expires=' + date.toGMTString(),
+        newCookie = name + '=' + value + expires + '; path=/';
+
+    document.cookie = newCookie;
+  };
+
+  readCookie = function(name) {
+    var nameSubstring = name + '=',
+        cookieArray = document.cookie.split(';'),
+        arrayLength = cookieArray.length,
+        nameSubstringLength = nameSubstring.length;
+
+    for (var i=0; i < arrayLength; i++) {
+      var curCookie = cookieArray[i],
+          curCookieLength = curCookie.length;
+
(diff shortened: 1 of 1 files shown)


djsegal:
@acorncom do you have any feedback on how to move forward?

djsegal:
@locks do you have any input on this?

acorncom:
@djsegal Sorry about the delay in response here.  We're aiming to have guides published both at specific version numbers and also at `/current`  The frustrations of not having Google indexing only the most current pages (and ending up on old pages) is something all of us experience (more background on the research I've done on this here: https://github.com/emberjs/guides/issues/1185)

If we add those two things, do you see any need for this as well?  It's a reasonable short-term fix, but I'm concerned that it would just add additional complexity to something that's already painful to maintain ...

djsegal:
Sorry, but what do you mean by two things? Adding:
+ `/current` route
+  `___what?___`

---------

but yea, i don't want to add unneeded complexity to the code. if you think in the long-term `/current` will beat out over `/v1.10.0`, i'm all for going that route

it's just good to have this alternate solution in the bag if the problem ever resurfaces

acorncom:
Sorry, missed your comment here.  Re: `/current` vs `/v1.10.0`, yes, the goal is to make all the Google links point to the very most current version of Ember (and have folks shift to old versions if they need to).  A brief explanation of this can be found here: https://github.com/emberjs/guides/issues/1584#issuecomment-236257773 along with links to additional background information.

For now, I think I'll re-label this to `[ON HOLD]` as we work on the other stuff and then we can revisit in a month or two when the other changes are in.  If you think the above items will deal with your problem, then feel free to close this PR in the meantime ...

djsegal:
@acorncom any word on this?

djsegal:
Closing this. Out of date

PR #1532, opened 2016-06-24
Use the `find` helper

yawboakye:
Apart from introducing it, it's less primitive than `this.$`

### DIFF
--- a/source/localizable/tutorial/acceptance-test.md
+++ b/source/localizable/tutorial/acceptance-test.md
@@ -82,7 +82,7 @@ To check that rentals are listed, we'll first visit the index route and check th
 test('should list available rentals.', function (assert) {
   visit('/');
   andThen(function () {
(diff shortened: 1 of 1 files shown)


rwjblue:
👍 - I am not sure that `this.$` is even a thing in acceptance tests (Ember doesn't set it up and I don't think `moduleForAcceptance` does either).

toddjordan:
@rwjblue it is a thing, works in super-rentals: https://github.com/emberjs/super-rentals/blob/master/tests/acceptance/list-rentals-test.js

That said, I am in favor of this change....

@yawboakye can you also PR the super-rentals repo?

yawboakye:
I remember you talked about that as the requirement for changing code in the tutorial guides so I did both at the same time :smile: 
Here's the PR on super-rentals: https://github.com/emberjs/super-rentals/pull/30

toddjordan:
@homu r+

Issue #1365, opened 2016-04-09
Clarify recommended places to store assets


MarkMT:
In [Managing Dependencies](https://guides.emberjs.com/v2.4.0/addons-and-dependencies/managing-dependencies/), under the h2 heading 'Other Assets', it says -

> Assets not available as an addon or Bower package should be placed in the vendor folder in your project.

Later on the same page there's an h3 heading with the same name, 'Other Assets' (and the same link url) with this -

> All other assets like images or fonts can also be added via import(). By default, they will be copied to dist/ as they are.

Yet the [Ember-CLI documentation](http://ember-cli.com/asset-compilation) states that raw assets should go in `public/assets`, which is what I actually use. Presumably the advice above is correct for the particular situations they're meant to refer to, but the distinction isn't clear to me and as far as I've been able to find, the use of `public/assets` isn't mentioned in the ember.js guides at all.

acorncom:
@locks comments here?

locks:
The usual distinction is that `vendor` -> third-party assets, `public` -> your assets. That's basically what I have to say on the subject :P

acorncom:
@MarkMT Is that discussion above clear enough?  Would you want to write something up for the guides on this?

PR #1536, opened 2016-06-27
[Doc clarification] Update recommended place to store assets

green-arrow:
Addresses #1365 

Let me know if there's something you think needs any further clarification :)

### DIFF
--- a/source/localizable/addons-and-dependencies/managing-dependencies.md
+++ b/source/localizable/addons-and-dependencies/managing-dependencies.md
@@ -27,9 +27,9 @@ Thus it reloads your app if you install new dependencies via `bower install <dep
 
 ## Other assets
 
-Third-party assets not available as an addon or Bower package should be placed in the `vendor` folder in your project.
+Third-party JavaScript not available as an addon or Bower package should be placed in the `vendor/` folder in your project.
 
-Your own assets (such as `robots.txt` and your `favicon`) should be placed in the `public` folder in your project.
+Your own assets (such as `robots.txt`, `favicon`, custom fonts, etc) should be placed in the `public/` folder in your project.
 
 ## Compiling Assets
 
@@ -111,13 +111,15 @@ All style assets added this way will be concatenated and output as `/assets/vend
 
 ### Other Assets
 
-All assets located in the `public` folder will be copied as is to the `dist/` folder.
+All assets located in the `public/` folder will be copied as is to the final output directory, `dist/`.
 
 For example, a `favicon` located at `public/images/favicon.ico` will be copied to `dist/images/favicon.ico`.
 
-All other third-party assets like images or fonts can also be added via `import()`.
-By default, they
-will be copied to `dist/` as they are.
+All third-party assets, included either manually in `vendor/` or via a package manager like Bower, must be added via `import()`.
+
+Third-party assets that are not added via `import()` will not be present in the final build.
+
+By default, `import`ed assets will be copied to `dist/` as they are, with the existing directory structure maintained.
 
 ```ember-cli-build.js
 app.import('bower_components/font-awesome/fonts/fontawesome-webfont.ttf');
--- a/source/localizable/addons-and-dependencies/managing-dependencies.md
+++ b/source/localizable/addons-and-dependencies/managing-dependencies.md
@@ -27,7 +27,9 @@ Thus it reloads your app if you install new dependencies via `bower install <dep
 
 ## Other assets
 
-Assets not available as an addon or Bower package should be placed in the `vendor` folder in your project.
+Third-party assets not available as an addon or Bower package should be placed in the `vendor` folder in your project.
+
+Your own assets (such as `robots.txt` and your `favicon`) should be placed in the `public` folder in your project.
 
 ## Compiling Assets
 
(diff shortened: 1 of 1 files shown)


source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -27,7 +27,9 @@ Thus it reloads your app if you install new dependencies via `bower install <dep
 
 ## Other assets
 
-Assets not available as an addon or Bower package should be placed in the `vendor` folder in your project.
+Third-party assets not available as an addon or Bower package should be placed in the `vendor` folder in your project.
```
tomdale: I think that this should say "Third-party JavaScript," not "Third-party assets." Are there any cases where someone would be using third-party non-JavaScript files that wouldn't be better served by being placed in `public/`?

source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -109,7 +111,11 @@ All style assets added this way will be concatenated and output as `/assets/vend
 
 ### Other Assets
 
-All other assets like images or fonts can also be added via `import()`.
+All assets located in the `public` folder will be copied as is to the `dist/` folder.
```
tomdale: Even though it's covered on this page, a new user skimming the documentation may not be clear about what the `dist/` folder is. Maybe this should mention that `dist/` is the final output directory. Also, for consistency, we should probably add a trailing slash to `public/`.

source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -109,7 +111,11 @@ All style assets added this way will be concatenated and output as `/assets/vend
...
+
+For example, a `favicon` located at `public/images/favicon.ico` will be copied to `dist/images/favicon.ico`.
+
+All other third-party assets like images or fonts can also be added via `import()`.
```
tomdale: This reads strangely to me because it says "all other third-party assets" comes after a discussion of assets that are explicitly _not_ third party. What does the other mean in this case?

It's not clear here whether this documentation is about files in `vendor/` or `public/`. Files in `public/` are included automatically, as you note. Files in `vendor/` **must** be `app.import()`ed. They won't work at all if you skip this step. We may want to state this in stronger language, and make it clear which rules apply to which directories.

TLDR:

1. Your own, non-JavaScript assets go in `public/` and are copied to the final output directory automatically. You should not (and indeed, cannot) `app.import()` anything in `public/`.
2. Third-party assets, _including_ JavaScript, go inside the `vendor/` directory. In order for a file to be included in the final build output, it _must_ be `app.import()`ed.

source/localizable/addons-and-dependencies/managing-dependencies.md
green-arrow: I've tried to think of a scenario where a non-bower / non-npm third-party asset might contain assets other than JS, but the only thing I can think of would be if they're using a font that's technically from a third-party. Even in that case though, I think the asset would be better served in `public`.

source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -27,7 +27,9 @@ Thus it reloads your app if you install new dependencies via `bower install <dep
 
 ## Other assets
 
-Assets not available as an addon or Bower package should be placed in the `vendor` folder in your project.
+Third-party JavaScript not available as an addon or Bower package should be placed in the `vendor/` folder in your project.
```
green-arrow: Updated to 'Third-party JavaScript' instead of 'Third-party assets'.

source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -109,9 +111,15 @@ All style assets added this way will be concatenated and output as `/assets/vend
 
 ### Other Assets
 
-All other assets like images or fonts can also be added via `import()`.
-By default, they
-will be copied to `dist/` as they are.
+All assets located in the `public/` folder will be copied as is to the final output directory, `dist/`.
```
green-arrow: Updated to define exactly what `dist/` is (final output directory).

source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -109,9 +111,15 @@ All style assets added this way will be concatenated and output as `/assets/vend
...
+
+For example, a `favicon` located at `public/images/favicon.ico` will be copied to `dist/images/favicon.ico`.
+
+All third-party assets, included either manually in `vendor/` or via a package manager like Bower, must be added via `import()`.
```
green-arrow: Updated phrasing to more strongly state that third-party assets need to be included via `import`, otherwise they won't make the build.

source/localizable/addons-and-dependencies/managing-dependencies.md
```diff
@@ -109,9 +111,15 @@ All style assets added this way will be concatenated and output as `/assets/vend
...
+
+Third-party assets that are not added via `import()` will not be present in the final build.
```
green-arrow: I think this may need a clarification that addons / NPM modules don't need the `import`, and this is only for things like bower packages and stuff in `vendor`. The previous sentence may be enough context, but I'm not sure. Thoughts?

tomdale:
@green-arrow Awesome improvements, thank you!

tomdale:
@emberjs/learning-team-managers LGTM, WDYT?

green-arrow:
@tomdale thanks for the feedback!

locks:
👌 

@homu r+

Issue #1213, opened 2016-02-03
Document how dasherized model names work in relationships / store calls


acorncom:
Some folks get confused without clarification on this (see https://github.com/emberjs/guides/issues/1192).  Suggested locations:

- `models/finding-records` (in addition to our `post` we might want to alternate with `blog-post`)
- `models/relationships` (again, might be worth changing `post` to `blog-post` under One to Many)

alexwilkinson:
I think this issue can be resolved now! : D

PR #1539, opened 2016-06-27
[Documentation] Dasherized Model Names

alexwilkinson:
Added documentation around dasherized model names in the Models section based on suggestions from https://github.com/emberjs/guides/issues/1213. Let me know if there are more places in the documentation we could add this kind of thing!


tomdale:
This looks awesome! We should also take a look at the API docs. Any API that is used in these guides probably should have the API docs audited.

source/localizable/models/relationships.md
```diff
@@ -43,7 +43,7 @@ import Model from 'ember-data/model';
 import { belongsTo } from 'ember-data/relationships';
 
 export default Model.extend({
-  post: belongsTo('post')
+  blogPost: belongsTo('blog-post')
```
acorncom: If we change this to `blogPost` from `post`, then we may want to rename all other cases of `blogPost` on this page, which are a ton ...

alexwilkinson: Yeah I guess it depends on whether we consider each paired example as isolated from the others and you can vary up file names to illustrate different points, or if the re-use of the same same example files throughout a chapter is illustrative by telling an implied narrative.

I think there is a lot of value to having this kind of example, to the point where changing all instances may be worth it if we think it's important to be consistent throughout the page. But I think it could be alright to vary up file names between individual examples.

locks: Having `post` and `blog-post` differ from example to example is needlessly confusing, and will trip people up, so it needs to be consistent.
The Guides do have an implied narrative, we try to lessen the cognitive overload by trying to make sure stuff is in the same "universe".

alexwilkinson: That makes sense! I will push up an update that changes `post` to `blog-post` throughout the whole page.

Thanks for the feedback guys!

alexwilkinson:
@homu Will do! I'll get back to this work this week, been abnormally busy.

alexwilkinson:
@acorncom @locks I just pushed an update to this branch that globally changes post -> blogPost in the two guides chapters that I edited. For url format, I followed the recommended of the JSON API spec to use hyphens to separate multiple words http://jsonapi.org/recommendations/#naming.

Let me know if there's anything else I should adjust!

yawboakye:
@alexwilkinson As it stands now I think only the merge conflict is preventing a merge. If you still have time, can you resolve it? Thank you :smile:

source/localizable/models/finding-records.md
```diff
@@ -8,15 +8,15 @@ to retrieve a record by its type and ID. This will return a promise that
...
+var blogPost = this.get('store').findRecord('blog-post', 1); // => GET /blog-posts/1
 ```
 
 Use [`store.peekRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_peekRecord)
 to retrieve a record by its type and ID, without making a network request.
 This will return the record only if it is already present in the store:
 
 ```javascript
-var post = this.get('store').peekRecord('post', 1); // => no network request
+var blogpost = this.get('store').peekRecord('blog-post', 1); // => no network request
```
yawboakye: @alexwilkinson `blogPost` instead of `blogpost`

alexwilkinson: @yawboakye nice catch thanks!

alexwilkinson:
@yawboakye Resolved the merge conflicts!

locks:
Thanks for your contribution!

Issue #1541, opened 2016-06-27
Typo in Building a simple component tutorial


dhiachou:
I am sure that the line 153 of the file `source/localizable/tutorial/simple-component.md` should become : 
```
Since we want the image to be smaller at first, we will set the property to start as `false`:
``` 
instead of 
```
Since we do not want the image to be smaller at first, we will set the property to start as `false`:
````

yawboakye:
Fixed in https://github.com/emberjs/guides/pull/1543. Issue can be closed.

toddjordan:
Thanks @yawboakye and @dhiachou for catching/fixing!  Closing as fixed by #1543.

PR #1543, opened 2016-06-27
Fix typo

yawboakye:
Addresses https://github.com/emberjs/guides/issues/1541


source/localizable/tutorial/simple-component.md
```diff
@@ -150,7 +150,7 @@ Let's use the `{{#if}}` helper to show our current rental image larger only when
 ```
 
 The value of `isWide` comes from our component's JavaScript file, in this case `rental-listing.js`.
-Since we do not want the image to be smaller at first, we will set the property to start as `false`:
+Since we do want the image to be smaller at first, we will set the property to start as `false`:
```
acorncom: Since we **want** the image

acorncom:
One minor comment and then let's merge this, thanks!

locks:
@homu r+

PR #1545, opened 2016-06-29
`todos.js` filename is not a valid component name

patocallaghan:
Trying to create a component with the name `todos` will generate the following error: `You specified "todos", but in order to prevent clashes with current or future HTML element names, you must include a hyphen in the component name.`. 

I've changed it to be `todo-list.js`


source/localizable/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -4,7 +4,7 @@ to calculate the incomplete todo's based on their `isDone` property.
 
 To facilitate this, Ember provides the `@each` key illustrated below:
 
-```app/components/todos.js
+```app/components/todos-list.js
```
acorncom: Could we change it to `todo-list` instead?

source/localizable/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -46,7 +46,7 @@ export default Ember.Component.extend({
 In both of the examples above, `incomplete` is an array containing the single incomplete todo:
 
 ```javascript
-import TodosComponent from 'app/components/todos';
+import TodosComponent from 'app/components/todos-list';
```
acorncom: Seems like it would be cleanest to change this line (and the lines beneath) to `TodoListComponent` as well ...

patocallaghan:
@acorncom Updated!

locks:
@homu r+

PR #1547, opened 2016-06-29
Explain what `ember install` does

yawboakye:
Just so the learner doesn't assume that Ember maintains its own package
manager and hosting. Also, remind them to restart their server, it case
it was running.


source/localizable/tutorial/installing-addons.md
```diff
@@ -18,7 +18,11 @@ Run the following command to install the addon:
 ember install ember-cli-tutorial-style
 ```
 
-Starting the server will incorporate the new CSS and refreshing our browser window will give you this:
+Since Ember addons are regular npm packages, `ember install` installs them in
```
toddjordan: Not quite "regular" as there is a bit more to them, such as blueprints.  Maybe phrase it, "Since Ember addons are npm packages, Running `ember install` puts them in the `node_modules` directory..."

source/localizable/tutorial/installing-addons.md
```diff
@@ -18,7 +18,11 @@ Run the following command to install the addon:
...
+the `node_modules` directory, and makes an entry in `package.json`. Be sure to
```
toddjordan: See CONTRIBUTING.md for line break guidance: 

When writing prose:

Use a single space after periods.
Manually break lines:
at the beginning of every new sentence;
after the last comma in sentences over 120 characters; or
after logical statements in sentences over 120 characters with no commas.

source/localizable/tutorial/installing-addons.md
yawboakye: @toddjordan I intended regular to mean installable via `npm install`, but, sure, if that's not coming across I'll make the changes.

source/localizable/tutorial/installing-addons.md
yawboakye: :+1:

locks:
If we're going to explain `ember install`, we should explain what it actually does, and why it's preferred over `npm install`.

yawboakye:
@locks Won't the Addons and Dependencies be the best place for all the details of `ember install`? I can update the PR to include more information but I want to be sure we won't be duplicating effort, or putting the content in the wrong place.

yawboakye:
@locks I'd love to complete any work remaining on this PR. What do you think of my suggestion to leave the details to Addons and Dependencies?

toddjordan:
@yawboakye I'm fine with explaining addons in more depth in the addons and dependencies page, and then linking back from this page.  (feel free to pick that up if you want ;-)  Otherwise I think this is good to go once you squash commits.

PR #1549, opened 2016-07-01
Fix an issue with the Google Maps service tutorial

frqnck:
Name collision between 2 variables in `pinLocation()` produced `gmaps.js:32 InvalidValueError: setTitle: not a string`.


toddjordan:
Awesome, thanks @frqnck for contributing!

PR #1550, opened 2016-07-01
[Documentation] Improve doc to findRecord/findAll

villander:
Guides page does not explain that findAll/findRecord searches the cache (store) first

### DIFF
--- a/source/localizable/models/finding-records.md
+++ b/source/localizable/models/finding-records.md
@@ -4,8 +4,7 @@ type.
 ### Retrieving a Single Record
 
 Use [`store.findRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_findRecord)
-to retrieve a record by its type and ID. This will return a promise that
-fulfills with the requested record:
+to retrieve a record by its type and ID. Ember Data will check if the record has already been loaded into store. If it is not loaded it will make a request to the Adapter. This will return a promise that fulfills with the requested record:
 
 ```javascript
 var post = this.get('store').findRecord('post', 1); // => GET /posts/1
@@ -22,7 +21,7 @@ var post = this.get('store').peekRecord('post', 1); // => no network request
 ### Retrieving Multiple Records
 
 Use [`store.findAll()`](http://emberjs.com/api/data/classes/DS.Store.html#method_findAll)
-to retrieve all of the records for a given type:
+to retrieve all of the records for a given type. Ember Data will check if all record has already been loaded into store. If they are not loaded it will make a request to the Adapter:
 
 ```javascript
 var posts = this.get('store').findAll('post'); // => GET /posts


source/localizable/models/finding-records.md
```diff
@@ -4,8 +4,7 @@ type.
 ### Retrieving a Single Record
 
 Use [`store.findRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_findRecord)
-to retrieve a record by its type and ID. This will return a promise that
-fulfills with the requested record:
+to retrieve a record by its type and ID. Ember Data will check if the record has already been loaded into store. If it is not loaded it will make a request to the Adapter. This will return a promise that fulfills with the requested record:
```
yawboakye: `findRecord` always returns a promise. If the record is already in the store then how long it takes the promise to resolve depends on the reload behavior. If the record is not already loaded, the store (not Ember Data) calls the appropriate adapter's `find` method.

villander:
@yawboakye you are very wrong look this: `In Ember Data, the store is responsible for managing the lifecycle of your models. Every time you need a model or a collection of models, you'll ask the store for it.`

https://github.com/emberjs/data#instantiating-the-store

not exist `store` without `ember-data`

villander:
@locks look this PR please.
@grapho too

yawboakye:
@villander Saying that Ember Data send a request to the Adapter insinuates that they're different systems, but they are not. Adapter is part of Ember Data. There's also Serializer, Model, etc. And the `store`, which has the `findRecord` and `findAll` methods in discussion. What calls Adapter's `find` method is the store (one part of Ember Data) not Ember Data (the whole system).

grapho:
@yawboakye  and @villander  are both correct.  findAll does always return a promise, there are conditions that will make that promise resolve "instantly" from store cache.  depending on your reload/backgroundReload adapter settings.. an http request will be made.

I agree that the guides should explain this process, at least in a small part.. with a link of two the the API docs

toddjordan:
Since the behavior around `findRecord` can be a bit more subtle than this if you take into account reloading and background reloading options.  A programmer can rely on the returned promise to access the found (or not found) record, regardless of the caching logic, so I think we are safe as-is.

If you want to add mention of caching, we should just mention simply that the behavior exists, and link to the explanation in the docs.  Something like:
"Ember Data leverages a caching strategy when retrieving records with `find`. Refer to the 'findRecord` API documentation to learn more about caching options and behavior

Also, it would be good to give a small example of accessing the results of the find call, and separate the inline comments out onto a new line (its being cut off by the the screen)

locks:
As @toddjordan says, the problem is more nuanced and I don't want to go for broke at this particular juncture :P
We already link to the API documentation, which contains said nuanced explanation so I'm closing this issue.
Feedback appreciated!

PR #1551, opened 2016-07-02
Improve doc find records

villander:
Guides page does not explain that findAll/findRecord searches the cache (store) first


villander:
@yawboakye

PR #1552, opened 2016-07-03
Explain how computed properties work

yawboakye:
The text has been corrected to say that the function backing the
computed property is run once,  and the resulting value is cached.
Subsequent calls read from the cache, until any of the dependent
properties change, where the function has to be run again to update the
cache.

### DIFF
--- a/source/localizable/object-model/computed-properties.md
+++ b/source/localizable/object-model/computed-properties.md
@@ -19,7 +19,7 @@ Person = Ember.Object.extend({
   })
 });
 
-var ironMan = Person.create({
+let ironMan = Person.create({
   firstName: 'Tony',
   lastName:  'Stark'
 });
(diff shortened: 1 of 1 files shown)


source/localizable/object-model/computed-properties.md
```diff
@@ -19,22 +19,24 @@ Person = Ember.Object.extend({
...
 ironMan.get('fullName'); // "Tony Stark"
 ```
 
-This declares the function to be a computed property, and the arguments tell Ember that it depends on the `firstName` and `lastName` attributes.
-
-Whenever you access the `fullName` property, this function gets called, and it returns the value of the function, which simply calls `firstName` + `lastName`.
+This declares `fullName` to be a computed property, with `firstName` and `lastName` as the properties it depends on. The
+first time you access the `fullName` property, the function backing the computed property (i.e. the last argument) will
+be run and the results will be cached. Subsequent access of `fullName` will read from the cache without calling the
+function.  Only when any of the dependent properties change will the function be run again, and the cache updated with
```
toddjordan: Instead of 'Only when any of the dependent properties change will the function be run again, and the cache updated with the new value.', say 'Changing dependent properties cause the cache to invalidate, so that the computed function runs again on the next access'

I think it works that way more than just running the function whenever the properties change (which is one reason why these are better than observers)

yawboakye:
@toddjordan Updated with your suggestion, and commits squashed.

toddjordan:
Looks great. Thanks @yawboakye !

toddjordan:
@homu r+

source/localizable/object-model/computed-properties.md
```diff
@@ -19,22 +19,24 @@ Person = Ember.Object.extend({
...
   lastName:  'Stark'
 });
 
 ironMan.get('fullName'); // "Tony Stark"
 ```
 
-This declares the function to be a computed property, and the arguments tell Ember that it depends on the `firstName` and `lastName` attributes.
-
-Whenever you access the `fullName` property, this function gets called, and it returns the value of the function, which simply calls `firstName` + `lastName`.
+This declares `fullName` to be a computed property, with `firstName` and `lastName` as the properties it depends on. The
```
locks: Every period should be followed by a newline, to make translations and future changes easier.

PR #1557, opened 2016-07-06
Indent these snippets?

sheriffderek:
Suggestion to indent these because the fall off screen even at full width.


locks:
Thanks for your contribution!

@homu r+

PR #1559, opened 2016-07-11
Improve font rendering in code examples

oskarrough:
This mini PR removes Courier from the monospace font-stack used for code examples in the guide. Not because Courier is bad but because Courier renders terribly on Fedora. I suspect other Linux distro as well.

Let's use what the system recommends. I left Menlo in there as I thought it might be defined in a styleguide?

Courier on Fedora

screenshot from 2016-07-11 21-01-26

`monospace` on Fedora

screenshot from 2016-07-11 21-01-42

### DIFF
--- a/source/stylesheets/base/_variables.scss
+++ b/source/stylesheets/base/_variables.scss
@@ -1,7 +1,7 @@
 // Typography
 $base-font-family: $helvetica;
 $heading-font-family: $base-font-family;
-$monospace-font-family: Menlo, Courier, monospace;
+$monospace-font-family: Menlo, monospace;
 $footer-font-family: proxima-nova, sans-serif;
 
 // Font Sizes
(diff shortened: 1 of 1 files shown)


acorncom:
@oskarrough Is there a Courier equivalent on Fedora / Linux that is commonly installed that we can specify before Courier?  Courier is quite common on Macs (Menlo isn't) and, if I remember right, looks far better than the default monospaced font.  Would love to help fix your issue without making things worse for everyone else ;-)

oskarrough:
Good point.

How about `Menlo, "DejaVu Sans Mono", "Bitstream Vera Sans Mono", Courier, monospace;`? That would be two default fonts on most Linux system. And users would still get Courier before system defaults.

acorncom:
Works for me, send it on over

oskarrough:
Hi, sorry for the delay. I've added the two extra "linux typefaces" after Menlo and before Courier now.

locks:
Thanks for your contribution!

@homu r+

PR #1565, opened 2016-07-18
Use ES6 constructs

yawboakye:
For some of the code examples, it reduced them to one-liners, which is a
good thing. I also changed the code example to mark expected results
with `//=>` where it made sense, but more importantly to maintain
consistency.


locks:
Hm, I've seen many people confused by the implicit return version of arrow functions. I wonder if we should apply this change at the risk of alienating newbies?

yawboakye:
@locks Correct me if I'm wrong but there's a link to an ES6 tutorial in the Guides? I was fairly new to ES6 when I started Ember, and I hadn't read any material or deep-dive into ES6 (not even what the Guides pointed to); most of what I know, I learnt from the examples in the Guides, the Ember code base, etc. If we were to vote I'd vote for going full throttle in the Guides so that it doubles also as a resource for learning ES6. Don't worry, our minds will learn.

toddjordan:
While I agree that arrow functions can confuse people coming in without any es6 knowledge, I think its fair to keep the es6 construct.  We specify it as an assumption to using the guides: https://guides.emberjs.com/v2.7.0/#toc_assumptions

I followed the es6 guide link there to functions and it was one of the first examples given for arrow functions.

PR #1567, opened 2016-07-20
Update service.md

ilkon:
Adding missing import. Without it the test fails.


locks:
Thanks for your contribution!

@homu r+

PR #1569, opened 2016-07-21
Reword notes about asynchrony

sheriffderek:
I'm not convinced that my wording is ideal, but I wanted to point out that it wasn't very clear to me - what was being said.

_Side note_: the line numbers don't seem to be helpful, and in many cases - (take the 2 tests on this page) are actually more confusing. It's hard to tell where that second test goes since it should be line 45 based on the previous snippet. Also - it's valuable space. Anyway... didn't see a good way to mention that.

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -16,10 +16,9 @@ Let's start with writing some tests to help us think through what we are doing.
 The filter component should yield a list of filtered items to whatever is rendered inside of it, known as its inner template block.
 We want our component to call out to two actions: one action to provide a list of all items when no filter is provided, and the other action to search listings by city.
 
-For our initial test, we'll simply check that all the cities we provide are rendered and that the listing object is accessible from the template.
+For our initial test, we can check that all the cities we provide are rendered and that the listing object is accessible from the template.
 
-Since we plan to use Ember Data as our model store, we need to make our action calls to fetch data asynchronous, so we'll return promises.
-Because accessing persisted data is typically done asynchronously, we want to use the wait helper at the end of our test, which will wait for all promises to resolve before completing the test.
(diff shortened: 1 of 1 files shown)


source/localizable/tutorial/autocomplete-component.md
```diff
@@ -16,10 +16,9 @@ Let's start with writing some tests to help us think through what we are doing.
...
 We want our component to call out to two actions: one action to provide a list of all items when no filter is provided, and the other action to search listings by city.
 
-For our initial test, we'll simply check that all the cities we provide are rendered and that the listing object is accessible from the template.
+For our initial test, we can check that all the cities we provide are rendered and that the listing object is accessible from the template.
 
-Since we plan to use Ember Data as our model store, we need to make our action calls to fetch data asynchronous, so we'll return promises.
-Because accessing persisted data is typically done asynchronously, we want to use the wait helper at the end of our test, which will wait for all promises to resolve before completing the test.
+Since we plan to use Ember Data as our model store, and accessing persisted data is typically done asynchronously, we'll need to be sure of a few things. Our action calls (to fetch data) need to be asynchronous. To ensure this, we can return promises, and then we can use the wait helper at the end of our test. This will wait for all promises to resolve before completing the test.
```
yawboakye: I'll replace *can* with *will*. In the tutorial we describe what we will do specifically.

yawboakye:
@sheriffderek What do you think of the changes I suggested? I think we should move quickly to get this update merged in.

source/localizable/tutorial/autocomplete-component.md
toddjordan: I think we should add links to these concepts so readers can get a better idea of what's being talked about.  

Also we probably don't need to mention ember data in this context.

How about this:
```
Our action call to filter by city will be made asynchronously and we will have to accommodate for this in our test.  
[We will leverage Ember closure actions here to handle asynchronous action completion](../../components/triggering-changes-with-actions/#toc_handling-action-completion) from our `filterByCity` call by returning a promise from our stubbed action.  

Note that we also need to add a `wait` call at the end of our test to assert the results.  [Ember's `wait` helper](../../testing/testing-components/#toc_waiting-on-asynchronous-behavior) waits for all promises to resolve before running the given function callback and finishing the test.
```

sheriffderek:
@yawboakye Sounds better.

locks:
Manually merged. Thanks everyone involved :)

Issue #1498, opened 2016-06-15
"Creating your app" step is a little outdated


jzaefferer:
At the end of https://guides.emberjs.com/v2.6.0/tutorial/ember-cli/ the guide says: "If we navigate to localhost:4200, we'll see our brand new app displaying the text "Welcome to Ember"."

Instead, in Ember(-cli) 2.6.0, it shows the Ember mascot with a hardhat and "Congratulations, you made it!" and some more text. Also `app/templates/application.hbs` is gone (the quickstart guide relies on that file).

PS: Hope someone appreciates these tickets, there's probably more coming as I make my way through the tutorial.

acorncom:
@jzaefferer Thanks, it's something we've known was in there, but hadn't had a chance to fix yet.  Would you be able to send in a PR with a note about creating the application.hbs file?

And yes, all tickets are helpful, PRs are even better ;-)

jzaefferer:
So ember-cli doesn't create that template anymore, but the user should create it as part of the tutorial?

acorncom:
Correct.  There is a new add-on that we added to Ember-CLI (https://github.com/ember-cli/ember-welcome-page).  As part of that, the standard `templates/application.hbs` was removed from Ember-CLI's standard blueprint.

acorncom:
This was closed by the PR referenced above ...

PR #1570, opened 2016-07-22
Update "Creating your app" section of the tutorial to account for the new welcome page.

toddjordan:
Address #1498


source/localizable/tutorial/ember-cli.md
```diff
@@ -131,5 +131,7 @@ or, for short:
 ember s
 ```
 
-If we navigate to [`http://localhost:4200`](http://localhost:4200), we'll see our brand new app displaying
-the text "Welcome to Ember".
+If we navigate to [`http://localhost:4200`](http://localhost:4200), we'll see the default welcome screen.
+Once we add our own `app/templates/application.hbs` file, the welcome screen will be replaced with our own content.
```
toddjordan: @acorncom @locks should we add an additional mention that creating a route makes it go away too?

locks: we should chance the message on the welcome addon itself

PR #1576, opened 2016-07-25
Explain conditions that trigger the `model` hook

yawboakye:
For routes with dynamic segments, the `model` hook is not executed at all times. This change explains what conditions will or will not cause an execution.

### DIFF
--- a/source/localizable/routing/specifying-a-routes-model.md
+++ b/source/localizable/routing/specifying-a-routes-model.md
@@ -25,10 +25,9 @@ export default Ember.Route.extend({
 ```
 
 Typically, the `model` hook should return an [Ember Data](../../models/) record,
-but it can also return any [Promise](https://www.promisejs.org/) object (Ember
-Data records are promises), or a plain JavaScript object or array. Ember will
-wait until the data finishes loading (until the promise is resolved) before
-rendering the template.
+but it can also return any [promise](https://www.promisejs.org/) object (Ember Data records are promises),
+or a plain JavaScript object or array.
+Ember will wait until the data finishes loading (until the promise is resolved) before rendering the template.
 
 The route will then set the return value from the `model` hook as the `model` property of the controller.
 You will then be able to access the controller's `model` property in your template:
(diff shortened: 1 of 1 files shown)


locks:
I didn't like the "Note:" in the first place, much less one that's this extensive. I like the sentiment @yawboakye, but I think a bigger restructure of the section is needed. I'll see what I can come up with and ping you.

source/localizable/routing/specifying-a-routes-model.md
```diff
@@ -9,8 +9,8 @@ Router.map(function() {
 });
 ```
 
-To load a model for the `favoritePosts` route, you would use the [`model()`][1]
```
locks: can you send this change separately?

source/localizable/routing/specifying-a-routes-model.md
```diff
@@ -23,7 +23,7 @@ export default Ember.Route.extend({
 ```
 
 Typically, the `model` hook should return an [Ember Data](../../models/) record,
-but it can also return any [promise](https://www.promisejs.org/) object (Ember
```
locks: I don't think this change is correct

yawboakye: I had to capitalize because it said *promise object*. I thought of it along the same lines as, say, an `EmberObject` object. If it said "but it can also return a promise" then I think it could stay that way.

source/localizable/routing/specifying-a-routes-model.md
yawboakye: https://github.com/emberjs/guides/pull/1577

source/localizable/routing/specifying-a-routes-model.md
locks: But it's not a `Promise` object as in the `Promise` class, it's an object that's also a promise. I don't think I've cleared anything up xD

yawboakye: How about we say promise instead of promise object? What do you think about that @locks?

locks:
@homu r+

PR #1578, opened 2016-07-25
Update guides and tutorial to use the import DS from 'ember-data'

bmac:
I have an open https://github.com/emberjs/data/pull/4481 on the ember data repo to change the generators back to use the `import DS from 'ember-data'` style. This pr updates the guides to match those generators.


locks:
What version of Ember Data is the revert going to be released?

bmac:
Ember Data 2.7.0. I have also back-ported the changes to Ember Data 2.5.4 and Ember Data 2.6.2.

PR #1582, opened 2016-07-28
Recommend `templateName` as way to set template

yawboakye:
Since `renderTemplate()` is one of those methods you can't call
`this._super()`, `templateName` is a cleaner way to change a route's
default template. Also, we will automatically benefit from improvements
to `renderTemplate()` if we don't override it for the simple reason of
choosing a different template.

### DIFF
--- a/source/localizable/routing/rendering-a-template.md
+++ b/source/localizable/routing/rendering-a-template.md
@@ -1,5 +1,4 @@
-One job of a route handler is rendering the
-appropriate template to the screen.
+One job of a route handler is rendering the appropriate template to the screen.
 
 By default, a route handler will render the template with the same name as the
 route. Take this router:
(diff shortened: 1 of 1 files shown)


source/localizable/routing/rendering-a-template.md
```diff
@@ -20,15 +19,18 @@ template. For example, the `posts.new` route will render its template into the
...
-  renderTemplate() {
-    this.render('favoritePosts');
-  }
+  templateName: 'posts/favorite-posts'
 });
 ```
+
+You can override the [`renderTemplate()`][2] hook if you want finer control over template rendering.
+Among other things, it allows you to choose the controller used to configure the template and specific outlet to render
+it into.
```
locks: why the newline?

yawboakye: @locks @toddjordan mentioned that it helps translation if every period was followed by a newline. I assume you're talking about the newline following line 31.

locks: no, I'm talking about the newline between "to render" and "it into", which seems unnecessary.

yawboakye:
@locks Fixed, and fixed my editor too while I was at it.

locks:
@acorncom @toddjordan 👍 ?

acorncom:
@homu r+

acorncom:
Thanks @yawboakye

Issue #1561, opened 2016-07-14
Finish shift to Algolia search


acorncom:
We are wanting to continue our shift to Algolia search, so are opening this up to ask for help :-)  An example PR showing the code changes that were needed for versions 2.3-2.6 of the guides can be found here (https://github.com/emberjs/guides/pull/1433)

Remaining items to get resolved (ordered by priority):
- [x] shift main website search to Algolia (@acorncom - https://github.com/emberjs/website/pull/2644)
- [x] shift v1.13 (https://github.com/emberjs/guides/pull/1583)
- [ ] shift v2.2 (uses same styling as current search)
- [ ] shift v2.1
- [ ] shift v2.0
- [x] shift v1.12 (https://github.com/emberjs/guides/pull/1622)
- [x] shift v1.11 (https://github.com/emberjs/guides/pull/1623)
- [x] shift v1.10 (https://github.com/emberjs/guides/pull/1659)

Get the Git updates applied to the https://github.com/emberjs/guides.emberjs.com repo
- [x] shift v1.13 (https://github.com/emberjs/guides.emberjs.com/pull/22)
- [ ] shift v2.2
- [ ] shift v2.1
- [ ] shift v2.0
- [ ] shift v1.12
- [x] shift v1.11
- [ ] shift v1.10 (blocked by build issues, see https://github.com/emberjs/guides/pull/1750)

We made a style shift from 2.1 -> 2.2, so changes to search for versions 2.1 and older will need to account for potential changes to styling (and HTML changes in the layout files).

To send in updates to the older versions, there are two steps that need to be taken (doing step 1 is a big help, doing step 2 as well is a huge help):
1. Branch off of the appropriate branch for the version you are updating.  Make your changes and send in a PR **to that branch** here on the guides repo (please also include a screenshot or two showing how things look).  We'll then be able to update the website using that work.
2. Our actual website is a set of static files deployed to Firebase (files located here: https://github.com/emberjs/guides.emberjs.com).  We'll need to build the results for each version, process them per the steps in the Readme (but ignoring the search step) and send a PR in to the repo above with the built output.

Fun times 😬

acorncom:
Handled by the shift to the new guides Ember app

PR #1583, opened 2016-07-28
[1.13] Adding algolia search

acorncom:
Helps with #1561 

Before:
screen shot 2016-07-28 at 1 02 43 pm

After:
screen shot 2016-07-28 at 1 02 28 pm

/cc @locks @toddjordan 

Once we've merged this, I'll go back port to the older 1.x versions and then build all of them simultaneously to refresh the site


toddjordan:
LGTM

PR #1586, opened 2016-07-29
Remove the 'image' class from the img tag

cyrusstoller:
I was confused why my test wasn't passing, and then I realized that there was an extra `image` class on the `img` tag.

Thanks for the awesome guide.


acorncom:
To confirm, we're *adding* the image class tag right, not removing it? ;-)

cyrusstoller:
Just moving it, so that it's only on the `a` tag and not on both the `a` and `img` tags

toddjordan:
@cyrusstoller you might need to try this update out in the super-rentals repo https://github.com/emberjs/super-rentals to make sure the CSS is not affected.

soulchainer:
Seems like more people (as me xD) are still getting confused about this #1607. Didn't read this PR on time.

locks:
Fixed by https://github.com/emberjs/guides/pull/1607 I think?

PR #1587, opened 2016-07-31
Expand guide on route redirection

yawboakye:


source/localizable/routing/redirection.md
```diff
@@ -22,12 +21,20 @@ currently in progress and start a new one, functioning essentially as a redirect
 If the new route has dynamic segments, you need to pass either a _model_ or an _identifier_ for each segment.
 Passing a model will skip the route's `model()` hook (since the model is already loaded).
 
+## The [`replaceWith()`][3] method
```
yawboakye: I'm kind of disappointed that I couldn't draw up a simple situation where `replaceWith` is preferred to `transitionTo`; makes this part of the Guide resemble the API doc. Thing is, the example I know is too complicated and might end up extending this section for no good reason. If anyone has examples of when to use `replaceWith` please share with me and I'll update.
cc/ @locks @toddjordan @alexspeller

alexspeller:
Basically, replaceWith is useful for redirection, but only in the right circumstances. Say you have an `index` route, but you always want to redirect to `/products/browse` from the index.

If you used `transitionTo` there, then you'd break the back button - clicking back would get you back to `index` which would redirect to where you just were, instead of the previous site.

But you can't use it if there has already been a transition, because then it blows out the current state. That's why I always use [this addon](https://github.com/stefanpenner/ember-redirect-to/) to handle router hook redirects. I think this is not very common knowledge and most redirects in route hooks are subtly broken in real world apps.

source/localizable/routing/redirection.md
```diff
@@ -1,24 +1,44 @@
...
+
+For example, if they're not logged in, you might want to prevent them from editing their profile, accessing private information,
+or checking out items in their shopping cart.
+Usually you want to redirect them to the login page, and after they have successfully logged in, redirect them back to the page the wanted to access.
+Or they're logged in already so you don't want them to be able to visit the login page again.
+
+There's many other reasons you probably want to have the last word on whether or not a user can access a certain route.
+Whatever the reason, Ember allows you that control with a combination of hooks and methods in your route.
+
+## The [`transitionTo()`][1] method
```
locks: I'm not a fan of these headers, as they veer into API documentation-land.
Can we switch it out for something more use case-based like "Transitioning Before the Model is Known" does?

yawboakye: Btw, I'm rewriting a little bit of the content (using @alexspeller's input). I should push an update later today.

yawboakye: @locks Hard to come up with better headings so I'll take them off instead.

yawboakye:
Updating with your contribution, @alexspeller.

In other news I'm seeing a behavior I'm not familiar with, or rather, don't expect. But that's probably because I still don't get routing well. I'll describe below so that you can help me understand.

This is my router:

```javascript
// app/router.js

Router.map(function() {
  this.route('about');
  this.route('favorites', { path: '/favs' });
}
```
and in the `about` route I redirect to `favorites` using `replaceWith`:

```javascript
// app/routes/about.js

import Ember from 'ember';

  beforeModel() {
    this.replaceWith('favorites');
  }
});
```
And the user takes the following steps:
- Open a new tab in their browser
- Visit https://google.com
- Visit [http://localhost:4200](http://localhost:4200) (by entering the URL in the URL bar)
- Click on `About` link on the index page
  - *it redirects to `favorites` route correctly*

In the process it substitutes the history entry for the `index` route and replaces it with the entry for `favorites` route, so that when we transition to `http://localhost:4200/favs` we only have 2 entries in our history. The other one being Google. Thus, clicking the Back button takes you to Google instead of the homepage, which, I submit, is wrong.

But I'm more interested in whether this is the expected behavior and any documents I can read to help me understand it.

Regards.

alexspeller:
@yawboakye that's the exact issue I mentioned above - using `transitionTo` or `replaceWith` in the validation hooks is a bug, which is why you must use the addon I linked above if you want to do that

yawboakye:
@alexspeller When you say validation hooks do you mean `beforeModel` and `afterModel`? Because as it stands now we recommend doing redirections there.

alexspeller:
@yawboakye the validation hooks are `beforeModel`, `model` and `afterModel`. Really you shouldn't redirect in the setup hooks like `setupController` because at that point the transition has already taken place, but the same bug would crop up if you redirected there

source/localizable/routing/redirection.md
```diff
@@ -86,10 +100,10 @@ route's `beforeModel`, `model`, and `afterModel` hooks will fire again within
 the new, redirected transition. This is inefficient, since they just fired
 before the redirect.
 
-Instead, we can use the [`redirect()`][5] method, which will leave the original
+Instead, we can use the [`redirect()`][6] method, which will leave the original
 transition validated, and not cause the parent route's hooks to fire again:
 
-[5]: http://emberjs.com/api/classes/Ember.Route.html#method_redirect
+[6]: http://emberjs.com/api/classes/Ember.Route.html#method_redirect
```
toddjordan: Where's 1-5?

yawboakye:
@toddjordan :smile: Ah, I missed that one. So a while back @locks suggested we inline all the links we point to. That's why 1-5 were gone. I inline the link referenced by 6 as well. Please take another pass; should be good now.

source/localizable/routing/redirection.md
yawboakye: Inlined link identified by 6 instead. Now the numbering system is replaced with inline links, as @locks recommended. @toddjordan

source/localizable/routing/redirection.md
```diff
@@ -1,24 +1,34 @@
...
+Sometimes you want to redirect a user to a different page than what they requested for.
 
-[1]: http://emberjs.com/api/classes/Ember.Route.html#method_transitionTo
-[2]: http://emberjs.com/api/classes/Ember.Controller.html#method_transitionToRoute
+For example, if they're not logged in, you might want to prevent them from editing their profile, accessing private information,
+or checking out items in their shopping cart.
+Usually you want to redirect them to the login page, and after they have successfully logged in, take them back to the page they originally wanted to access.
+Another use case could be preventing a user who has already logged in from visiting the login page again.
+
+There's many other reasons you probably want to have the last word on whether a user can or cannot access a certain page.
```
locks: "There are"

locks: I think this sentence and the following could the condensed into something like "There are many reasons you might want to control user access to a certain page, and Ember provides APIs to do so in your route."

source/localizable/routing/redirection.md
```diff
@@ -1,24 +1,34 @@
...
+Whatever the reason, Ember allows you that control with a combination of hooks and methods in your route.
+
+One of the methods is [`transitionTo`](http://emberjs.com/api/classes/Ember.Route.html#method_transitionTo).
+Calling `transitionTo` from a route or
+[`transitionToRoute`](http://emberjs.com/api/classes/Ember.Controller.html#method_transitionToRoute) from a controller will stop any transitions currently in progress and start a new one, functioning as a redirect.
+`transitionTo()` behaves exactly like the [link-to](../../templates/links) helper.
+
+The other one is [`replaceWith()`](http://emberjs.com/api/classes/Ember.Route.html#method_replaceWith) which works the same way as `transitionTo()`.
```
locks: Why `transitionTo`, but `replaceWith()`?

source/localizable/routing/redirection.md
```diff
@@ -1,24 +1,34 @@
...
+The only difference between them is how they manage history.
+`replaceWith()` substitutes the current route entry and replaces it with that of the route we are redirecting to,
+while `transitionTo()` leaves the entry for the current route and creates a new one for the redirection.
 
 If the new route has dynamic segments, you need to pass either a _model_ or an _identifier_ for each segment.
-Passing a model will skip that segment's `model()` hook (since the model is
-already loaded).
+Passing a model will skip the route's `model()` hook (since the model is already loaded).
```
locks: Can you remove the parens as well? Thanks!

source/localizable/routing/redirection.md
```diff
@@ -1,24 +1,34 @@
...
-If you want to redirect from one route to another, you can do the transition in
-the [`beforeModel()`][3] hook of your route handler.
-
-[3]: http://emberjs.com/api/classes/Ember.Route.html#method_beforeModel
+Since a route's [`beforeModel()`](http://emberjs.com/api/classes/Ember.Route.html#method_beforeModel) executes before the `model()` hook,
+it's a good place to do a redirect if you don't need any information that is contained in the model.
 
 ```app/router.js
 Router.map(function() {
+  this.route('index', { path: '/' });
```
locks: Why the explicit index?

source/localizable/routing/redirection.md
```diff
@@ -27,23 +37,27 @@ Router.map(function() {
...
-  beforeModel() {
-    this.transitionTo('posts');
+  beforeModel(/* transition */) {
+    this.transitionTo('posts'); // Implicitly aborts the on-going transition.
   }
 });
 ```
 
+`beforeModel()` receives the current transition as an argument, which we can store and retry later.
+This allows us to return the user back to the original route.
```
locks: Unsure if this sentence clarifies, what do you think about cutting it?

source/localizable/routing/redirection.md
```diff
@@ -27,23 +37,27 @@ Router.map(function() {
...
+For example, we might redirect a user to the login page when they try to edit their profile, and immediately redirect
```
locks: We should more explicitly explain this part. Maybe we can discuss some diagrams to inline!

source/localizable/routing/redirection.md
yawboakye: [L5](https://github.com/emberjs/guides/pull/1587#discussion-diff-86676364R5) completes the line it follows. But L6 stands on its own so should be possible to take it out

source/localizable/routing/redirection.md
yawboakye: As in specifying the path? or you want it taken out completely? Although `index` routes are available there's cases where you need them explicitly:
  - when the index path isn't the `/` that Ember assumes it is
  - Nested routes where the `model` hook shouldn't go in the parent route

I added the `index` route to give a glimpse of the possibility to implement your own.

source/localizable/routing/redirection.md
yawboakye: Good catch :+1:

source/localizable/routing/redirection.md
yawboakye: I added a link to the Guides on [Storing and Retrying Transitions](https://guides.emberjs.com/v2.9.0/routing/preventing-and-retrying-transitions/#toc_storing-and-retrying-a-transition). If that's not enough we could use diagrams.

source/localizable/routing/redirection.md
locks: Not sure that's useful here, I believe index routes are explained before. And it might lead people into thinking they have to.

yawboakye:
@locks explicit `index` route has been removed.

locks:
@homu r+

PR #1588, opened 2016-08-02
Update app/routes.js at tutorial of routes and templates

raphaelcuba:
The docs show the ```app/routes.js``` without the config of ```rootURL``` defined by the environment.


locks:
Thanks for your contribution!

@homu r+

PR #1592, opened 2016-08-03
Use peekRecord() instead of findRecord()

shotgundebugging:
findRecord() triggers a scheduleFetch(), which will cause the adapter to request the record after it's been deleted.

See: https://github.com/emberjs/data/blob/v2.7.0/addon/-private/system/store.js#L612


locks:
cc @bmac

source/localizable/models/creating-updating-and-deleting-records.md
```diff
@@ -151,14 +151,12 @@ deletion can then be persisted using `save()`.  Alternatively, you can use
 the `destroyRecord` method to delete and persist at the same time.
 
 ```js
-store.findRecord('post', 1).then(function(post) {
```
bmac: Lets keep the `findRecord` because the record to be deleted may or may not be in the store but change the example to use the `backgroundReload` to disable the potential for a `scheduleFetch()`.

`store.findRecord('post', 1, { backgroundReload: false }).then(function(post) {`

bmac:
👍

shotgundebugging:
@bmac, I modified it as you suggesting, but at a second glance, passing `backgroundReload: false` will still trigger `scheduleFetch()` since `adapter.shouldBackgroundReloadRecord()` [defaults to `true`](https://github.com/emberjs/data/blob/v2.7.0/addon/adapter.js#L589)

bmac:
What version of Ember Data are you using @shotgundebugging? Since 2.7.0 the `backgroundReload` flag should override `adapter.shouldBackgroundReloadRecord()`. If it doesn't then we have a bug.

shotgundebugging:
@bmac, I'm on 2.7.0. It's working alright, though. I believe I didn't test this properly earlier.

source/localizable/models/creating-updating-and-deleting-records.md
toddjordan: I think we should explain or link to the what/why of backgroundReload.

PR #1593, opened 2016-08-03
Explain the `backgroundReload` option

shotgundebugging:
Detail why the `backgroundReload: false` option is used in the `findRecord()` examples

cc @bmac @toddjordan


source/localizable/models/creating-updating-and-deleting-records.md
```diff
@@ -162,3 +162,5 @@ store.findRecord('post', 2, { backgroundReload: false }).then(function(post) {
   post.destroyRecord(); // => DELETE to /posts/2
 });
 ```
+
+The `backgroundReload` option is used to prevent the fetching of the destroyed record, since [`findRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_findRecord) automatically schedules a fetch of the record from the adapter.
```
yawboakye: Put the link to the `findRecord()` API documentation at the end of the paragraph instead and replace with a reference.

PR #1594, opened 2016-08-03
[2.4] Revert DS import issues

acorncom:


source/localizable/models/customizing-adapters.md
```diff
@@ -138,9 +134,7 @@ The `namespace` property can be used to prefix requests with a
 specific url namespace.
 
 ```app/adapters/application.js
-import JSONAPIAdapter from 'ember-data/adapters/json-api';
```
toddjordan: For each of these, should you `+import DS from 'ember-data';`?

PR #1600, opened 2016-08-05
Fix jshint error in `list-filter-test.js` snippet

evsheffield:
Linting error: 'Ember' is defined but never used.
ember-guide-jshint


toddjordan:
Thanks for the contribution!

toddjordan:
@homu r+

PR #1601, opened 2016-08-05
Minor update to simple-component.md: Use consistent styling in the code examples.

jsArgus:
We should use consistent styling in the examples when adding or removing lines. For example in this case line 4 is added.


jsArgus:
If this is considered a useful thing, I could go through the Tutorial pages and check for similar inconsistencies.

toddjordan:
Thanks @jsArgus I think highlighting the changes are good for cases where only a few lines change to guide the user to what they need to be looking for.  In cases where there are a lot of changes we tend not to and just show the updated file.  This change is fine though.

toddjordan:
@homu r+

PR #1603, opened 2016-08-05
add subroutes tutorial

toddjordan:


toddjordan:
Cooresponing updates are also in a pr to super-rentals
https://github.com/emberjs/super-rentals/pull/35

source/localizable/tutorial/acceptance-test.md
```diff
@@ -82,7 +82,7 @@ Ember provides test helpers that we can use to perform common tasks in acceptanc
 tests, such as visiting routes, filling in fields, clicking on elements, and
 waiting for pages to render.
 
-We want our the main focus of our site to be selecting rentals, so we plan to redirect traffic going to the roote URL `/`, to our `rentals` route.
+We want our the main focus of our site to be selecting rentals, so we plan to redirect traffic going to the root URL `/`, to our `rentals` route.
```
sbatson5: `our the main`.  Should be `our main focus` or `the main focus`

toddjordan:
currently we have a route of /rentals, but also we fetch our rentals api from /rentals as well.  I think the only reason this works is because mirage is hijacking our url.  I need to change the ember data url to something like `/api/rentals`

sbatson5:
@toddjordan I can address this `api` namespace issue (since I exposed it :P ).  Have a PR here: https://github.com/emberjs/super-rentals/pull/40

Will update the guides accordingly in the `installing-addons` section if you think that's the appropriate change to make: https://github.com/emberjs/guides/blob/master/source/localizable/tutorial/installing-addons.md#ember-cli-mirage

toddjordan:
hey @sbatson5 that would be awesome.  Just pr it into this branch.

sbatson5:
PR opened here: https://github.com/emberjs/guides/pull/1609

I believe this covers introducing the namespace.  @toddjordan do you think the adapter should be part of the ember-data tutorial page or it best to group it with introducing mirage?

source/localizable/tutorial/routes-and-templates.md
```diff
@@ -194,22 +219,40 @@ Unlike the other route handlers we've made so far, the `index` route is special:
...
+
+```app/routes/index.js
+import Ember from 'ember';
+
+export default Ember.Route.extend({
+  beforeModel() {
+    this._super(...arguments);
+    this.transitionTo('rentals.index');
+  }
+});
```
acorncom: Hmm, not sure I'm a fan of doing things this way.  It makes it clearer to someone working with Ember for the first time (so I like that) but do we want to be adding in `transitionTo` calls in a `beforeModel` hook to our tutorial?  Seems like we could alternately map rentals to the path of `rentals.index` like the following:

```js
this.route('rentals', { path: '/' });
```

@locks do you have thoughts here?

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,425 @@
...
+We will add a new route handler to mirage to handle when the http request comes in to fetch a rental.
+
+```mirage/config.js{+57,+58,+59,+60}
+export default function() {
+  this.namespace = '/api';
+
+  let rentals = [
+    {
+      type: 'rentals',
+      id: 'grand-old-mansion',
```
acorncom: Are we switching to string ids here?  It's quite elegant if we can do it, but we'll want to indicate that we changed these, as our current mirage conf uses numeric ids instead

acorncom:
Haven't reviewed this all the way, will be back later ...

source/localizable/tutorial/subroutes.md
toddjordan: Yes.  Good catch.  I actually would like to just give it string ids from the start.

source/localizable/tutorial/routes-and-templates.md
toddjordan: I'm ok going either way

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,425 @@
...
+    {{#each rentals as |rentalUnit|}}
+      <li>{{rental-listing rental=rentalUnit}}</li>
+    {{/each}}
+  </ul>
+{{/list-filter}}
+{{outlet}}
+```
+
+Finally, we need to make our controller that has our filter action available to the new nested index route.
+Instead of moving the whole controller file over to `app/controller/rentals/index.js` from `app/controller/rentals.js`, we'll just take advantage of the route's `controllerName` property to just point at our existing `rentals` controller.
```
acorncom: Hmm, I don't think we should be recommending that folks do this here.  It's a nifty feature (and one that I can see being useful), but I'd rather be encouraging best practices in the guides and this smells a bit funny.

@locks thoughts on this one?

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,425 @@
...
+
+Next, we can update the template for our show route (`app/templates/rentals/show.hbs`) and list the information for our rental.
+
+```app/templates/rentals/show.hbs
+<div class="jumbo">
+  <a href="#" class="button right up-11">Edit</a>
+  <h2 style="margin-bottom:15px">{{model.title}}</h2>
+  <div class="right" style="width:50%;padding-left:30px">
+    <div class="map right" style="background: url({{model.address}}) center center;"></div>
+    <div class="detail" style="margin-top:10px">
```
acorncom: Do we want to have hard-coded CSS here?

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,425 @@
...
+    </div>
+    <p>&nbsp;</p>
+    <p>{{nmodel.description}}</p>
+  </div>
+  <img src="{{model.image}}" alt="" class="" style="width:50%;height:initial;position:static">
+</div>
+```
+
+We pass our `model` to the `rental-listing` component and add a new section listing the rental's `title`.
+Browse to `localhost:4200/rentals/grand-old-mansion` and you should see the information listed for that specific rental.
```
acorncom: I'm not seeing a `rental-listing` component here, do we mean the `rental-property-type` component?

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,419 @@
...
+    <div class="detail">
+      <strong>Location:</strong> {{model.city}}
+    </div>
+    <div class="detail">
+      <strong>Number of bedrooms:</strong> {{model.bedrooms}}
+    </div>
+    <p>&nbsp;</p>
+    <p>{{nmodel.description}}</p>
+  </div>
+  <img src="{{model.image}}" alt="" class="" style="width:50%;height:initial;position:static">
```
acorncom: Apparently my comment got removed after my code push.  Do we want to do inline styles throughout this `.hbs` file?  I'd prefer to see them go away ...

acorncom:
After the removal of the unit test that @toddjordan was doing, I think it'd be good to get @locks feedback on the two "best practices" questions above.  Other than that, I'm a 👍 on merging this, thanks for all your work @sbatson5 and @toddjordan!

source/localizable/tutorial/subroutes.md
toddjordan: yeah, we'll need to update ember-cli-tutorial-style.  will do

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,419 @@
...
+Now that we can load pages for individual rentals, we'll add a link (using the `link-to` helper) within our `rental-listing` component to navigate to individual pages.
+Clicking on the title will load the detail page for that rental.
+
+```app/templates/components/rental-listing.hbs{+6}
+<article class="listing">
+  <a {{action 'toggleImageSize'}} class="image {{if isWide "wide"}}">
+    <img src="{{rental.image}}" alt="">
+    <small>View Larger</small>
+  </a>
+  <h3>{{link-to rental.title "rentals.show" rental.id}}</h3>
```
toddjordan: should be rental instead of rental.id

source/localizable/tutorial/subroutes.md
toddjordan: This applies to the next section.  will move or delete.

source/localizable/tutorial/subroutes.md
toddjordan: I went ahead and converted all ids to strings

source/localizable/tutorial/routes-and-templates.md
```diff
@@ -171,9 +171,34 @@ Now, we'll add a link to our contact page so we can navigate from back and forth
...
+  {{#link-to 'about' class="button"}}
+    About Us
+  {{/link-to}}
+</div>
+```
+
 ## An Index Route
 
 With our two static pages in place, we are ready to add our home page which welcomes users to the site.
+At this point our main page in our application is our rentals page, which we've already created a route for.
```
sbatson5: Nitpick: ending the sentence with a preposition. `for which we've already created a route.`

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,361 @@
+Up to this point, we've generated four top level routes.
+
+* An `about` route, that gives information on our application.
+* A `contact` route, with information on how to contact the company.
+* A `rentals` route, where we will allow users to browse rental properties.
+* and the `index` route, which we've set up to redirect to the `rentals` route.
```
sbatson5: Should this `and` be capitalized? Or perhaps use a comma instead of a period on the line above?

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,361 @@
...
+Our `rentals` route is going to serve multiple functions.
+From our [acceptance tests](../acceptance-test), we've shown that we want our users to be able to browse and search rentals, as well as see detailed information for individual rentals.
+To satisfy that requirement, we are going to make use of Ember's [nested route capability](../../routing/defining-your-routes/#toc_nested-routes).
+
+By the end of this section we want to have created the following new routes:
+
+* A `rentals/index` route that displays the rental page's general information, and also lists available rentals.
+The index nested route is shown by default when the user visits the `rentals` URL.
+* A `rentals/show` route that still displays the rental page's general information, while also showing detailed information about a selected rental.
+The `show` route will get substituted with the id of the rental being shown. (for example `rentals/grand-old-mansion`).
```
sbatson5: Thoughts on removing the period after `shown` since we have one after the parens?

source/localizable/tutorial/subroutes.md
sbatson5: I actually didn't know you could do this. Very cool 👍

source/localizable/tutorial/subroutes.md
acorncom: I believe it can be either, correct?  Thought it might make more sense to use `rental.id` to make it clearer what is going on.  I see folks regularly getting trolled by the fact that they can follow a `{{rentals.show rentals}}` link that works but then when refreshing on that page (if they haven't added a route `model` hook) the page then breaks.  This doesn't explain that fully, but at least is one step toward the right direction ...

locks: can we please pretend that the inline `link-to` doesn't exist? :X

source/localizable/tutorial/subroutes.md
toddjordan: will remove

source/localizable/tutorial/subroutes.md
toddjordan: Agree its awkward.  Will update

source/localizable/tutorial/routes-and-templates.md
toddjordan: 👍 good catch

source/localizable/tutorial/subroutes.md
toddjordan: So if I didn't do this I would have to add a good sized section explaining that we need to remove this controller, create a new controller, and move the logic.  I think its easier to do it this way, plus you are introducing a potentially useful feature.  I don't think its really that awkward since the controllers attached to the parent route. which makes sense.

source/localizable/tutorial/subroutes.md
toddjordan: @locks can you explain?

locks: @toddjordan I find (for myself, and for others) the inline version of `{{link-to}}` to be confusing, especially because the first argument is the text of the link, and not the route to link to. So what I'm suggesting is using the block form throughout the guides.

toddjordan: sounds good. will update

source/localizable/tutorial/routes-and-templates.md
toddjordan: if we keep `transitionTo`, we need to replace it with `replaceWith`.  This is because transitionTo  still updates the history I believe, so you could break the back button.

source/localizable/tutorial/subroutes.md
```diff
@@ -0,0 +1,364 @@
...
+Now browse to `localhost:4200/rentals/grand-old-mansion` and you should see the information listed for that specific rental.
+
+![Rental Page Nested Show Route](../../images/subroutes/subroutes-super-rentals-show.png)
+
+## Linking to a Specific Rental
+
+Now that we can load pages for individual rentals, we'll add a link (using the `link-to` helper) within our `rental-listing` component to navigate to individual pages.
+Here, the `link-to` helper takes the route name and the rental model object as arguments.
+When you pass an object as second argument into the `link-to` block helper, it will by default [serialize](http://emberjs.com/api/classes/Ember.Route.html#method_serialize) the object to the ID of the model into the URL. 
+Alternately, you may just pass `rental.id` for clarity.
```
toddjordan: @acorncom I put some explanation around the use of rental in the link-to.  I like that fact that it exposes that route serialization is a thing.

toddjordan:
@acorncom @locks do you guys think we are good to go?

source/localizable/tutorial/routes-and-templates.md
toddjordan: @acorncom I think we have to leave as is because what you suggest is making rentals the index path.  If rentals is the index path, then we can't have nested routes!  I don't see a way around replaceWith for the tutorial.  Let me know if I'm wrong and you see another way...

acorncom: @toddjordan Oops, my mistake. Sounds good, let's skip it 😃

sbatson5:
@toddjordan looks like this need a rebase. Anything I can do to help push this through?

toddjordan:
@sbatson5 thanks will bring it up to date.  discussed at last weeks learn meeting, and its ready.  Just trying to get the right approval.

toddjordan:
@emberjs/core Could we have a pair of eyeballs from core on this section before pulling the trigger? Thanks!

nathanhammond:
Do you wish to address loading substates (`_loading`) and child states (`application/loading`)?

source/localizable/tutorial/subroutes.md
ebryn: I don't think we should teach `controllerName` I would recommend using a module reexport instead if possible?

acorncom:
Re: addressing loading substates, I would vote for leaving that for another PR.  This one has been in-process for the last three months 😀

source/localizable/tutorial/subroutes.md
toddjordan: Good feedback, will remove the usage of controllerName.  Thanks

toddjordan:
@nathanhammond thanks for your feedback!  While I think loading substates are something to think about when developing a production app, we want to keep a line between a beginner tutorial and showing off various features.  Maybe if we can create a PR that introduces this in a straightforward, simple way we can incorporate, but I'm with David with the mind that we should get this first pass out there.

source/localizable/tutorial/subroutes.md
```diff
@@ -131,15 +131,13 @@ Now that we are returning all of our rentals to the nested route's model, we wil
 ```
 
 Finally, we need to make our controller that has our filter action available to the new nested index route.
-Instead of moving the whole controller file over to `app/controller/rentals/index.js` from `app/controller/rentals.js`, we'll just take advantage of the route's `controllerName` property to just point at our existing `rentals` controller.
+Start by running `ember g route rentals/index` to create an index controller for our nested route.
```
toddjordan: @ebryn @acorncom for this section I'm using module re-export instead of `controllerName`.  Does this look ok?

paul-howard:
Feel free to correct me if I'm wrong, but I think this PR broke the tutorial??

The sample code on the `routes-and-templates/` page now assumes the existence of a `rentals.index` route when introducing the `replaceWith` method, but that sub-route ("nested route"? which term is preferred? they seem to be used interchangeably?) won't be created until much later in the tutorial!

Changing all instances of `rentals.index` to simply `rentals` seems to fix everything.

toddjordan:
yep.  the 'rentals.index' there is a typo.  Creating a pr to correct.  Thanks for reporting!

paul-howard:
Happy to give back any way I can! 😄

PR #1606, opened 2016-08-06
Clarify we're creating a template for the application

soulchainer:
In section **Adding a Banner with Navigation** it states that  `ember g template application` if for «create the application». It isn't: the thing it creates is a template for the application.

### DIFF
--- a/source/localizable/tutorial/routes-and-templates.md
+++ b/source/localizable/tutorial/routes-and-templates.md
@@ -214,7 +214,7 @@ Let's update our `index.hbs` with some HTML for our home page and our links to t
 
 In addition to providing button-style links in each route of our application, we would like to provide a common banner to display both the title of our application, as well as its main pages.
 
-First, create the global template for the application by typing `ember g template application`.
+First, create the application template for the application by typing `ember g template application`.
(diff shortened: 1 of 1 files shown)


source/localizable/tutorial/routes-and-templates.md
```diff
@@ -214,7 +214,7 @@ Let's update our `index.hbs` with some HTML for our home page and our links to t
 
 In addition to providing button-style links in each route of our application, we would like to provide a common banner to display both the title of our application, as well as its main pages.
 
-First, create the application by typing `ember g template application`.
+First, create the global template for the application by typing `ember g template application`.
```
locks: I'd rather insert the missing "template" instead, as in "create the application template". What do you think?

soulchainer: I think you're **so right**, so change done.

soulchainer:
Damn, I'm so clumsy, sorry (I had my mind in another place >_<).The squash and merge button will do the magic, I think.

locks:
No worries, thanks for your contribution!

@homu r+

PR #1607, opened 2016-08-07
Remove class "image" from <img>

soulchainer:
Just because it's only there to be error prone. And I say this because, obviously, I suffered it ^_^U.

1. It doesn't have anything to do with the styles of the application.

   If you check the styles provided by `ember-cli-tutorial-style` addon, the class `image` is only used to style the `a` tag (~~I didn't find a repo for it~~ [CSS here](https://github.com/emberjs/super-rentals/blob/master/vendor/ember-tutorial.css)):

  `vendor/ember-tutorial.css`

   ```css
   ...
   .listing a.image.wide {
     max-width: 100%;
     position: relative;
     z-index: 999;
   }
  ...
  .listing a.image {
    max-width: 150px;
    display: block;
  }
  ...
   ```

   Just that.It only styles the `a` tag.
2. Later, in this same tutorial page, in the section [Hiding and Showing an Image](https://guides.emberjs.com/v2.7.0/tutorial/simple-component/#toc_hiding-and-showing-an-image), the template `app/templates/components/rental-listing.hbs` is edited:

   ```hbs
   <a class="image {{if isWide "wide"}}">
     <img src="{{rental.image}}" alt="">
     <small>View Larger</small>
   </a>
   ```

   The other lines edited are highlighted with a very visual green background and marked with a `+`, because are new lines. There is no warning at all that the `` tag here is also being edited (the *not used in any moment* `class="image"` is removed here). You can miss that change, because your attention is lured to that lines with a green background (yes, I totally miss this «silent» change).
3. *Very much later*, when you hit component `rental-listing-test.js` integration test (and you hit that at the very end of the **Super Rentals** Tutorial), you realise that not all your tests are passing. And guess what, yes: the error is caused because you didn't removed that `class="image" in the `` that never has any mean. The error is caused because this:

   ```js
     this.render(hbs`{{rental-listing rental=rentalObj}}`);
    assert.equal(this.$('.image.wide').length, 0, 'initially rendered small');
    this.$('.image').click();
    assert.equal(this.$('.image.wide').length, 1, 'rendered wide after click');
    this.$('.image').click();
    assert.equal(this.$('.image.wide').length, 0, 'rendered small after second click');
   ```

   You see the problem? The `click` event is triggered over the `.image`. But you have both, `img` and `a` elements with the `image` class, so both are clicked, click event propagates and well, the thing is that this assertion:

  ```js
    assert.equal(this.$('.image.wide').length, 1, 'rendered wide after click');
  ```

  always fail, because `a` is always clicked twice, first directly and second when `` inside it is clicked and the event bubbles, I guess. The fact is that the `length` is always zero (because you didn't delete the `class="image"` from the ``).

So, like I don't see here any reason for the initial existence of `class="image"` in the `` tag, and considering the fact that it's error prone, I strongly suggest this change ;). Unless I'm missing something, I think that's better off. All tests pass, everything continues working the same, everyone is happy :).

### DIFF
--- a/source/localizable/tutorial/simple-component.md
+++ b/source/localizable/tutorial/simple-component.md
@@ -66,7 +66,7 @@ To start, let's move the rental display details for a single rental from the `in
 
 ```app/templates/components/rental-listing.hbs{+2}
 <article class="listing">
-  <img src="{{rental.image}}" class="image" alt="">
+  <img src="{{rental.image}}" alt="">
   <h3>{{rental.title}}</h3>
   <div class="detail owner">
     <span>Owner:</span> {{rental.owner}}


soulchainer:
Note: I just saw the PR #1586, that verses about the same problem, but in different files, I guess, so I think this still applies.

soulchainer:
Note 2: not. Is the same... Yeah: that's me wasting your time :(. Closing this.

soulchainer:
Not. Isn't the same. Just almost the same xD. Different lines. Seriously. I'm so clumsy lately.

toddjordan:
Hi @soulchainer thanks for working ont his.  I'll take a look to make sure.  Note that there is a functional super-rentals app at https://github.com/emberjs/super-rentals, so you can test any updates you make where as well.

toddjordan:
The image class doesn't exist in the super-rentals repo https://github.com/emberjs/super-rentals/blob/master/app/templates/components/rental-listing.hbs#L3 , so its a good idea to sync the guides. Thanks again!

toddjordan:
@homu r+

PR #1611, opened 2016-08-13
Correct dependencies in bower.json

chuckdishmon:
Remove Ember CLI Test Loader as a dependency found in bower.json. This is a dev dependency in package.json.


locks:
Thanks for your contribution!

@homu r+

PR #1614, opened 2016-08-17
Fix broken helper links

dubbs:
This fixes duplicate reference-style links.  Previously, `{{get}}` linked to `{{concat}}`.


toddjordan:
Good catch.  Thanks, @dubbs

toddjordan:
@homu r+

PR #1620, opened 2016-08-25
update example id

chuckdishmon:
Make the example id consistent with example urls and subsequent id requests.


locks:
Thanks for your contribution!

@homu r+

PR #1621, opened 2016-08-25
correct a reference to .jshintrc's predef

mminkoff:
The predef section of .jshintrc is an array of strings, not an object.  One does need to add "option":true elsewhere in the file to turn on features, but to tell jshintrc to ignore predefined variable names you just add the name to this array, not 'name:true'.  

Note here: http://jshint.com/docs/options/#predef it says "This option allows you to control which variables JSHint considers to be implicitly defined in the environment. Configure it with an array of string values."


acorncom:
@homu r+ Thanks!

PR #1624, opened 2016-08-26
Inline event handler with action

pixelhandler:
- Add clarity for using an event parameter with an action, only when assigning an action to an inline event handler
  - Update `handling-events` page https://guides.emberjs.com/v2.7.0/components/handling-events/
- See https://github.com/emberjs/ember.js/issues/13806

inline-event-handler


source/localizable/components/handling-events.md
```diff
@@ -73,6 +73,46 @@ export default Ember.Component.extend({
 });
 ```
 
+In the above component "didDrop", the action that was passed in, is called
+with the `id` value found through the `drop` event object.
+
+
+Another way to preserve native event behaviors and use an action is to
+assign a (closure) action to an inline event handler. Consider the
+`onlcick` handler of an element…
```
Serabe: `onclick`

nathanhammond:
LGTM, much more clear.

source/localizable/components/handling-events.md
```diff
@@ -73,6 +73,46 @@ export default Ember.Component.extend({
 });
 ```
 
+In the above component "didDrop", the action that was passed in, is called
```
toddjordan: surround `didDrop` with backticks.  otherwise it fails spellcheck

source/localizable/components/handling-events.md
pixelhandler: lol, oops.

pixelhandler:
@emberjs/learning-team-managers I updated, squashed and rebased.

PR #1625, opened 2016-08-27
Minor grammar updates

stavarotti:
A few minor improvements to reduce word count and increase reading ease.


locks:
Thanks for your contribution!

@homu r+

PR #1626, opened 2016-08-30
Match code with the latest version

gnowoel:
The code listing of `app/router.js` in the tutorial does not match the one generated with latest `ember-cli` (v2.7.0).


locks:
Thanks for your contribution!

@homu r+

PR #1628, opened 2016-08-30
Update creating-updating-and-deleting-records.md

jsArgus:
The "Handling Validation Errors" example misses the correct trigger for the "highlight ribbon" class.


locks:
Thanks for your contribution!

@homu r+

PR #1629, opened 2016-08-30
Use the adverb instead of adjective

Mak-Di:


source/localizable/tutorial/autocomplete-component.md
```diff
@@ -18,7 +18,7 @@ We want our component to call out to two actions: one action to provide a list o
 
 For our initial test, we'll simply check that all the cities we provide are rendered and that the listing object is accessible from the template.
 
-Since we plan to use Ember Data as our model store, we need to make our action calls to fetch data asynchronous, so we'll return promises.
+Since we plan to use Ember Data as our model store, we need to make our action calls to fetch data asynchronously, so we'll return promises.
```
locks: Hi @Mak-Di, I disagree with this change. What I interpret from the original text is that the action calls need to be asynchronous. The confusing part is that fetching data is also done asynchronously, hence the action call needing to return a promise.

Mak-Di: @locks, I've understood you, thank you for the explanation.

PR #1631, opened 2016-09-01
Enable default browser :focus outline

krishandley:
I found the ember guides to be very frustrating to use without a mouse.


acorncom:
@krishandley Could you send over a few screenshots showing what this looks like with these changes?

krishandley:
## Before

without-focus

## After

with-focus

locks:
Thanks for your contribution!

@homu r+

PR #1633, opened 2016-09-02
Applies #1578 to 2.6 guides

acorncom:


acorncom:
There's something weird going on here, will be checking this later ...

acorncom:
2.6 already has the new imports.

PR #1635, opened 2016-09-05
Correct “will be ran” to “will be run”

CodingItWrong:
I’m not sure of the grammatical term for this, and it sounds weird if you think about it too much, but I’m pretty sure the tenses are:

- run
- ran
- will be run
- will have been run

etc.


acorncom:
@homu r+  Yep, you're right, nice catch ;-)

CodingItWrong:
@acorncom sorry for missing "test exempted"—I'll try to get that right next time!

PR #1638, opened 2016-09-09
change rentals.index to rentals

toddjordan:
fixing a typo which was causing tutorial not to work.

PR #1639, opened 2016-09-09
rentals.index doesn't work when following tutorial

sblosser:
I'm following along with the tutorial (from the start) and when using "rentals.index" instead of "rentals", my local copy behaves differently than what is shown in the tutorial. If I replace all instances of "rentals.index" with "rentals", everything works the same as the tutorial. I'm not sure where the error might be as I'm clearly an Ember newbie, which is why I'm following the tutorial!


toddjordan:
A fix is already forthcoming. Thanks for reporting!

toddjordan:
Thanks again... fixed by #1638 .  Looks like the fixes came in within an hour of each other :-P

PR #1643, opened 2016-09-12
Point a link to the correct version of the doc

gnowoel:
A link in the Quick Start page (v2.8.0) incorrectly points to an older version of the doc (v2.7.0).

### DIFF
--- a/source/localizable/getting-started/quick-start.md
+++ b/source/localizable/getting-started/quick-start.md
@@ -19,7 +19,7 @@ npm install -g ember-cli
 ```
 
 Don't have npm? [Learn how to install Node.js and npm here][npm].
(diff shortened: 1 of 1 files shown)


acorncom:
@gnowoel Thanks for catching that.  Rather than coding in an absolute path (that includes the version number) could you send in a patch with a relative path that works instead?  That way we don't have to remember to update that link in the future.

gnowoel:
@acorncom Nice idea. I've updated the pull request.

acorncom:
@gnowoel Thanks very much.  Hate to ask you to change one last thing but making changes to all the localized files (which was a nice touch on your part) actually breaks things for our tooling.  Could you just edit the `quick-start.md` file?  The other ones automatically get synced.

Thanks!

gnowoel:
@acorncom Oh, sorry about that! I just updated the PR again.

acorncom:
Thanks @gnowoel, that works great!

@homu r+

PR #1644, opened 2016-09-12
Fixing missing handlebars

jsfuchs:
Found this bug while going through the tutorial. As it is now, you will see the string literal 'rental.type'.


acorncom:
@homu r+ Thanks and good catch!

/cc @toddjordan

PR #1645, opened 2016-09-12
Update simple-component.md

wairton:
Remove duplicated {{#each model as |rental|}} code


source/localizable/tutorial/simple-component.md
```diff
@@ -100,7 +100,6 @@ with our new `rental-listing` component:
 
 {{#each model as |rentalUnit|}}
   {{rental-listing rental=rentalUnit}}
-{{#each model as |rental|}}
```
toddjordan: The issue is that when we initially define this template in the model-hook tutorial step, we define the variable as `rental`.  

The real issue here is that the add/remove line numbers need to be slid upward:
`app/templates/rentals.hbs{+12,+13,-14,-15,-16,-17,-18,-19,-20,-21,-22,-23,-24,-25,-26,-27,-28,-29}`

Then we don't have to delete the `#each`.

toddjordan:
@wairton thanks for contributing, the same exact fix came in on another PR and we merged that one, but we thank you for pitching in.

PR #1646, opened 2016-09-12
Controllers instead of components have access to the store.

gagamil:
Hi,
I'm new to Ember. Found a possible error in docs.
If I get it right the _Controllers_ and _Routes_ have access to toe shared store (not the _Components_).
Thanks.


toddjordan:
@gagamil thanks for contributing.  Components also have access to the store.  I think the text is ok as-is as controllers are on their way out in a future release.  If anything I would add controllers to the other two.

gagamil:
Oh, ok.
Seems that I got confused somewhere while reading the docs.
The example in the guide states that the component should not fetch the data. It should present the data passed to it. But again, all that was before the store concept was introduced.
I'll close this PR. 
Thanks for your comment.

toddjordan:
@gagamil I would agree that its generally a best practice to do store operations as high up in you tree as you can, and for most simple applications that can be in the controller/route, so you are not wrong.  That said you _can_ access the store from components, like you can any service ( the store is a service) and there are cases where it may make sense to do so.

gagamil:
@toddjordan thank you for your explanations. I'm a Django dev and relatively new to the front-end and JS.
Sounds logic that a component might need access to the store.

PR #1650, opened 2016-09-14
Update the link for Node.js installation on Linux

gnowoel:
The guide for installing Node.js via package manager has moved out of the old Wiki, where the link currently points to.


locks:
Thanks for your contribution!

@homu r+

PR #1651, opened 2016-09-15
Add explicit hbs dependency in Component tests

zzarcon:
Include **hbs** import in Component tests example. Otherwise the test will fail sin it will undefined and I think is not that easy to figure out that the dependency come form **htmlbars-inline-precompile** package.

Thanks!


toddjordan:
Thanks for contributing! 
@homu r+

zzarcon:
@toddjordan welcome!

PR #1653, opened 2016-09-18
Format problem

ubuntuvim:
The last piece of code is missing a symbol.


locks:
Thanks for your contribution!

@homu r+

PR #1654, opened 2016-09-19
Remove explicit names from initializers.

eventualbuddha:
Since [ember-load-initializers already adds names to initializers based on the filename](https://github.com/ember-cli/ember-load-initializers/blob/7ebe1d308cbbbfc3e6566cd05699e9a36fb7bddb/addon/index.js#L19-L20), we should not bother specifying them in the initializer files themselves. This brings initializers more in-line with how helpers are handled. This commit also makes better use of ES2015 features.


source/localizable/applications/initializers.md
```diff
@@ -34,8 +34,7 @@ export function initialize(application) {
 };
 
 export default {
-  name: 'shopping-cart',
-  initialize: initialize
+  initialize
```
eventualbuddha: I added it like this because it's fairly common to add `before` and `after` and I wanted to maintain symmetry. I can change it though.

eventualbuddha: Done.

rwjblue:
@stefanpenner - Any objections?

kellyselden:
I'm cool with this change.

eventualbuddha:
I don't have time to work on this right now, but I believe the maintainers have edit access for this branch.

source/localizable/applications/initializers.md
locks: Just wanted to say you were right, it was a better idea to keep it multiline ;)

locks:
**NOTE** ONLY MERGE WHEN https://github.com/emberjs/ember.js/pull/14338 IS IN A RELEASE

rwjblue:
This was released in ember-source@2.15

locks:
Thanks @eventualbuddha, sorry for letting it slide for so long 😅 .

PR #1659, opened 2016-09-22
[v1.10] Added in new Algolia search system

dfalling:
This commit applies the same search changes as in v1.11.0.


data/search.yml
```diff
@@ -1,3 +1,3 @@
-engine_key: 'UdruvUp_iEcu2wdYqw9D'
-document_type: 'guides-v1-10-0'
-revision: 'v1.10.0'
+api_key: '760969ef081fcadc7e0e60faefdb0907'
+document_type: 'guides-v1-11-0'
+revision: 'v1.11.0'
```
acorncom: Could you change these back to v1.10.0?

acorncom:
Helps with https://github.com/emberjs/guides/issues/1561

data/search.yml
dfalling: Fixed.

acorncom:
CI failures were occurring for guides at this point in time, these issues are not with this PR

PR #1660, opened 2016-09-22
Google Maps Utility Typo

Wylan:
Line 275 reads "Our app needs a single function, `createMapElement`" however it appears the name of the function on line 288 is actually `createMap`.


locks:
Thanks for your contribution!

@homu r+

PR #1663, opened 2016-09-26
Quick Start Header Heirarchy

una:
This is a small change, but both semantically and visually I think it would make more sense to use an `<h1>` in `application` to show that it is a level "above" any of the other created routes (which use `<h2>`)


locks:
Thanks for your contribution!

@homu r+

Issue #1661, opened 2016-09-23
Problems encountered in Adding Nested Routes


gnowoel:
I had some small problems when following the [Adding Nested Routes](https://guides.emberjs.com/v2.8.0/tutorial/subroutes/) section of the tutorial.

### Generating a Nested Index Route

> If you open up your Router (app/router) you may notice that nothing has updated.

This seems no longer true. After running the `ember g route rentals/index` command, the route entry for `rentals` would be changed from:

    this.route('rentals');

to this:

    this.route('rentals', function() {});

### Setting up Data for the Nested Detail Route

We've added `description` fields in the rentals data. But to make them work, we may also need to update the model:

    export default DS.Model.extend({
      // ...
      description: DS.attr()
    });

### Linking to a Specific Rental

In [the acceptance tests](https://guides.emberjs.com/v2.8.0/tutorial/acceptance-test/), there's a line specifying that rental descriptions should be showing:

    assert.equal(find('.description').length, 1, 'should list a description of the property');

The test case would fail because we missed the `description` class in the template of the show route. We need to update the following line:

    <p>{{model.description}}</p>

to look like this:

    <p class="description">{{model.description}}</p>

sbatson5:
I believe this can be closed @toddjordan

locks:
Appreciated @sbatson5!

PR #1666, opened 2016-09-27
[#1661] address missing subroutes content

sbatson5:
addresses: https://github.com/emberjs/guides/issues/1661


toddjordan:
I think we should fix it by just adding description to the original ember-data tutorial page.  That will keep the size of the nested routes section down.

sbatson5:
@toddjordan Added it to the ember-data tutorial and the initial time `mirage` is set up. Let me know if this looks right.

toddjordan:
Excellent!  Thanks @sbatson5 !!!
@homu r+

PR #1670, opened 2016-09-29
Use the verb 'be' in a plural

Mak-Di:


locks:
Thanks for your contribution!

@homu r+

PR #1674, opened 2016-09-30
Fix typo in index controller path

adibsaad:


acorncom:
@toddjordan thoughts on this one?

toddjordan:
Yes, excellent catch.  Thanks!

@homu r+

PR #1677, opened 2016-10-01
Added missing hyphen

Mak-Di:


locks:
Hi @Mak-Di, I believe we actually want the verb here, not the noun. You'd have to change the order to use the noun, like: "Thankfully, `RSVP` has a built-in solution for this problem."

PR #1679, opened 2016-10-03
Added missing hyphen

Mak-Di:
In this case it's an adjective


locks:
Indeed! Appreciated 😄 

@homu r+

Issue #1665, opened 2016-09-27
Inconsistent testname


ingobecker:
At the end of the section [The Model Hook](https://guides.emberjs.com/v2.8.0/tutorial/model-hook/) the tutorial reffers to a test named `should initially list 3 rentals`. I guess it's reffering to the test actually named `should list available rentals.` on the page [Setting up Tests](https://guides.emberjs.com/v2.8.0/tutorial/acceptance-test/).

toddjordan:
Good catch!  If you have some time would love help updating that text :-)  if not will get to it soon...

ingobecker:
Should the text be changed or should i make a new screenshot reflecting the names in the text? I also found a minor issue in [Installing Addons](https://guides.emberjs.com/v2.8.0/tutorial/installing-addons/), reffering to an incomplete path for the mirage config which i could fix as well. Should i make changes to the english version only or correct the translations as well?

acorncom:
@ingobecker if you could send in a new screenshot, that'd be great.  Re: your other issue, just make your changes to the english version, the translations sync automatically

toddjordan:
Thanks! make sure you run the screenshot through a program like https://tinypng.com  to ensure fast loading of the image :-)

toddjordan:
also named wrongly in the image in "Services and Utilities".  Part of this fix should include updating that as well.

williamhaley:
@toddjordan I believe I have the appropriate screenshot for the "Services and Utilities" section in this PR https://github.com/emberjs/guides/pull/2129

Please let me know if this is wrong, or if there are any other updates I can make to help out with this issue. Thanks!

PR #1683, opened 2016-10-04
updated the screenshot to be consistent with the guide, has gotten ou…

jaschwar:
…t of date


locks:
It's a bit smaller than the original, do you have a way to test visually on a retina screen?

toddjordan:
The font turns out a bit small to read.  Could you retake the screen shot with a narrower window?  The image's width is capped to the width of the screen, so I think a narrower window will show the image bigger. Here's what it looks like today.  I'd like the get the text size closer to the tutorial text size.

image

toddjordan:
Will address #1665 when merged

jaschwar:
I updated with a large screenshot, let me know what you think ?

toddjordan:
not quite.  😬   I was talking about less the pixel size of the image, but rather the size of the font within the image.  The images get resized to fit the screen, so what I'm looking for is a smaller window size, so that the font shows bigger.  either that or a bigger font size.

acorncom:
@jaschwar Any updates here?  Would love to get this merged ...

PR #1684, opened 2016-10-04

jaschwar:
So, not sure this is the best place to put this comment or not, but when I worked through this tutorial it tripped me up that the main route was broken until the index was added back in.  Could also change the approach -- add the index 1st.  But, that would be a much larger re-write.

### DIFF
--- a/source/localizable/tutorial/routes-and-templates.md
+++ b/source/localizable/tutorial/routes-and-templates.md
@@ -7,6 +7,8 @@ Let's take a look at the application from the user perspective before we get sta
 We arrive at the home page which shows a list of rentals.
 From here, we will be able to navigate to an about page and a contact page.
 
+Keep in mind that generating the routes will break the main route until its added back in index route section, and means that navigating to [`http://localhost:4200/`](http://localhost:4200/) yields an loading an empty white page. 
+
 Now, let's start by building our "about" page.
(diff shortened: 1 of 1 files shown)


source/localizable/tutorial/routes-and-templates.md
```diff
@@ -7,6 +7,8 @@ Let's take a look at the application from the user perspective before we get sta
 We arrive at the home page which shows a list of rentals.
 From here, we will be able to navigate to an about page and a contact page.
 
+Keep in mind that generating the routes will break the main route until its added back in index route section, and means that navigating to [`http://localhost:4200/`](http://localhost:4200/) yields an loading an empty white page.
```
toddjordan: Instead of saying adding a route breaks the main route, you can explain that adding your first route removes the Ember welcome screen.

```
Keep in mind that generating your first route will remove the default Ember welcome page.  
Now when you navigate to [`http://localhost:4200/`](http://localhost:4200/), the app will load an empty white page.  
We will fix this behavior later on in this section.
```

yawboakye:
@jaschwar I think this PR should be good once you consider @toddjordan's comment and resolve the merge conflict 😄

toddjordan:
No longer true in recent ember releases.  welcome page is on application.hbs until you remove it.

PR #1685, opened 2016-10-05
updated the screenshot with the passing navigation tests to be more c…

jaschwar:
…omplete -- showing passing and failed tests


toddjordan:
The text on the new image is considerably smaller, making the new image harder to read (at least on my macbook display). If you reduce the size of the window, or increase the font size, you should get a better shot.

jaschwar:
I updated with a bigger screenshot let me know if works ? @toddjordan

toddjordan:
Hi @jaschwar the size is 👌  now.  Sorry to be a PITA, but I need to to take another screenshot of the same, except this time uncheck "Hide Passed Tests".  We want to show the user the test that _should be_ passing as well.

toddjordan:
also, like I mentioned in your other PR I prefer larger font relative to window size, so the reader can read it.  Sorry I didn't communicate that better 😦

PR #1688, opened 2016-10-06
Update model-hook.md

jaschwar:
This updates the object model to be consistent object used in the next tutorial -- tripped me up as there was no good reason it had to be different.

### DIFF
--- a/source/localizable/tutorial/model-hook.md
+++ b/source/localizable/tutorial/model-hook.md
@@ -14,30 +14,39 @@ Let's open `app/routes/rentals.js` and add our hard-coded data as the return val
 ```app/routes/rentals.js
 import Ember from 'ember';
 
-let rentals = [{
-  id: 'grand-old-mansion',
-  title: 'Grand Old Mansion',
-  owner: 'Veruca Salt',
-  city: 'San Francisco',
-  type: 'Estate',
-  bedrooms: 15,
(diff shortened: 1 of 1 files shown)


locks:
There is a very good reason, @jaschwar.

In "The Model Hook" we are returning a Plain Old JavaScript Object (POJO) from the `model()` hook so you can use it directly.
In the "Installing Addons" section we set up mirage, which is a development API server. What we're returning there is actually what a JSONAPI payload from an API server should look like, hence the difference. After creating the `DS.Model` in "Using Ember Data", you won't need to change the template, you'll still refer to it as `{{rental.owner}}`.

We will look into the possibility of moving some things around to make the connection of mirage<->Ember Data clearer, thanks for the feedback.

jaschwar:
@locks what is the reason we are using POJO from the model hook?  For simplicity?  seems cleaner to me to have a similar object model.  If nothing else we should change the documentation in the "Installing Addons" spell out the change in the object model and why.  I think verbosity would help.

Thanks

locks:
But we do:
* https://guides.emberjs.com/v2.8.0/tutorial/ember-data/#toc_updating-the-model-hook
* https://guides.emberjs.com/v2.8.0/tutorial/installing-addons/#toc_ember-cli-mirage

PR #1689, opened 2016-10-07
Update description of queryRecord()

MarkMT:
This updates the description of `store.queryRecord()` to make it clear that in order for the method to return the required record without throwing an exception the adapter must support server requests that retrieve single record objects.

The existing description and example suggest that out of the box `queryRecord()` will return a single record provided a query is passed that defines a unique record. However https://github.com/emberjs/data/pull/4300 added API documentation and assertions that make it clear that the uniqueness of the query result is not the only requirement; the method can only be used if the server's response data contains a single record object, not an array, even if the array contains only one element.

Ember Data's default json-api adapter, which uses endpoint definitions from the REST adapter, returns an array even if the query defines a unique record. Therefore to use `queryRecord()` successfully requires a customized adapter with its own `queryRecord()` method that is guaranteed to return a single object. The example provided in this update is from the API documentation and illustrates how this can be done.


source/localizable/models/finding-records.md
```diff
@@ -69,22 +69,42 @@ this.get('store').query('person', {
 
 ### Querying for A Single Record
 
-If you know your query will return only one result Ember Data provides
-a convenience method that will return a promise that resolves with a
-single record. Calling
-[`store.queryRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord)
-will make a `GET` request with the passed object serialized as query params.
+If you are using an adapter that supports server requests capable of returning a single model object, Ember Data provides
+a convenience method [`store.queryRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord) that will return a promise that resolves with that single record. The request is made via a method of the same name defined on the adapter.
```
bmac: > The request is made via a method of the same name defined on the adapter.

I think this line is a bit confusing to read. Maybe change it to be something like "The request is made via a the adapter's [queryRecord](http://emberjs.com/api/data/classes/DS.Adapter.html#method_queryRecord) method."?

bmac:
The overall content looks good. 👍

source/localizable/models/finding-records.md
```diff
@@ -69,22 +69,42 @@ this.get('store').query('person', {
...
+[`store.queryRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord)
+will retrieve that object from the server:
+
+```javascript
+store.queryRecord('user', {}).then(function(user) {
+  let username = user.get('username');
+  console.log(`Currently logged in as ${username}`);
+});
+```
+As in the case of `store.query()`, a query object can also be passed to `store.queryRecord()` and is available for the adapter's `queryRecord()` to use to qualify the request. However the adapter must return a single model object, not an array containing one element, otherwise Ember Data will throw an exception.
```
pangratz: I think it might be good to add a paragraph explaining that if the API returns an array for the endpoint (probably with only 1 element), then one could re-use the `query` behavior and do:

```js
store.query("user", {}).then(function(users) {
  return users.get("firstObject");
});
```

---

We probably don't want to advertise the usage of addons in the guide, but just for the sake of completeness, there would be one available https://emberobserver.com/addons/ember-data-query-first...

pangratz:
Thanks a bunch for tackling this @MarkMT! Much appreciated!!

source/localizable/models/finding-records.md
MarkMT: I'm ok with that, though how about "The request is made via a method `queryRecord()` defined by the adapter."? 

What I'm trying to give a little emphasis to is just that a custom method does need to be defined; the example in the [API docs](http://emberjs.com/api/data/classes/DS.Adapter.html#method_queryRecord) and indeed the default implementation of `queryRecord()` in the adapter base class are actually flawed since they use pluralized URLs created by `buildURL()`. However I'm happy to accept your judgment on this.

source/localizable/models/finding-records.md
MarkMT: Sure, happy to add that.

source/localizable/models/finding-records.md
bmac: That sounds good to me. 👍

source/localizable/models/finding-records.md
```diff
@@ -69,22 +69,55 @@ this.get('store').query('person', {
...
+
+Note that Ember's default [json-api adapter](http://emberjs.com/api/data/classes/DS.JSONAPIAdapter.html) does not provide the functionality needed to support `queryRecord()` directly as it relies on REST request definitions that return result data in the form of an array.
```
locks: JSON API instead of json-api

source/localizable/models/finding-records.md
```diff
@@ -69,22 +69,55 @@ this.get('store').query('person', {
 
 ### Querying for A Single Record
 
-If you know your query will return only one result Ember Data provides
-a convenience method that will return a promise that resolves with a
-single record. Calling
+If you are using an adapter that supports server requests capable of returning a single model object, Ember Data provides
+a convenience method [`store.queryRecord()`](http://emberjs.com/api/data/classes/DS.Store.html#method_queryRecord) that will return a promise that resolves with that single record. The request is made via a method `queryRecord()` defined by the adapter.
```
locks: newline after period please

locks:
Thank you!

MarkMT:
Thanks all for your input.

Issue #1690, opened 2016-10-07
IN simple-component page, wrong lines are marked as added and removed in app/templates/rentals.hbs snippet


shatran:
In [simple-component page](https://guides.emberjs.com/v2.8.0/tutorial/simple-component/) in app/templates/rentals.hbs snippet. It's not really a problem with the markup code, It's a problem with the way it's being displayed. Line 12 is not shown as added (+) and line 30 is not shown as removed (-) although in the markup code they are. this mentioned snippet:
```app/templates/rentals.hbs{+12,+13,-14,-15,-16,-17,-18,-19,-20,-21,-22,-23,-24,-25,-26,-27,-28,-29}
<div class="jumbo">
  <div class="right tomster"></div>
  <h2>Welcome!</h2>
  <p>
    We hope you find exactly what you're looking for in a place to stay.
  </p>
  {{#link-to 'about' class="button"}}
    About Us
  {{/link-to}}
</div>

{{#each model as |rentalUnit|}}
  {{rental-listing rental=rentalUnit}}
{{#each model as |rental|}}
  <article class="listing">
    <h3>{{rental.title}}</h3>
    <div class="detail owner">
      <span>Owner:</span> {{rental.owner}}
    </div>
    <div class="detail type">
      <span>Type:</span> {{rental.type}}
    </div>
    <div class="detail location">
      <span>Location:</span> {{rental.city}}
    </div>
    <div class="detail bedrooms">
      <span>Number of bedrooms:</span> {{rental.bedrooms}}
    </div>
  </article>
{{/each}}
```

locks:
@shatran can you send a PR for the relevant section, please?

shatran:
I can send but I'm afraid it will be just a hack around it.

locks:
This was fixed by https://github.com/emberjs/guides/commit/d472ce95c9ff2bd64828395622b587d698e85007#diff-3edfc48f157f70b0c0c757a3d1c49ace, just not deployed yet. Thanks for reporting!

PR #1691, opened 2016-10-07
Update simple-component.md

shatran:
a hacky solution to the issue described in issue #1690


toddjordan:
Seems to be working for me on chrome.  I'm afraid the "hacky" solution won't work across browsers.  I if this is a problem in certain browsers it would have to be fixed by updates to syntax highlighting code.

image

locks:
@shatran thank you for the work but I think this has already been addressed by https://github.com/emberjs/guides/commit/d472ce95c9ff2bd64828395622b587d698e85007#diff-3edfc48f157f70b0c0c757a3d1c49ace.
Can you check https://guides.emberjs.com/v2.8.0/tutorial/simple-component/ and confirm? Thank you!

PR #1693, opened 2016-10-11
Add a missing line to this particular guide

teunsteenbekkers:
The guide is missing a crucial line here; without this line of code I cannot get the example to pass the tests, or function at all for that matter.


acorncom:
Interesting.  Per our "follow-along app" (https://github.com/ember-learn/super-rentals/blob/master/tests/integration/components/location-map-test.js) it doesn't look like it's needed there ...

toddjordan:
getMapElement is called in app/components/location-map.js in the didInsertElement hook, and does not need to be called in the test. See https://github.com/ember-learn/super-rentals/blob/master/app/components/location-map.js#L9

toddjordan:
No response since Oct 13.  Closing.  Pls reopen if new data provided.

PR #1698, opened 2016-10-13
Link to super rentals repo at the beginning of the tutorial

toddjordan:
I frequently get people on slack or github issues that are stuck at various points in the tutorial.  The first thing I typically do is direct them to the super rentals repo to compare their work against.  Its worth adding a reference to the tutorial.


locks:
We should deploy super-rentals to GH, possibly on CI, and link to that as well so people can check out the final result?

toddjordan:
Sure.  I'm providing the link with fairly high frequency, so would you mind deploying this first and then creating an issue so we can circle back and work on public deployment?

acorncom:
LGTM

PR #1700, opened 2016-10-13
Change example of computed property example to avoid shared state confusion

### LINKED ISSUE #1699: Computed properties example uses unexpected shared state in component
Hi,

Looking at the example on https://github.com/emberjs/guides/blob/master/source/localizable/object-model/computed-properties-and-aggregate-data.md:

```
  todos: [
    Ember.Object.create({ isDone: true }),
    Ember.Object.create({ isDone: false }),
    Ember.Object.create({ isDone: true })
  ],

  incomplete: Ember.computed('todos.@each.isDone', function() {
    var todos = this.get('todos');
    return todos.filterBy('isDone', false);
  })
});
```

is not clear that the value of `todos` will be shared across all instances of the component in the app. Even if the component is only used once in the application, multiple running tests will share that state and produce undesired effects.

It might be worth thinking if we could modify the example to be:

```
  todos: Ember.computed(function() {
    return [
      Ember.Object.create({ isDone: true }),
      Ember.Object.create({ isDone: false }),
      Ember.Object.create({ isDone: true })
    ];
  }),

  incomplete: Ember.computed('todos.@each.isDone', function() {
    var todos = this.get('todos');
    return todos.filterBy('isDone', false);
  })
});
```

To avoid the side effects mentioned above.

juanazam:
Closes #1699

### DIFF
--- a/source/localizable/object-model/computed-properties-and-aggregate-data.md
+++ b/source/localizable/object-model/computed-properties-and-aggregate-data.md
@@ -6,11 +6,15 @@ To facilitate this, Ember provides the `@each` key illustrated below:
 
 ```app/components/todo-list.js
 export default Ember.Component.extend({
-  todos: [
-    Ember.Object.create({ isDone: true }),
-    Ember.Object.create({ isDone: false }),
-    Ember.Object.create({ isDone: true })
-  ],
+  todos: null,
+
+  init() {
+    this.set('todos', [
+      Ember.Object.create({ isDone: true }),
+      Ember.Object.create({ isDone: false }),
+      Ember.Object.create({ isDone: true }),
+    ]);
+  },
 
   incomplete: Ember.computed('todos.@each.isDone', function() {
     var todos = this.get('todos');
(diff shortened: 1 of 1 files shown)


source/localizable/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -6,11 +6,13 @@ To facilitate this, Ember provides the `@each` key illustrated below:
 
 ```app/components/todo-list.js
-  todos: [
-    Ember.Object.create({ isDone: true }),
-    Ember.Object.create({ isDone: false }),
-    Ember.Object.create({ isDone: true })
-  ],
+  todos: Ember.computed(function() {
```
locks: It is preferable to initialize this sort of data in an init, like so:

```
  todos: null,

  init() {
    this.set('todos', [
      Ember.Object.create({ isDone: true }),
      Ember.Object.create({ isDone: false }),
      Ember.Object.create({ isDone: true }),
    ]);
  }
});
```

juanazam: 👍

juanazam: Done!

locks:
Thanks for your contribution!

@homu r+

PR #1706, opened 2016-10-19
Removed uls surronding rental-listing render

albabar:
This `ul` and `li` is breaking the design. I also don't see any need for them to be there!


toddjordan:
hi @albabar thanks for contributing. Could you give more detail on how the design is breaking?  Like before/after screenshot?  Typically you would use list elements to describe lists in an web app.

albabar:
Oops, sorry. My mistake. I think that day my CSS file was not being picked up, or I made mistake elsewhere.
Couldn't reproduce today.

PR #1707, opened 2016-10-21
Typo: component generation

lisaychuang:
Ember tutorial: Building a Complex Component

Corrected line 11 typo: we are generating a new Component template, not "Handlebar" template.

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -8,7 +8,7 @@ We'll call this component `list-filter`, since all we want our component to do i
 ember g component list-filter
 ```
 
-As before, this creates a Handlebars template (`app/templates/components/list-filter.hbs`),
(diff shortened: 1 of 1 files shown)


locks:
Hi @lisaychuang. The current text is correct, as a component is made up of a Handlebars template file and a JavaScript file, but I do understand the confusion this might cause. What do you say we split the difference and only say that it "this creates a template"?

mike-north:
Agree with @lisaychuang that the terms used to refer to certain things are inconsistent throughout the tutorial. Yes it is a Handlebars template, and yes it does belong to the component in this case. Can we pick whichever one(s) best serve the purpose of making the concept clear, and stick with it?

locks:
We're editing the tutorial top to bottom and this is in the list of concerns. We'll track it there so I'm closing this one for now, thank you for the feedback!

PR #1711, opened 2016-10-24
modified punctuation and style

srukle:
- semi-colons are stylized and often confusing in technical docs
- removed semi-colon, replaced with period
- added to lines to separate titles and code
- kept capital case for Click
- added comma between quotation marks to keep MLA American style
- suggestion: if you want to use BBC style, move comma outside "Allow access"


locks:
Thanks for your contribution!

@homu r+

Issue #1710, opened 2016-10-22
Sharing Component Data / Using Block Params


MarkMT:
I think there are a number of issues with the way the use of block parameters with components is described that need addressing. I'd like to do some work on this but want to summarize my concerns before I get started in case others have some input.

1.. Organization
Block parameters are currently dealt with in two different places:

- [Sharing Component Data with its Wrapped Content](https://guides.emberjs.com/v2.8.0/components/wrapping-content-in-a-component/#toc_sharing-component-data-with-its-wrapped-content) under page heading [Components - Wrapping Content in a Component](https://guides.emberjs.com/v2.8.0/components/wrapping-content-in-a-component/). This deals principally with contextual components yet uses concepts that are not introduced till much later in:

- [Components - Using Block Params](https://guides.emberjs.com/v2.8.0/components/block-params/)

I propose to get rid of the latter page and incorporate the material it contains into the former, so the idea of sharing data with a block can be made clear in the general sense before contextual components are discussed as a particular case. Currently, under the heading [Sharing Component Data with its Wrapped Content](https://guides.emberjs.com/v2.8.0/components/wrapping-content-in-a-component/#toc_sharing-component-data-with-its-wrapped-content), the only examples provided involve sharing of components.

2.. Terminology
In  [Components - Using Block Params](https://guides.emberjs.com/v2.8.0/components/block-params/), the terminology of "return values" from a component to describe block parameters seems like a poor choice of words. In the context of functions or methods, return values represent what is returned to the calling context after the code in question has terminated, but that's not what is happening here. The construct is essentially the same as a block in ruby, where parameters passed to the block are said to be yielded.

3.. Component Naming
Under [Sharing Component Data with its Wrapped Content](https://guides.emberjs.com/v2.8.0/components/wrapping-content-in-a-component/#toc_sharing-component-data-with-its-wrapped-content) there's an example of a component used as follows:
```
{{#blog-post editStyle="markdown"}}
  <p class="author">by {{author}}</p>
  {{body}}
{{/blog-post}}
```
and defined as:
```
<h2>{{title}}</h2>
<div class="body">{{yield (hash body=(component editStyle))}}</div>
```
However this seems to amount to the component being specified with the name `markdown`, which is invalid without a dash.

4.. Variable Naming
In the preceding example, `editStyle` is said to represent a choice between 'html' and 'markdown' and is used to define the component that is used in the block. But in this example:
```
{{#blog-post editStyle="markdown" as |post|}}
[...]

PR #1714, opened 2016-10-28
Revise "Defining a Component" and "Wrapping Content in a Component"

MarkMT:
This PR addresses https://github.com/emberjs/guides/issues/1710, and covers a number concerns about the organization of the "Components" guide, particularly with respect to the treatment of the block component usage, including block parameters and contextual components. These changes impact the following pages:
- [Defining a Component](https://guides.emberjs.com/v2.8.0/components/defining-a-component/) (revised)
- [Passing Properties to a Component](https://guides.emberjs.com/v2.8.0/components/passing-properties-to-a-component/) (removed)
- [Wrapping Content in a Component](https://guides.emberjs.com/v2.8.0/components/wrapping-content-in-a-component/) (revised)
- [Using Block Params](https://guides.emberjs.com/v2.8.0/components/block-params/) (removed)

**Defining a Component**
Currently, separate pages in the guide deal with passing of properties to a component and the use of components in block form. However the introductory examples in the first section, "Defining a Component", use and presume an understanding of both of these concepts, despite them not being presented in depth until later on.

To address this and other minor issues, this PR:
- adds a brief description of what a component is to "Defining a Component".
- removes block examples from "Defining a Component", deferring this to "Wrapping Content in a Component"
- uses initial examples in "Defining a Component" that involve both static HTML and passed properties in order to illustrate the basic concepts of component definition in the context of inline usage. Because passing of properties is explained here, the first half of the material in the page "Passing Properties to a Component" is redundant. One paragraph in the latter on binding is moved into "Defining a Component." The material on positional parameters is moved as well and "Passing Properties to a Component" is removed.
- adds a paragraph about component files, including the javascript file and `Ember.Component` subclass, immediately following the explanation of component generation and naming, rather than deferring mention of javascript to the existing section "Defining a Component Subclass", which currently comes after the initial examples. The rationale is two-fold. First it seems useful to note that the `generate` command will create skeleton files for the component (so the reader understands what they see in their terminal when they run it). So some mention of what those files are seems appropriate. Secondly the notion that a component template has a javascript context has been introduced as a fairly fundamental concept of Ember back in "Templates - Handlebars Basics" under "Displaying Properties" as well as in "Templates - Actions." So it seems reasonable to make the connection with what the reader has already been exposed to at a fairly early point in the discussion rather than later. A couple of sentences from the existing section "Defining a Component Subclass" have been moved in to the new paragraph and that section has been removed.
- makes minor changes to the section "Dynamically Rendering a Component," e.g. adding text noting the ability to specify properties to be passed to a dynamically selected component.

Note: As a result of these changes, the page "Defining a Component" is reasonably lengthy. I think a case could possibly be made that the section "Dynamically Rendering a Component" maybe doesn't need to be included here. I wouldn't be opposed to putting it on a page by itself, but I've left that for separate consideration.

**Wrapping Content in a Component**
In addition to the above, changes have been made to the treatment of block usage of components in "Wrapping Content in a Component". This page currently has a section "Sharing Component Data with its Wrapped Content" that has a couple of related problems:
- the only example given involves sharing of components. Despite the section heading, there is almost no indication of the broader notion of sharing component data with wrapped content.
- while the example (rightly) makes use of and requires an understanding of block parameters, the concept of block parameters is not discussed in depth until a separate later page "Using Block Params."

Also, the use of `hasBlock` to support both block and inline component usage is currently presented only on page "Using Block Params" when in fact it is relevant to any component that accepts a block, not just one that yields parameters to the block.

To address these and other minor issues, this PR:
- revises the introductory text and example for "Wrapping Content in a Component" to align with the previous inline example given in "Defining a Component." (Currently the page begins with "Sometimes, you may want to define a component that wraps content provided by other templates. For example,..." but then shows an _inline_, rather than block, example)
- slightly revises the first block-style example, again to align with the initial example which passes a `post` object.
- adds a discussion of block parameters in general to the section "Sharing Component Data with its Wrapped Content" where it is needed, rather than in the later page "Using Block Params," which has now been removed.
- moves the discussion of `hasBlock` out of the context of block parameters and inserts it following the overview of block usage in "Wrapping Content in a Component". The example has been changed slightly to reflect this.
- adds a note highlighting the relationship between block usage of components and Ember's internal block helpers.
- creates a separate section heading "Contextual Components" for the material dealing with that topic
- revises the contextual component example to improve clarity. The current example adds elements of the final solution progressively without making clear that that is what is happening. As a result, for example, the final listing for `blog-post.hbs` includes a variable `postData` with no indication of where it came from, while the listing of index.hbs before that has no representation of the data to be rendered at all. Also, the current example includes an invalid component name value `markdown`, and uses the variable `editStyle` to represent two different things, with the second actually being a text style configuration.
- in the contextual component example adds a description of what the `hash` helper does and also references the earlier section on the `component` helper contained in "Defining a Component".


MarkMT:
https://github.com/emberjs/guides/pull/1715 made changes to a file that I'd removed. I have merged master, which involved undeleting that file, and then re-removed it.

source/localizable/components/defining-a-component.md
```diff
@@ -41,37 +68,73 @@ export default Ember.Route.extend({
   }
 });
 ```
+Note that the properties passed to the component stay in sync (technically known as being "bound"). That is, if the value of componentProperty changes in the component, outerProperty will be updated to reflect that change. The reverse is true as well.
```
MarkMT: Even if it's an anti-pattern, I think it's worth highlighting so that developers are aware of potential side-effects.

MarkMT:
Thanks for the input. Changes added.

MarkMT:
Conflict with upstream changes resolved.

locks:
Hi @MarkMT, a quick thing. You should leave the translation files untouched as mentioned in [CONTRIBUTING.md](https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md#style-guide), as they are automatically synchronized and managed off-repository.

A not-so-quick thing, table of content changes are to be considered breaking changes similarly to changing public API. We do not have a process in place, neither do we explicitly mention this, but we're hoping to document an RFC process soon to help contributors. However, regardless of TOC changes, you should coordinate with the learning team as early in the process as possible to make sure the direction is aligned, to avoid re-work, and to make the reviewing process more streamlined.

MarkMT:
Thanks @locks. I guess I should have realized about the translation files. I have restored them and also the TOC. The TOC also required the 2 deleted chapters to pass the build test, so I have restored them also and just deleted their contents.

If there is a better way to coordinate I'm happy to do whatever needs to be done. I did endeavor in https://github.com/emberjs/guides/issues/1710 to give as much detail about my thinking as there was in my head before I started work on this and I did flag the likely removal of the Block Params page.

I have further fairly substantial revisions on the drawing board related to action handling and I'd like again to submit an issue outlining my thinking ahead of a PR. However in this case I'm finding that I'm having to start actively writing in order to be able to organize my thoughts about what needs to be done well enough that I can put them in an issue.

sivakumar-kailasam:

PR #1715, opened 2016-10-29
Shift from triple curlies to double curlies

acorncom:


locks:
Thanks for your contribution!

@homu r+

PR #1716, opened 2016-11-01
Fixed the layout grid

mellatone:
Updated the grid for the Guides to match the rest of the Ember.js sites. The jump can be a bit jarring 😉


locks:
👌 

@homu r+

PR #1720, opened 2016-11-04
remove new code line marker

mdarmetko:
we're editing a newly created file, so the marker is not needed and confusing


acorncom:
@homu r+  Thanks @mdarmetko!

PR #1721, opened 2016-11-05
added link-ext icon

mellatone:
Added the following icon from Fontello

`.icon-link-ext`

image


locks:
Thanks for your contribution!

@homu r+

PR #1724, opened 2016-11-07
consistent use of this.get('store')

davidpett:
the `subroute` step in the tutorial uses `this.store.find...` where all others use `this.get('store')`


locks:
Thanks for your contribution!

@homu r+

PR #1728, opened 2016-11-11
Explain what an outlet is.

sheriffderek:
I'm not sure of the best way to word it, but the outlet should be explained. I pulled this copy from 2.7's guide.

### DIFF
--- a/source/localizable/getting-started/quick-start.md
+++ b/source/localizable/getting-started/quick-start.md
@@ -132,6 +132,8 @@ In your browser, open
 see the `<h2>` you put in the `scientists.hbs` template, right below the
 `<h1>` from our `application.hbs` template.
 
+Templates are rendered into the {{outlet}} of its parent route's template.
+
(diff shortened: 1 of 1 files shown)


locks:
I think the explanation would be better placed after https://github.com/emberjs/guides/pull/1728/files#diff-41d19b956479ca7b821c55def80da82bR89. Something like: "When we add routes later on, they will be rendered inside `{{outlet}}`".
What do you think?

sheriffderek:
When you say, ..."better placed after..." I'm not sure where you mean, but anyplace is fine with me. I just don't want it to go unmentioned. I was showing someone the walkthrough and they were thrown off by that.

locks:
> When you say, ..."better placed after..." I'm not sure where you mean, but anyplace is fine with me.

I mean line 89 of the document, after "You should see that the welcome page has been replaced by "PeopleTracker".

locks:
Thank you for your help!

PR #1729, opened 2016-11-11
Clarify context in discussion of component actions

MarkMT:
This change clarifies the point that the discussion following is specifically about actions that are handled by an external context, rather than just being handled within the component in which the event is detected.


locks:
@homu r+

PR #1730, opened 2016-11-11
Clarify wording on effect of component action

MarkMT:
To a new reader, the current wording "have the action delete the user's account in one place, and send a message in another place" could be misconstrued as describing two things you want to have happen when the action occurs. i.e. "one place...another place" could be taken to mean different parts of the code rather than different results of separate events occurring in different places in the application where the component is used.


locks:
@homu r+

PR #1731, opened 2016-11-11
Clarify context of action implementation

MarkMT:
This is a minor change to help the reader parse the discussion that follows. The fact that two different usages of the component are planned has already been introduced in the section "Creating the Component." Just saying "in this case" is potentially confusing given that two different cases have been foreshadowed and the second case of message sending is discussed later in the presentation.


locks:
Thanks for your contribution!

@homu r+

PR #1732, opened 2016-11-11
Emphasize change of context to child component

MarkMT:
This is a minor change to help the reader parse the explanation of how component actions are implemented. In the preceding section we have already said "Let's look at the parent component's JavaScript file". This proposed change just emphasizes to the reader that we are now shifting focus to the internal implementation of the component.


locks:
@homu r+

PR #1734, opened 2016-11-12
Clarify wording on passing an action to a component

MarkMT:
Currently the section headed "Passing the Action to the Component" begins with the sentence "Now we need to make it so that the `onConfirm()` event in the button-with-confirmation component triggers the `userDidDeleteAccount()` action..." However, since `onConfirm` hasn't yet been defined, the use of the definite article makes it challenging for a reader unfamiliar with the material to follow the thought process.

The proposed wording change avoids referring to `onConfirm` until after the example that introduces it. Instead we simply explain that we are going to pass the parent action to the component as a property. A note is also added to explicitly highlight the role that the `action` helper plays in the process.

### DIFF
--- a/source/localizable/components/triggering-changes-with-actions.md
+++ b/source/localizable/components/triggering-changes-with-actions.md
@@ -133,20 +133,18 @@ based on the value of `confirmShown`.
 
 ## Passing the Action to the Component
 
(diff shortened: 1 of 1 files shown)


locks:
Thanks for your contribution!

@homu r+

Issue #1727, opened 2016-11-10
Event handling


MarkMT:
There are some issues related to the way ember's handling of events is described in the guides that I think need some attention. Part of my concerns were detailed previously in https://github.com/emberjs/guides/issues/1345. However one aspect of that issue (the use of `sendAction` to invoke actions in a component) has since been addressed in https://github.com/emberjs/guides/commit/4f6e84ad701e03ef85e87be2299252f0d0197364 and there are some additional related concerns not mentioned there that I think also need improvement. So the following is intended to capture a summary of my current concerns as a point of reference for a coming PR.

Currently event handling is dealt with in three places:
1. [Templates - Actions](https://guides.emberjs.com/v2.9.0/templates/actions/) - Deals with the use of actions to respond to events on an HTML element within a single execution context, such as a controller or component.
2. [Components - Handling Events](https://guides.emberjs.com/v2.9.0/components/handling-events/) - Deals with the use of event handlers (as opposed to actions) to respond to events on a component as a whole, plus some 'other stuff.'
3. [Components - Triggering Changes with Events](https://guides.emberjs.com/v2.9.0/components/triggering-changes-with-actions/) - Deals with the use of external actions to respond to events within a component.

**Issues**
In [Templates - Actions](https://guides.emberjs.com/v2.9.0/templates/actions/), two different usages of the `action` helper are presented in different sections. The initial examples use the helper in the element space context (e.g. `<button {{action "toggleBody"}}>`;  terminology from the [API docs](http://emberjs.com/api/classes/Ember.Templates.helpers.html#method_action)), while later there is a discussion under "Modifying the action's first parameter" that uses an example of the helper in the attribute context to illustrate the use of the `value` option (i.e. `onblur={{action "bandDidChange"}}`). However there is no mention of the fact that the latter example represents a different usage from the one presented earlier or of its significance. There is no explanation of the fact that using the `action` helper in the attribute context produces a closure action or what that means. While the example that uses the `value` option mentions the fact that the action in this context takes the event object as an argument, since the fact that this usage is different from the original examples is never mentioned, it's not at all clear to a new reader that this (i.e. the use of the event object) applies specifically to that context and not to the element space context.

Because there is no identification of closure actions as being distinct from those defined in the element space context, there is also no discussion of the use of explicit action arguments in that context and no indication of how explicit arguments and the implicit event argument relate to one another, i.e. what order they get passed to the action.

[...]

PR #1735, opened 2016-11-13
Clarify passing of action arguments

MarkMT:
This change seeks to clarify the discussion under the heading "Passing Arguments," addressing the following concerns:

- The current text uses an incorrect action name: "For example, we'll update the `send-message` action" - should be `sendMessage`.
- The wording used in the current text to describe how arguments are handled is potentially confusing for readers not already familiar with the material. For example, it says "the action helper may be invoked with arguments," but later "It will still invoke `onConfirm` with no arguments," followed by "The action helper will add the arguments provided in the template to the call" (Added how? Which template?). The point is clear to people familiar with these concepts, but is very likely to cause difficulty to new developers.
- The sentence "Since the `button-with-confirmation` component doesn't know or care about what type of message its collecting, we want to provide a message type from `send-message` when we define the action," is potentially confusing for a couple of reasons:

  - the message type is provided when _passing_ the action to the component, not when _defining_ it, if we understand the action to be defined in the `actions` hook.
  - the logical connection between the two clauses is unclear unless you already understand the overall process. Naively, taken at face value, if the component doesn't care, why would it be important to provide it?
- minor typo in 4th sentence: "its" instead of "it's"

### DIFF
--- a/source/localizable/components/triggering-changes-with-actions.md
+++ b/source/localizable/components/triggering-changes-with-actions.md
@@ -255,14 +255,14 @@ For example, if we want to use the button to send a message of type `"info"`:
 {{button-with-confirmation text="Click to send your message." onConfirm=(action "sendMessage" "info")}}
 ```
 
-Within `button-with-confirmation`, the code does not change.
+Within `button-with-confirmation`, the code in the `submitConfirm` action does not change.
 It will still invoke `onConfirm` without explicit arguments:
 
 ```app/components/button-with-confirmation.js
 const promise = this.get('onConfirm')();
 ```
 However the expression `(action "sendMessage" "info")` used in passing the action to the component creates a closure,
-binding the parameter we've provided to the function specified.
+i.e. an object that binds the parameter we've provided to the function specified.
 So now when the action is invoked, that parameter will automatically be passed as its argument, effectively calling `sendMessage("info")`,
 despite the argument not appearing in the calling code.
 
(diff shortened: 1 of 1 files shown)


source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -232,18 +232,39 @@ export default Ember.Component.extend({
...
 ```
 
-In this case, the code in `button-with-confirmation` does not change.
-It will still invoke `onConfirm` with no arguments.
-The action helper will add the arguments provided in the template to the call.
+Within `button-with-confirmation`, the code does not change.
+It will still invoke `onConfirm` without explicit arguments:
+
+```app/components/button-with-confirmation.js
+const promise = this.get('onConfirm')();
```
locks: Where is this code in the component?

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -232,18 +232,39 @@ export default Ember.Component.extend({
...
+```
+However the expression `(action "sendMessage" "info")` used in passing the action to the component creates a closure,
```
locks: What's a closure?

source/localizable/components/triggering-changes-with-actions.md
MarkMT: I've added some clarification in the wording. I'm happy to provide more context in the code block itself if you wish, though my intent here was to draw attention specifically on that one line.

source/localizable/components/triggering-changes-with-actions.md
MarkMT: I've modified the wording to try to clarify this. Part of my reason for mentioning this at all is that the terminology is used in the API docs in distinguishing two fundamentally different usages of the `action` helper. As noted in https://github.com/emberjs/guides/issues/1727 the guides currently lack any explanation of these two usages. I currently have another PR in preparation that will address this.

PR #1736, opened 2016-11-13
Clarify use of an action with a curried argument

MarkMT:
This change concerns the second, extended example covered under the heading "Passing Arguments" in the [Components - Triggering Changes With Actions](https://guides.emberjs.com/v2.9.0/components/triggering-changes-with-actions/). This is a challenging example for readers unfamiliar with the material partly because it requires changes to multiple files to make it work and also because it relies on a concept, currying, that will likely not be familiar to many new Ember developers.

A couple of examples of how the existing presentation may trip readers up:

- This paragraph... "Action arguments curry, meaning that you can provide partial arguments to the action helper and provide the rest of the arguments when you call the function within the component javascript file. For example, our `button-with-confirmation` component will now yield the content of the confirmation dialog to collect extra information..."

  - The context isn't clear. It's not obvious what the purpose is of providing additional arguments or what these represent ("extra information"?). It's not until the very last listing in the example that it becomes clear that we are trying to send the message text along with the call to `sendMessage`.
  - At face value, this sounds like either "yielding content" or "collecting extra information" is an example of "providing partial arguments to the action and the rest of the arguments in the javascript call". Clearly that's not exactly the case. Rather they are elements of an overall solution in which currying will play a role.
- Because of the order in which the elements of the solution are presented, it's difficult initially to see why the property `confirmValue` is introduced. It's significance doesn't become evident until the listing for 'button-with-confirmation.js' is shown well into the example.

The proposed update starts the example by explicitly noting the goal of providing the message text to the external action when the confirmation button is clicked. It explains why currying is relevant to this objective and explicitly notes that there are multiple steps involved in accomplishing the objective and unpacks each of them in turn.

### DIFF
--- a/source/localizable/components/triggering-changes-with-actions.md
+++ b/source/localizable/components/triggering-changes-with-actions.md
@@ -245,10 +245,60 @@ In this case, the code in `button-with-confirmation` does not change.
 It will still invoke `onConfirm` with no arguments.
 The action helper will add the arguments provided in the template to the call.
 
-Action arguments curry, meaning that you can provide partial arguments to the action helper and provide the rest of the
-arguments when you call the function within the component javascript file.
-For example, our `button-with-confirmation` component will now [yield](../wrapping-content-in-a-component/) the content
-of the confirmation dialog to collect extra information to be sent along with the `onConfirm` action:
+So far in our example,
+the action we have passed to `button-with-confirmation` is a function that accepts one argument, `messageType`.
+Suppose we want to extend this by allowing `sendMessage` to take a second argument,
+the actual text of the message the user is sending:
+
+```app/components/send-message.js
+export default Ember.Component.extend({
+  actions: {
+    sendMessage(messageType, messageText) {
+      //send message here and return a promise
+    }
+  }
+});
+```
+
+We want to arrange for the action to be invoked from within `button-with-confirmation` with both arguments.
+A feature of action arguments that applies to this situation is that they _curry_.
+In general (not just in the context of Ember), this refers to the idea that a function with multiple arguments can be expressed as a sequence of functions each taking one argument and returning the next function in the sequence that takes the next argument,
+and so on.
+Evaluating the last function produced in this way is equivalent to evaluating the original function with all of its arguments.
+
+For our purposes, currying means that when we pass the action to a component, if we provide the action helper with a subset of arguments (e.g. by providing the value of `messageType` when we pass `sendMessage` to `button-with-confirmation`), the result of the helper (in our case `onConfirm`) is a function that can be invoked within the component javascript file with just the remaining arguments,
+in our case `messageText`.
+
+Remember, however,
+that our `button-with-confirmation` component does not know or care that it is being used here in a messaging application.
+So to enable it to call the action represented by `onConfirm` with the required remaining argument,
+we will do three things:
+
+_First_, in the component we will use a property `confirmValue` to contain a parameter to be passed to `onConfirm`:
+
+```app/components/button-with-confirmation.js
+export default Ember.Component.extend({
+  actions: {
+    //...
+    submitConfirm() {
+      //call onConfirm with a second argument
+      const promise = this.get('onConfirm')(this.get('confirmValue'));
+      promise.then(() => {
+        this.set('confirmShown', false);
+      });
+    },
+    //...
+  }
+});
+```
+
+When we invoke `onConfirm` with this as an explicit argument,
+because of currying that argument will be passed to `sendMessage` as its _second_ argument.
+Although in our particular example, `confirmValue` will be used to pass the message text,
+using this ensures that the implementation of the component remains independent of the specific purpose it is being used for.
+
+_Second_,
+we will modify the component so that it can be used in block form and we will [yield](../wrapping-content-in-a-component/) `confirmValue` to the block.
 
 ```app/templates/components/button-with-confirmation.hbs
 <button {{action "launchConfirmDialog"}}>{{text}}</button>
@@ -261,8 +311,11 @@ of the confirmation dialog to collect extra information to be sent along with th
 {{/if}}
 ```
 
-The `send-message` component provides an input as block content to the `button-with-confirmation` component, setting
-`confirmValue`.
+Note that `yield` is placed inside the `"confirm-dialog"` element, so that when the component is used,
+the content contained in the block will only appear once the button is clicked.
+
+_Third_, in `send-message`, we will use the component to wrap a text input element whose `value` is `confirmValue`.
+When the user enters their message into the input element, that text will be available to the component as `confirmValue`.
 
 ```app/templates/components/send-message.hbs
 {{#button-with-confirmation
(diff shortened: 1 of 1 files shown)


source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
 
-Now when the `submitConfirm` action is invoked, we call it with the value provided by our yielded input.
+Remember, however,
+that our `button-with-confirmation` component does not know or care that it is being used here in a messaging application.
+So to enable it to call the action represented by `onConfirm` with the required remaining argument,
+we will do three things:
 
-```app/components/button-with-confirmation.js
-import Ember from 'ember';
+_First_, in the component we will use a property `confirmValue` to contain a parameter to be passed to `onConfirm`:
```
locks: Why is this italics?

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
+      //send message here and return a promise
+    }
+  }
+});
 ```
 
-The `send-message` component provides an input as block content to the `button-with-confirmation` component, setting
-`confirmValue`.
+We want to arrange for the action to be invoked from within `button-with-confirmation` with both arguments.
+A feature of action arguments that applies to this situation is that they _curry_.
```
locks: I don't agree with introducing this concept.

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
+In general (not just in the context of Ember), this refers to the idea that a function with multiple arguments can be expressed as a sequence of functions each taking one argument and returning the next function in the sequence that takes the next argument,
```
locks: Unnecessarily confusing.

toddjordan:
I think the order is generally ok as is.  We show the declaration of the top level component, introducing the thing we are explaining (button-with-confirmation with an extra argument)

We then show each template, where/how the action is generated (from the input and clicking submitConfirm).
Next we show the javascript code that handles form submit and invokes the action,
Finally we show the parent component handling the action with both arguments.

I think the curry paragraph discusses it in the context of you you use it in ember actions, which is what we are trying to communicate.  If anything we should send the reader to further reading with a link on the word curry, but I think trying to explain the general concept here is more confusing than it already is.

That all said, I'm will to go with go with more changes if @locks or @acorncom agree its necessary (maybe I'm too close to the subject), but my opinion is that the structure is ok apart from some clarity tweaks.

MarkMT:
I can only speak as a fairly new user who has had to wrestle with the current guide fairly recently. I found the presentation extremely difficult to parse. I think the current ordering as @toddjordan outlines it makes sense if you already know where the narrative is going. However if you're not familiar with the material, the fact that this example is designed to invoke an action with two arguments doesn't crystallize until the very last listing. It is understanding that objective that makes the rest of the example make sense. The wording "collect extra information to be sent along with the `onConfirm` action" by itself really doesn't convey this clearly. This is why I propose pulling the send-message.js listing forward to make the goal explicit from the beginning.

The second challenge is that in the two template listings, `confirmValue` plays a critical role. But it doesn't become clear what that role is until you see the modified call to `onConfirm` in the javascript code. For me, seeing that call first enabled me to understand why the property was being introduced into the templates.

Added to this is the fact that there are multiple steps needed in making this work. Calling out the "_First_", "_Second_", "_Third_" elements of this explicitly is just designed to help the reader mentally navigate what is going on, with the italics adding emphasis to that structure. There's no easy way to number or bulletize these sections.

I agree that the currying discussion is awkward. While not inherently complex, it's a challenging concept if we only have a sentence or two in which to get the idea across. However I don't think we can avoid it altogether as it is fundamental to understanding how you can call a function that apparently takes two arguments but specify only one of them.

I would be happy, for example, to take out the sentences describing currying in the general sense and go directly to what it signifies in Ember, e.g. either the paragraph starting "For our purposes, currying means..." or something closer to the current wording, "meaning that you can provide partial arguments to the action helper and provide the rest..."

locks:
> Added to this is the fact that there are multiple steps needed in making this work. Calling out the "First", "Second", "Third" elements of this explicitly is just designed to help the reader mentally navigate what is going on, with the italics adding emphasis to that structure. There's no easy way to number or bulletize these sections.

Descriptive sub-headers would be more appropriate here.

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
-  </div>
-{{/if}}
+```app/components/send-message.js
+export default Ember.Component.extend({
+  actions: {
+    sendMessage(messageType, messageText) {
+      //send message here and return a promise
+    }
+  }
+});
```
acorncom: I like the reworking of the text that we've got here so far, as it simplifies the explanation by explaining what we're trying to achieve.

source/localizable/components/triggering-changes-with-actions.md
acorncom: And I would agree with @locks here that this gets complex

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
-    as |confirmValue|}}
-  {{input value=confirmValue}}
-{{/button-with-confirmation}}
-```
+For our purposes, currying means that when we pass the action to a component, if we provide the action helper with a subset of arguments (e.g. by providing the value of `messageType` when we pass `sendMessage` to `button-with-confirmation`), the result of the helper (in our case `onConfirm`) is a function that can be invoked within the component javascript file with just the remaining arguments,
+in our case `messageText`.
 
-Now when the `submitConfirm` action is invoked, we call it with the value provided by our yielded input.
+Remember, however,
+that our `button-with-confirmation` component does not know or care that it is being used here in a messaging application.
```
acorncom: Nicely put

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
-```app/templates/components/send-message.hbs
-{{#button-with-confirmation
-    text="Click to send your message."
-    onConfirm=(action "sendMessage" "info")
-    as |confirmValue|}}
-  {{input value=confirmValue}}
-{{/button-with-confirmation}}
-```
+For our purposes, currying means that when we pass the action to a component, if we provide the action helper with a subset of arguments (e.g. by providing the value of `messageType` when we pass `sendMessage` to `button-with-confirmation`), the result of the helper (in our case `onConfirm`) is a function that can be invoked within the component javascript file with just the remaining arguments,
+in our case `messageText`.
```
acorncom: From a reader's point of view, do we even care what currying is?  It's a valuable concept to know about, but I'm not sure we need to introducing that concept here (as @locks mentioned earlier)

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
   actions: {
-    launchConfirmDialog() {
-      this.set("confirmShown", true);
-    },
-
+    //...
     submitConfirm() {
-      //call onConfirm with the value of the input field as an argument
+      //call onConfirm with a second argument
       const promise = this.get('onConfirm')(this.get('confirmValue'));
```
acorncom: I'd like to shift this to `let` from `const`, but we can address that in a separate PR

source/localizable/components/triggering-changes-with-actions.md
```diff
@@ -245,76 +245,90 @@ In this case, the code in `button-with-confirmation` does not change.
...
   }
 });
 ```
 
-This action will call our bound `sendMessage` function with both the message type we provided earlier, and the template
-and the message value provided in the component JavaScript.
+When we invoke `onConfirm` with this as an explicit argument,
+because of currying that argument will be passed to `sendMessage` as its _second_ argument.
+Although in our particular example, `confirmValue` will be used to pass the message text,
+using this ensures that the implementation of the component remains independent of the specific purpose it is being used for.
```
acorncom: Here again, mentioning currying may actually reduce clarity here.  Maybe something like this?

> Because the first argument needed in our `sendMessage` action has already been filled in in the `send-message` template, when we call `onConfirm` our `confirmValue` property will be be used as the second argument.  This makes it possible to re-use our `button-with-confirmation` component in different places where the action it calls may have very different method signatures.

Even my suggestion here may need further wordsmithing, but I think it conveys the same concepts in simpler language.

MarkMT:
Thanks all for the input. I've simplified the explanation significantly (I think) by avoiding any explicit explanation of what currying is, apart from a parenthetical reference. This revision also avoids the need for extra sub-headings or numbered paragraphs. I remain convinced that the order of presentation proposed here is preferable.

PR #1739, opened 2016-11-23
Update component names to have a dash

niamleeson:
Components cannot be named without a dash


toddjordan:
Thanks ❤️ @niamleeson 
@homu r+

Issue #1738, opened 2016-11-23
Clarify the source of postData variable in the guide


niamleeson:
https://guides.emberjs.com/v2.9.0/components/wrapping-content-in-a-component/

In this block of code, it is not clear where the variable `postData` in `postData=postData` is coming from.
```javascript
// app/templates/components/blog-post.hbs
<h2>{{title}}</h2>
<div class="body">{{yield (hash body=(component editStyle postData=postData))}}</div>
```

Could you expand the context a little more?

Thank you.

toddjordan:
The component is showing a blog post, but yielding its data to its block so that the consumer of the component has control over how to format it. 

`postData` is essentially the text from the blog post.  Its showing that you can pass arguments into the component helper so that the consumer of the component (in this case index.hbs) doesn't have to worry about passing data.

Some text on what postData is should do, and maybe renaming it to something like `blogText`

niamleeson:
Is it something like this?
```javascript
// app/index.js
import Ember from 'ember';

export default Ember.Component.extend({
  blogTexts: [..., ..., ...]
});

// app/templates/index.hbs
{{#each blogTexts as |blogText|}}
  {{#blog-post editStyle="markdown-style" postData=blogText as |post|}}
    <p class="author">by {{author}}</p>
    {{post.body editStyle="compact-style"}}
  {{/blog-post}}
{{/each}}
```

Or more like this?
```javascript
// app/templates/index.hbs
{{#blog-post editStyle="markdown-style" as |post|}}
  <p class="author">by {{author}}</p>
  {{post.body}}
{{/blog-post}}

// app/components/blog-post.js
import Ember from 'ember';

export default Ember.Component.extend({
  blogText: '...'
});

// app/templates/components/blog-post.hbs
<h2>{{title}}</h2>
<div class="body">{{yield (hash body=(component editStyle postData=blogText))}}</div>
```

Thank you.

toddjordan:
A blog post would have one text, so the second.  It might be simpler to bind the text down like this:

```javascript
// app/templates/index.hbs
{{#blog-post editStyle="markdown-style" text=myText as |post|}}
  <p class="author">by {{author}}</p>
  {{post.body}}
{{/blog-post}}

// app/templates/components/blog-post.hbs
<h2>{{title}}</h2>
<div class="body">{{yield (hash body=(component editStyle text=text))}}</div>
```

niamleeson:
[...]

PR #1740, opened 2016-11-25
Clarify the source and usage of postData variable.

niamleeson:
I think newcomers will benefit from an extended explanation of the block component.

### DIFF
--- a/source/localizable/components/wrapping-content-in-a-component.md
+++ b/source/localizable/components/wrapping-content-in-a-component.md
@@ -81,7 +81,7 @@ To load a different body component based on editing style, you can yield the com
 <div class="body">{{yield (hash body=(component editStyle))}}</div>
 ```
 
-Once yielded the data can be accessed within wrapped content by referencing the `as` variable.
+Once yielded the data can be accessed within wrapped content by referencing the `as` variable. Now a component called `markdown-style` will be rendered in `{{post.body}}`.
 
 ```app/templates/index.hbs
 {{#blog-post editStyle="markdown-style" as |post|}}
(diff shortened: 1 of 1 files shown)


source/localizable/components/wrapping-content-in-a-component.md
```diff
@@ -93,6 +93,14 @@ Once yielded the data can be accessed within wrapped content by referencing the
 Finally we want to share the model of the data a user fills out for the post within our `blog-post` and body components.
 To share the `postData` object with the new body component, you can add arguments to the component helper.
```
toddjordan: "Finally, we want to share the the blog text that a user fills out for the post within our `blog-post` and body components. To share the blog text with the new body component, we'll add a `postData` argument to the component helper."

source/localizable/components/wrapping-content-in-a-component.md
```diff
@@ -93,6 +93,14 @@ Once yielded the data can be accessed within wrapped content by referencing the
...
 
+```app/components/blog-post.js
```
toddjordan: I think binding postData directly into the blog-post component declaration above would be better than giving it internal state.

source/localizable/components/wrapping-content-in-a-component.md
```diff
@@ -105,9 +113,26 @@ When `{{post.body}}` is instantiated, it will have both the edit style and the `
 ```app/templates/index.hbs
 {{#blog-post editStyle="markdown-style" as |post|}}
   <p class="author">by {{author}}</p>
-  {{post.body editStyle="compact-style"}}
+  {{post.body bodyStyle="compact-style"}}
+{{/blog-post}}
+```
+
+Alternatively, `postData` can be passed to `blog-post` at the block level if it was made available outside the component block as `myText`.
```
toddjordan: I don't think we need to give both alternatives here.  Just use this as the main example (like commented above).  It distracts from what the main point of the section is.

source/localizable/components/wrapping-content-in-a-component.md
niamleeson: Thank you for the improvement. The text has been updated.

source/localizable/components/wrapping-content-in-a-component.md
niamleeson: Removed the alternative as you suggested.

toddjordan:
Thanks, @niamleeson 

@homu r+

toddjordan:
#1738

Issue #1333, opened 2016-03-17
Improve https://guides.emberjs.com/v2.4.0/routing/defining-your-routes/#toc_wildcard-globbing-routes


locks:
Mention how to access the wildcard segment.

acorncom:
@locks Could you provide a little more detail here so we can have other folks help here?

jeradg:
You mean something like:

jeradg:
You mean something like:

```js
// app/router.js
Router.map(function() {
  this.route('page-not-found', { path: '/*wildcard' });
});

// app/routes/page-not-found.js
export default Ember.Route.extend({
  model(params) {                  
    return params;                 
  }                                
});                                

// app/templates/page-not-found.hbs

<p>
  Path '{{model.wildcard}}' not found.
</p>
```

jeradg:
(And yes, that's me volunteering to make a PR.)

locks:
@jeradg waiting for that PR ;)

yawboakye:
@locks as mentioned on Slack I'm taking over. I'll submit the PR.

acorncom:
Fixed in #1741

PR #1741, opened 2016-11-29
Add a wildcard route example

yawboakye:
Address https://github.com/emberjs/guides/issues/1333


locks:
You might want to rebase because of your previous PR to add a leading / to paths.

source/localizable/routing/defining-your-routes.md
```diff
@@ -250,16 +249,23 @@ In the next section, [Specifying a Route's Model](../specifying-a-routes-model),
...
-You can define wildcard routes that will match multiple URL segments. This could be used, for example,
-if you'd like a catch-all route which is useful when the user enters an incorrect URL not managed
-by your app.
+You can define wildcard routes that will match multiple URL segments.
+This could be used, for example, if you'd like a catch-all route which is useful when the user enters an incorrect URL not managed by your app.
+Wildcard routes begin with an asterisk.
 
 ```app/router.js
 Router.map(function() {
-  this.route('page-not-found', { path: '/*wildcard' });
```
yawboakye: Changed from `page-not-found` to `not-found` because that's what I found a lot. I changed `*wildcard` to `*path` for 2 reasons:  
1. It's a wildcard path, and I thought `*path` reads as such i.e. `*` means wildcard.
2. I didn't want any possible confusion where learners took the wrong message that all wildcard routes should be called `*wildcard`. I guess this could happen with `*path` as well but at least it avoids the wildcard keyword.

PR #1742, opened 2016-12-03
Add slashes to route path

yawboakye:
Since most route paths have slashes as first character, add slashes to
the remaining.


locks:
😭

PR #1745, opened 2016-12-09
Have set( ) as recommended method to re-render each-in

akashdsouza:
As per [264023460](https://github.com/emberjs/ember.js/issues/14661#issuecomment-264023460) .

The existing documentation seems fine. But should this change be made to show that set( ) is the recommended way to do this? If so, can the whole section(Re-rendering) be removed?

Looks like [this](https://github.com/emberjs/ember.js/issues/13946) could also benefit from this


locks:
@akashdsouza You are right, the whole Re-rendering section can be removed! Can you do that change so this gets merged?

PR #1747, opened 2016-12-13
Fix the link to Ember.TextField.

MonsieurDart:
A subsequent use of "[1]" was conflicting with the first occurence of "[1]".


locks:
@MonsieurDart thanks for spotting this! We have been moving to inlining links to avoid exactly this problem. Could you inline this one before I merge?

source/localizable/templates/input-helpers.md
```diff
@@ -91,9 +91,9 @@ Which can be bound or set as described in the previous section.
 
 Will bind the value of the text area to `name` on the current context.
 
-[`{{textarea}}`][1] supports binding and/or setting the following properties:
+[`{{textarea}}`][4] supports binding and/or setting the following properties:
```
locks: Inline it, s'il vous plaît.

MonsieurDart:
Yeah! Done. 😊

MonsieurDart:
Oups…

locks:
Thanks!

PR #1748, opened 2016-12-15
Update simple-component.md

DaleHe:
I guess there is something wrong with the order between the two words


toddjordan:
Thanks @DaleHe,  OOYL ;-)

image

PR #1768, opened 2017-01-12
Code error in the example for rentals.hbs template

travismchan:
The {{#each model as |rental|}} under {{rental-listing rental=rentalUnit}} should be deleted with the red overlay. I'm not certain how to do this on the proposed changes since this is my first proposed file change. Also the closing {{/each}} tag should be green as well but has a red overlay. This is for the 2.7.0 documentation.


toddjordan:
Thanks for the submission @travismchan !

This is what I see as is currently deployed for the 2.10 documentation (which is guides master):
image

I do see the issue you describe with the 2.7 documentation, but the fix you are providing will not address this, since  master points at 2.10.  To fix, we can look at back-porting the current simple-component tutorial page back to 2.7.  I can bring this up at the next learn meeting.

PR #1776, opened 2017-01-17
Improve acceptance test diff in tutorial service page

kennyadsl:
Reading the tutorial it wasn't immediatly clear where to add code. Adding diff lines could improve comprehension.

I can change other languages service pages if needed.

This is the result:

acceptance-test-diff


toddjordan:
I'm ok with this.  Thanks @kennyadsl 
@homu r+

PR #1777, opened 2017-01-18
Adds a note on how to think about mobile perf

tomdale:
One common misconception people often have is that JavaScript frameworks will give them good performance for free. While they can certainly help, it's all too easy for inefficient code in the app to significantly slow down performance on mobile devices.

This PR adds a note advising readers that, when it comes to mobile performance, there's no such thing as a free lunch. Many performance problems are architectural ones that need to be caught early or else they become very painful to unwind.

### DIFF
--- a/source/localizable/index.md
+++ b/source/localizable/index.md
@@ -36,6 +36,33 @@ strings, destructuring, and more. You can check the
 [Babel.js documentation][babeljs] or read [Understanding ECMAScript 6][es6]
 online.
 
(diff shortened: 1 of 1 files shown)


bmeurer:
Very nice! Thanks for helping with raising awareness of mobile performance issues early on.

addyosmani:
Just saw this. Thanks for committing to including messaging about testing on real devices, @tomdale 👍

PR #1778, opened 2017-01-18
Fixing CD for guides

acorncom:
Right now, due to version sorting, we're deploying all latest changes to v2.9.0 guides instead of to the v2.10.0 guides as intended


locks:
@homu r+

PR #1394, opened 2016-04-23
Update quick-start.md

Barnoid-:
- version bump to 2.5
- fix broken link to ember-cli-deploy
- add info on how to deploy to an Apache server
  (there are lots of hits in Google about this and many misleading/nonworking suggestions. The simplest working one seems to be FallbackResource)

PR #1779, opened 2017-01-18
Update deploying.md to help Apache users

sinrise:
Most websites are hosted on Apache servers so many users will need to enable and configure mod-rewrite in order for Ember routing (or most MVC routing) to work properly. This gives the most common example.

### DIFF
--- a/data/spelling-exceptions.txt
+++ b/data/spelling-exceptions.txt
@@ -8,6 +8,7 @@ ajax
 allget
 amd
 americentric
+apache
 api
 apis
 app's
@@ -109,9 +110,11 @@ hashchange
 hasmany
 hoc
 homebrew
+htaccess
 html
 htmlbars
 http
+httpd
 https
 iframe
 iframes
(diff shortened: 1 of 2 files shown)


acorncom:
@sinrise We're generally trying to be web server agnostic.  Do the changes at https://github.com/emberjs/guides/pull/1394 not suffice?

sinrise:
No, the short note at the end of that doc suggests creating a virtual host,
which isn't going to work for most people in a shared hosting environment
when they don't have access to the actual server config. I wasn't going to
make a pull req but locks asked me to on the forum
http://discuss.emberjs.com/t/getting-started-dist-404/12284/3

acorncom:
@sinrise Sounds good. Was discussing it further with @locks and we think we'd like to shift to having a static servers header (with the size heading you have right now for Apache) and then your Apache info as a sub-heading of that.  It'll allow us to add in an Nginx config as well when we get asked for that as well 😀 Would you mind making that change and fixing the build as mentioned on Travis?  We should then be able to merge this

sinrise:
No problem at all! I'll do it right now. That makes more sense to me, too.

acorncom:
@homu r+

Thanks @sinrise !

Issue #1781, opened 2017-01-20
Components Wrapping content refers to helpers without linking


theroncross:
The 'Components - Wrapping Content in a Component' is a bit of a tough one to get through for beginners. One small part of that is that it refers to the component and hash helpers, but doesn't include links to the API docs. Although this is probably a bigger issue generally, I figured I could tackle this page quickly unless there's a reason to keep linking to a minimum. I'd be happy to work on more pages if that seems useful.

toddjordan:
hi @theroncross, thank you! I know that @sclatter was doing something along similar lines.  Her initial pr was to add a section on layouts https://github.com/emberjs/guides/pull/1763, and the latest comment is to encorporate that introduction into this page.

That said, the hash helper part certainly use love and is outside the scope of that pr, so if you have some love to give, it would be appreciated ❤️

theroncross:
@toddjordan do you have page in mind for this?

toddjordan:
I actually was referring to the wrapping content page you mention here.  (didn't notice you already had a pr).  What I was trying to say is that there was work happening on the first part of the page round layouts/block components, but you were free to improve the more advanced contexual component piece.

theroncross:
By the time you stare at these docs long enough to change them, you lose sight of what you found difficult, confusing, or lacking in the first place.

toddjordan:
lol, truth

toddjordan:
Closing this issue out as it appears that we've merged content to address it.  lmk if there's more...

PR #1782, opened 2017-01-20
Adds Links to template helper API docs

theroncross:
towards #1781

Adds links to the component and hash helper docs


locks:
Thanks!

@homu r+

PR #1788, opened 2017-01-23
Typo Error

babbarankit:


locks:
Hi @babbarankit, thank you for your contribution, but this isn't actually a typo. `ember server` is actually an alias to `ember serve`, so you can use either one, or even `ember s`!
We are thinking of standardizing to `ember serve` because that's the primary command, and what appears in `ember help`. Are you interested in updating the rest of the guides?

PR #1792, opened 2017-01-26
Tweaking info related to Bower.json

acorncom:
Adjusting our instructions related to bower, as we no longer need it as much


acorncom:
Closing in favor of #1805

PR #1796, opened 2017-01-28
Add quotation marks around Google Maps url

bzf:
When trying to run the command from the document in `zsh`, I got the following
error output:

```shell
$ curl -o vendor/gmaps.js https://maps.googleapis.com/maps/api/js?v=3.22
zsh: no matches found: https://maps.googleapis.com/maps/api/js?v=3.22
```

By adding quotation marks around the URL the command works as expected in both
`zsh` and `bash`.


locks:
@bzf appreciated!

@homu r+

PR #1797, opened 2017-01-30
Polymorphism docs

briangonzalez:


source/localizable/models/relationships.md
```diff
@@ -138,6 +138,113 @@ export default DS.Model.extend({
...
+
+Note that, for polymorphism to work, Ember Data expects a
+"type" declaration polymorphic type via the reserved `type`
+property on the model. Confused? See the API response below.
+
+First, let's look at the model definitions:
+
+```app/models/user.js
+import DS from 'ember-data';
+let { hasMany } = DS;
```
locks: use `DS` and `Ember` namespaces like other samples

briangonzalez: @locks Fixed.

locks:
Thanks!

@homu r+

PR #1800, opened 2017-02-02
Update algolia version

Tonours:
Related to https://github.com/emberjs/website/pull/2799

- Updated algolia by modifiying CDN namespace version in order to display an error message when no results found.
- Updated _docsearch.scss vendors accordingly to algolia CDN update due to some markup update from Algolia.
- Remove display none of .algolia-docsearch-suggestion--text when parent node has a class of --no-results otherwise "no results" message was not visible.

Now when there is no result we have a UI feedback

result

### DIFF
--- a/source/javascripts/app/docsearch.js.erb
+++ b/source/javascripts/app/docsearch.js.erb
@@ -2,7 +2,7 @@
   var script = document.createElement('script');
   script.type = 'text/javascript';
   script.async = true;
-  script.src = '//cdn.jsdelivr.net/docsearch.js/1/docsearch.min.js';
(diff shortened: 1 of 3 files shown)


locks:
👏  Thanks for this!

acorncom:
@homu r+  Thanks @Tonours !

acorncom:
@Tonours by the way, not sure if you have time or interest, but we're always looking for folks who want to pitch in with us to tackle various infrastructure / design projects for Ember. If you're on the Ember community Slack, look us up on `#-team-learning` if you'd like to discuss bigger projects 😀

Tonours:
Hi @acorncom ! Yes of course, see you at Slack chan :)

PR #1803, opened 2017-02-03
corrects instructions for deploying to surge.

sintaxi:
The previous instructions would not deploy to the same domain as expected as surge takes two arguments (eg. `surge <project> <domain>`)

before:
```
surge funny-name.surge.sh
```

after:
```
surge dist funny-name.surge.sh
```

this would also work:
```
surge --project dist --domain funny-name.surge.sh
```

### DIFF
--- a/data/spelling-exceptions.txt
+++ b/data/spelling-exceptions.txt
@@ -71,6 +71,7 @@ deserializing
 destructuring
 django
 dom
+DNS
 draggable
 ds
 durations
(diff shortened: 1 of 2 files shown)


sintaxi:
Looks like spellcheck failed on `subdomain` and `DNS`. let me know if you would like that changed.

acorncom:
@homu r+  Thanks @sintaxi

PR #1805, opened 2017-02-03
correct obsolete references to bower.json

jaredcnance:


acorncom:
@homu r+  Thanks @jaredcnance that's quite helpful

PR #1807, opened 2017-02-07
Serve vs server

NathanTheGr8:
I was following the guide and thought I found a typo, turns out server, serve, and s are all aliases of each other. Explaining that in the guide would be useful.


locks:
Thanks!

PR #1808, opened 2017-02-09
Update installing-addons.md

rayattack:
The rental type output as depicted in the image resource @ ../../images/installing-addons/styled-super-rentals-basic.png contains a compound string that differs from the rental type property returned in the model hook.

PS: Kindly modify the text 'template helpers' to 'Creating a Handlebars Helper'


toddjordan:
@rayattack Thanks for contributing!  I'd rather we update the image to just show type, rather than add this big block of text/code just to justify it.

rayattack:
@toddjordan That would work as well, but then the hyphenated compound rental types are used in later on in the Creating a Handlebars helper section.

rayattack:
@toddjordan I finally see what you mean by correcting the image.

PR #1810, opened 2017-02-10
Update specifying-a-routes-model.md

Leooo:
@samselikoff I think your article is so good on an important subject that it needs a proper reference in Ember doc.

### DIFF
--- a/source/localizable/routing/specifying-a-routes-model.md
+++ b/source/localizable/routing/specifying-a-routes-model.md
@@ -163,3 +163,5 @@ each record in the song model and album model:
   {{/each}}
 </ul>
 ```
+
+If you are using Ember-Data, it is in general [bad practice](https://medium.com/@sam.selikoff/because-returning-a-single-domain-object-from-the-route-is-the-ember-pattern-the-very-name-of-the-94c4abf4ad58) to return multiple models in a route: in the above, prefer returning the `album` model in which you have setup the `songs` [hasMany](https://guides.emberjs.com/v2.1.0/models/working-with-relationships/) relationship, and access your songs in your templates using `album.songs` (or `album.get('songs')` in your components). 


locks:
Thanks for your submitting @Leooo, but I am against incorporating this change for a couple of reasons. The link is to a comment on a blog post, not the blog post itself. The relevant information should be incorporated into the document instead of linked out. Finally, this opinion regarding the `model()` hook does not necessarily align with the opinion of the learning team and the framework maintainers.

Leooo:
thanks @locks.
* I could incorporate the information in the PR, but I was worried to add too many lines to the guides while users could alternatively refer to the article if needed.
* Link to comments instead of article I'n not sure, looking at the url it doesn't seem to link to a comment?
* More importantly, just wondering what you mean by "it does not necessarily align with the opinion of the learning team and the framework maintainers": is it a polite way to say that Ember-Data core team disagree with this "bad practice" comment (in which case shouldn't be merged), or that they wont necessarily agree with all such comments, in which case we should take opinions from some core Ember-Data  contributors, as I think this is a very relevant and important point to precise.

L

samselikoff:
locks and I chatted in slack, because the topic is nuanced I think it deserves further discussion in the guides, most likely under the Models topic. Then we could add a sentence here that links to it:

> It's worth noting that multiple models that are related can often sideloaded, so all the data is loaded using a single HTTP request. Read [this discussion] for more details.

samselikoff:
Also fwiw people who have disagreed with my comment above on this, have usually mentioned edge cases. My comment was about the 95%, and about the mistakes I see most often within teams I've worked with (and also my own code!).

Leooo:
So, I think we have two related but different problems here:

1) because we don't have a perfect API, people in my company tend to overuse hashes in route models instead of properly setuping relationships, which can be done by tweaking the relationship links in the serializers: so if you use Ember-Data, and want to work on a solid domain model, taking the very good example of @SaladFork in Slack, instead of:

```
//route.js
 model() {
    return Ember.RSVP.hash({
      user: store.findRecord('user', userId),
      countries: store.findAll('country')
    });
  }
```

prefer using:
```
//route.js
import Ember from 'ember';
 model() {
    return store.findRecord('user', userId)
  }
});
```

```
//models/user.js
import DS from 'ember-data';
  countries: DS.hasMany('country')
});
```

```
//serializers/user.js
import DS from 'ember-data';
export default DS.RESTSerializer..extend({
  normalize(modelClass, resourceHash) {
    let links = resourceHash.links || {};
    links.countries= `/api/countries`;
    // or use links.countries= `/api/countries?userId=${resourceHash.id}`; depending of the API
    resourceHash.links = links;
    return this._super(modelClass, resourceHash);
  }
});
```

I think this example is good because I've seen it dozens of times in cases where the API urls were not relationship-compliants.


2) With the setup above, we would still have two requests, so it would be better to sideload the `countries` models when loading the `user` if possible, but I think it is a slightly different subject.


Not sure where is the best place to document that but I think giving this example would help a lot, for example:
* add a "Using relationships when your API doesn't" paragraph at the end of https://guides.emberjs.com/v2.6.0/models/relationships/ 
* add a line in the "multiple models" section of https://guides.emberjs.com/v2.6.0/routing/specifying-a-routes-model/ like: "it is often ([sometimes], [in some cases]) better to properly setup relationships between models in Ember-Data models / serializers, return only the core model for the route in the hook, and access its related models in the templates / components, so as to preserve a solid domain model in your app and DRY. For more information see [link]". (Sentence probably has to be changed but you see my point). @samselikoff @locks thoughts?

samselikoff:
If @locks is on board I can take a stab at updating the Multiple Models section, and maybe we can leave some of the more prescriptive language out. But something that nudges people in the right direction. Something like

"Oftentimes a route will need data from multiple models to render its template. For example, a blog post page might need to render the blog post itself as well as the post's related comments. 

If there is a clear relationship between the models being loaded (the `post` and `comments` in the example above), the standard way to fetch them all is by side-loading the secondary data along with the primary model.

APIs handle side-loading in a variety of ways, for example through query parameters sent along with the request:

```js
model() {
  return this.store.findRecord('post', 1, { include: 'comments' });
}
```

You can read more about side-loading in the [models guide]().

If a route needs data from multiple models but those models are not related, you can use Ember.RSVP.hash to make multiple requests... (rest of RSVP.hash stuff).

Leooo:
@samselikoff I still think that you can return one model in the route even without sideloading the models (which needs the API to be compatible): return `this.store.findRecord('post', 1)` and use `post.comments` to launch a new request to backend (so lazyload your comments in your templates / components when needed). Sideloading is one step bove that.

samselikoff:
Ok so maybe we alter it to something like

If there is a clear relationship between the models being loaded (the post and comments in the example above), the standard way to fetch them all is via Ember Data relationships. You can read more about it [in the models guide](), but for example you might lazily fetch related resources in your templates

{{post.comments}}

or eagerly through side-loading

{ include: 'comments' }

Read more here blah blah.

Leooo:
@samselikoff sounds great, let me know if I can help.

SaladFork:
@samselikoff thanks for the blog post! I had actually wanted to set up these sorts of relationships on my models, but the issue I was having is having no control over a backend that does not sideload or send relationship information. That is, a `post` model did not include the IDs for its `comments` (though comments do know what post they belong to). Against the other example in this thread, a `user` response did not include the IDs of all the `countries`. This resulted in ED never making a request for the relationship even if you asked for it, as it thought it was already loaded and fulfilled (as an empty array). I felt this forced me into using an `RSVP.hash` in my model hook.

@Leooo helpfully introduced me to the `links` property on Slack and explained how to take advantage of it (even though I'm using a `RESTSerializer`). It seems to do exactly what I wanted and accomplishes the goal laid out by the blog post.

Documentation for something like "how to use relationships in your apps when your API doesn't send IDs for all the related models" would have been immensely helpful. As far as I can tell, there is no documentation for `links` at all on `RESTSerializer` or any of the model guides.

Leooo:
replaced by https://github.com/emberjs/guides/pull/1810 @locks ?

PR #1811, opened 2017-02-10
[DOC release] Replace var with let

Turbo87:
/cc @locks


locks:
You don't need to at me ;P

@homu r+

Turbo87:
> You don't need to at me ;P

🤷‍♂️ 

Just making sure you're seeing this. I don't "watch" half the repos I maintain... 😜

PR #1812, opened 2017-02-10
Typo in the routes and templates md 

brimatteng:
We want to change the application.hbs template not the application.js file to include the markup for the top nav bar.


locks:
Thanks!

@homu r+

PR #1813, opened 2017-02-13
Changed double quotes to single quotes around description property value

benjaminokeefe:
Changed the double quotes surrounding the description property's text to single quotes for consistency with the rest of the string data.


locks:
Appreciated!
@homu r+

Issue #1718, opened 2016-11-03
Tutorial TDD


toddjordan:
Feedback from new Developers has been that walking through the tutorial using TDD-style approach is a hindrance to learning the brand new technology, because we are essentially starting the reader off with something that by definition doesn't work.  Also, we've heard that testing first requires some knowledge of the system under test or the underlying technology, which a new beginner will not yet possess.  Finally, a lot of new concepts are being introduced at once, and including one more, especially if the user has no experience in testing and TDD, adds to feelings of being overwhelmed.

On the flip side, others have mentioned that there is value in showing a workflow around how one might create an Ember application, and highlight how TDD might work in that context.  This is likely more tailored to developers with some SPA experience and some previous exposure to automated testing.  Earlier feedback suggested that providing examples of TDD workflow in Ember is helpful to developers.

Some Options (so far): 

- Introduce Acceptance testing at the end of the tutorial, after the application is already working.  Same for unit/integration: introduce at the end of the section that they are introduced in. (ex. component integration tests at the end of the components section.
- Have 2 tutorials: One that is tailored to newer developers and that very slowly and intentionally layers basic concepts to get to a basic knowledge of using Ember.  The other would assume more initial knowledge of SPA and development concepts focusing rather on a typical workflow one might use in developing an ember application.
- Keep TDD, but be a lot more deliberate and careful in introducing and explaining it.  Give the user the option (with links) to skip if they want to get directly into app code.

mike-north:
It seems like there are two issues that relate to this discussion

* Should the primary path for teaching those new to ember (in the tutorial) be via TDD?
* Should we provide some material about using TDD workflows with ember?

I don't think anyone disagrees with the idea that TDD should have a place in the guides, but I'm not sure that these are two sides of the same issue.

theroncross:
As a former teacher, I can confirm that attention is VERY limited for beginners and repetition is a major key to learning. My first instinct would be to present every topic twice in the tutorial:
1. First, make a thing,
2. Second (optionally), write tests for the thing,
3. Third, TDD another similar thing.

toddjordan:
Thanks @theroncross , that's really good feedback.  I really like the approach you mention and will take that into consideration when we address this.

mike-north:
I'm curious -- who is the developer (our audience) whose learning of core ember concepts for the first time, is enhanced (rather than muddled) simultaneously learning about testing controllers and ember's testing tools, while learning about controllers for the first time?

[...]

Issue #1708, opened 2016-10-21
Perspective of someone new to SPAs: Large tutorial in guides is incredibly confusing


lisaychuang:
**My background**
I have some basic programming background but have never used a SPA framework before. The idea of using something like Ember is appealing because I don't have the experience to make all of the stack/architecture decisions myself.

**Going through the guides**

The number of new concepts introduced in very early steps (i.e., as early as steps 1 and 2) were totally overwhelming. After speaking with people who know ember, I understand now that I was asked to grasp the concepts of
- a new build tool
- a web framework
- a "data persistence" library
- a testing framework
- a test assertion library
- the concept of a single page application
- "next-generation javascript", including a new "import" syntax I was previously not familiar with
- JSHint

Even later, there are concepts that were introduced with no explanation. Again, after asking people who have more experience, I now understand that these are:
- Promises
- const vs. let
- the idea of using components `{{like-this}}` or `{{#like-this}} {{/like-this}}`

And through the tutorial I also was left with a lack of understanding on certain things that seemed important to me. I expect that I could have looked at other documents to figure this stuff out, but it seems like the guides would benefit from including links / references to explanation right there.
- params vs. hash for handlebars helpers
- QUnit
- the router vs. routes

Finally, there doesn't seem to be a well thought out order/ organization of how the topics are introduced. Here's my detailed feedback on a per-page basis, so the experts who maintain this material can understand a beginner's perspective:
- Setting up the tests
  - The concept of acceptance tests "verifying that a route is loaded" is mentioned, but I had no idea what a "route" is at this point!
  - The "andThen" helper is introduced before "asynchronous test helpers" are mentioned. I was quite confused as to why I would need to "wait" for anything to complete
  - Although a screenshot was added within the last week, there were references to "the data" and "search field" thrown around, for which I had no context at that point.
  - By and large, because this was so overwhelming to grasp, I just copied/pasted and wasn't able to absorb or understand much.
- Routes and Templates
  - This sentence is supposed to help me understand what's going on, but cannot be understood without any of the concepts being introduced yet `Let's start by building our "about" page. Remember, when the URL path /about is loaded, the router will map the URL to the route handler of the same name, about.js. The route handler then loads a template`. I still have no idea what a template is, and have to infer what a route and router might be...
  - "route" and "route handler" seem to be used interchangeably. Are they the same? 
  - Use of ember-specific terminology is mysterious for newcomers (i.e., "the generator has **mapped** a new route for us").
[...]

PR #1818, opened 2017-02-20
More gradual testing intro

toddjordan:
Addresses #1718 
Fixes #1863 

Still in need of some proof reading, and would like a couple of people to walk through this version of the tutorial, since its a pretty big change.


locks:
Poo, it's a significant conflict :\

toddjordan:
Also made several updates that address some of the initial comments in #1708

source/localizable/tutorial/acceptance-test.md
```diff
@@ -52,133 +70,63 @@ test('visiting /', function(assert) {
...
 
-test('should redirect to rentals route', function (assert) {
-});
+By default, when you run `ember test --server`, Ember CLI runs the [Testem test runner](https://github.com/testem/testem),
+which in turn runs the qunit test framework on Chrome and [PhantomJS](http://phantomjs.org/) out of the box.
 
-test('should list available rentals.', function (assert) {
-});
+The launched Chrome web browser shows 10 successful tests.
+Ember generates tests that lint each file you create using [JSHint](http://jshint.com/).
```
theroncross: Should this be changed to eslint now?

source/localizable/tutorial/acceptance-test.md
```diff
@@ -52,133 +70,63 @@ test('visiting /', function(assert) {
...
 ```
 
-Of course because we have not implemented this functionality yet, our tests will all fail.
-So, your test output should now show all failed tests when running `ember test --server`, which gives us a todo list for the rest of the tutorial.
-
-![failing tests](../../images/acceptance-test/failed-acceptance-tests.png)
+Running `ember test --server` will show a total of 7 failed tests out of 15.
+Each of the 6 tests defined above will fail, plus you'll notice one JSHint failure saying, `assert is defined but never used`.
+The tests we defined fail because a QUnit test fails if it is empty.
+Tests require a check for a specific condition (known as an `assert`).
```
theroncross: I think this is a great compromise on de-emphasizing testing.

source/localizable/tutorial/ember-data.md
```diff
@@ -102,5 +102,5 @@ When we call `findAll`, Ember Data will attempt to make a GET request to `/renta
 
 You can read more about Ember Data in the [Models section](../../models/).
 
-Since we're using Mirage in our development environment, Mirage will return the data we've provided.
-When we deploy our app to a production server, we will need to provide a backend for Ember Data to communicate with.
+Since we've already set up Ember Mirage in our development environment, Mirage will return the data we requested.
+When we deploy our app to a production server, we will likely want to provide a backend server for Ember Data to communicate with to store and retrieve persisted data from.
```
theroncross: ...for storing and retrieving persisted data. Anything saved only in Ember Data's local store will be lost when we close our app.

source/localizable/tutorial/installing-addons.md
```diff
@@ -8,8 +8,13 @@ For Super Rentals, we'll take advantage of two addons: [ember-cli-tutorial-style
 ### ember-cli-tutorial-style
 
 Instead of having you copy/paste in CSS to style Super Rentals, we've created an addon called [ember-cli-tutorial-style](https://github.com/ember-learn/ember-cli-tutorial-style) that instantly adds CSS to the tutorial.
-The addon works by creating a file called `ember-tutorial.css` and putting that file in the super-rentals `vendor` directory.
-As Ember CLI runs, it takes the `ember-tutorial` CSS file and puts it in `vendor.css` (which is referenced in `app/index.html`).
+The addon works by generating a file called `ember-tutorial.css` and putting that file in the super-rentals `vendor` directory.
+
+The [`vendor` directory](../../addons-and-dependencies/managing-dependencies/#toc_other-assets) in Ember is a special directory where you can include content that gets compiled into your application.
+
+As Ember CLI runs, it takes the `ember-tutorial` CSS file and puts it in a file called `vendor.css`.
```
theroncross: When Ember CLI builds our app from our source code, it copies `ember-tutorial.css` into a file called `vendor.css`.

source/localizable/tutorial/service.md
```diff
@@ -1,185 +1,146 @@
...
+* It is injected with a [service locator](https://en.wikipedia.org/wiki/Service_locator_pattern),
+  meaning it will abstract the maps API from the code that uses it,
+  allowing for easier refactoring and maintenance.
 * It is lazy-loaded, meaning it won't be initialized until it is called the first time.
-In some cases this can reduce your app's processor load and memory consumption.
-* It is a singleton, which will allow us cache map data.
-* It follows a lifecycle, meaning we have hooks to execute cleanup code when the service stops, preventing things like memory leaks and unnecessary processing.
+  In some cases this can reduce your app's processor load and memory consumption.
+* It is a singleton,
+  which will allow us to keep map data will the user navigates around the app,
```
theroncross: It is a singleton (i.e., all references to the service point to the same object, and it remains in memory until we explicitly end the service or close our app), which will allow us to keep map data __while__ the ...

source/localizable/tutorial/service.md
```diff
@@ -1,185 +1,146 @@
...
+  so that returning to a page doesn't require it to reload its maps.
+* It follows a lifecycle,
+  meaning we have hooks to execute cleanup code when the service stops,
```
theroncross: We'll use hooks to execute cleanup code when we stop the service to prevent memory leaks and unnecessary processing.

source/localizable/tutorial/service.md
```diff
@@ -1,185 +1,146 @@
...
-```
-
-Note that the test uses a dummy object as the returned map element.  This can be any object because it is only used to assert that the cache has been accessed.
-Also note that the location has been `camelized` in the cache object, so that it may be used as a key.
-
-Now implement the service as follows.  Note that we check if a map already exists for the given location and use that one, otherwise we call a Google Maps utility to create one.
-We abstract our interaction with the maps API behind an Ember utility so that we can test our service without making network requests to Google.
+Now implement the service as follows.
+Note that we check if a map already exists for the given location and use that one,
+otherwise we call a Google Maps utility to create one.
```
theroncross: Duplication of the lines above?

source/localizable/tutorial/service.md
```diff
@@ -216,97 +177,190 @@ export default Ember.Service.extend({
...
-Once in the vendor directory, the script can be built into the app.
-We just need to tell Ember CLI to import it using our build file:
+Running this command generates three files: a component JavaScript file, a template, and a test file.
 
-```ember-cli-build.js{+22}
-/*jshint node:true*/
-/* global require, module */
-let EmberApp = require('ember-cli/lib/broccoli/ember-app');
+Let's start by adding a `div` element to the component template.
+This `div` will act as a place for the 3rd part map API to render the map to.
```
theroncross: 3rd-__party__

source/localizable/tutorial/service.md
```diff
@@ -216,97 +177,190 @@ export default Ember.Service.extend({
...
+    let mapElement = this.get('maps').getMapElement(location);
+    this.$('.map-container').append(mapElement);
+  }
+});
 ```
 
-The CLI `generate util` command will create a utility file and a unit test.
-We'll delete the unit test since we don't want to test Google code.
-Our app needs a single function, `createMap`, which makes use of `google.maps.Map` to create our map element, `google.maps.Geocoder` to lookup the coordinates of our location, and `google.maps.Marker` to pin our map based on the resolved location.
+Finally open the template file for our `rental-listing` component and add the new `location-map` component.
```
theroncross: You may have noticed that `this.get('location')` refers to a property `location` we haven't defined. This property will be passed in to the component by its parent template.

source/localizable/tutorial/service.md
```diff
@@ -216,97 +177,190 @@ export default Ember.Service.extend({
...
-  init() {
-    this.set('geocoder', new google.maps.Geocoder());
-  },
+You may now either move onto the [next feature](../subroutes/), or continue here to test the maps feature we just added.
 
+###  Unit testing a Service
+
+We'll use a unit test to validate the service.
+Unit tests are more isolated than integration tests and acceptance test,
+and are intended for testing specific login within a class.
```
theroncross: ~~login~~ logic

source/localizable/tutorial/service.md
```diff
@@ -216,97 +177,190 @@ export default Ember.Service.extend({
...
+  });
+  let mapService = this.subject({ cachedMaps: stubCachedMaps });
+  let element = mapService.getMapElement('San Francisco');
+  assert.equal(element, DUMMY_ELEMENT, 'element fetched from cache');
+});
+```
+When the service calls `createMap` on our fake utility, we will run asserts to validate that it is called.
+In our first test notice that we expect four asserts to be run in line 17. Two of the asserts run in the test function, while the other two are run when `createMap` is called.
+
+In the second test, only one assert is expected (line 26), since the map element is fetch from cache and does not use the utility.
```
theroncross: fetch**ed**

source/localizable/tutorial/simple-component.md
```diff
@@ -201,7 +173,175 @@ export default Ember.Component.extend({
...
+  this.set('rentalObj', rental);
+});
+
+test('should toggle wide class on click', function(assert) {
+  this.set('rentalObj', rental);
+});
+```
+
+Now lets render our component using the `render` function.
+The `render` function allows use to pass a template string, so that we can declare the component in the same way we do in our templates.
```
theroncross: us~~e~~

source/localizable/tutorial/simple-component.md
```diff
@@ -201,7 +173,175 @@ export default Ember.Component.extend({
...
+
+Finally, lets add our actions and assertions.
+
+In the first test, we just want to verify the output of the component, so we just assert that the title and owner text match what we provided in the fake `rental`.
+
+```tests/integration/components/rental-listing-test.js
+test('should display rental details', function(assert) {
+  this.set('rentalObj', rental);
+  this.render(hbs`{{rental-listing rental=rentalObj}}`);
+  assert.equal(this.$('.listing h3').text(), 'test-title');
```
theroncross: Add messages for continuity with earlier tests?

toddjordan:
Thank you so much @theroncross for reviewing!   I'll address these this weekend.

source/localizable/tutorial/acceptance-test.md
toddjordan: As of 2.11 (which I think the tutorial is at), its still installing jshint.

source/localizable/tutorial/ember-data.md
toddjordan: The last sentence might be confusing because we are talking in this paragraph about using mirage vs a remote server.  Will have to think about how to elaborate on providing a backend.
"...we will likely want to replace Mirage with a remote server for Ember Data to communicate with for storing and retrieving persisted data.  A remote server will allow for data to be shared and updated across users."

theroncross: Makes sense.

lazaridis-com:
@toddjordan, can you please find a way to bring the new version live somewhere, so I and other can review with no effort? This "Middleman" is a hell to install on Windows (see #1852).

lazaridis-com:
@toddjordan , ok, I managed to fight through installing the "middleman" thing on windows. The new version is an estimated 3% step towards the right direction. But I really think that there must be multiple tutorials.

In any way, I think you should merge this thing, thus the work can be go on.

toddjordan:
if you want to review the tutorial without installing middleman, one option is to walk through the rendered markdown files in order: 

* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/ember-cli.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/acceptance-test.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/routes-and-templates.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/model-hook.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/installing-addons.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/simple-component.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/hbs-helper.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/ember-data.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/autocomplete-component.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/service.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/subroutes.md
* https://github.com/toddjordan/guides/blob/more-gradual-testing-intro/source/localizable/tutorial/deploying.md

acorncom:
@toddjordan Nice work on this.  Just sent in a PR to tweak, refine and simplify the language on the first three pages.  I think we should merge what we have here and we can keep editorializing on it as we go.

toddjordan:
thanks @acorncom !

PR #1822, opened 2017-02-21
Aerobatic Instructions

ivanoats:
[Aerobatic](https://www.aerobatic.com) is a deployment service for front-end apps. It is similar to Surge, which is already in the docs here. Directions are included here in this pull request.

### DIFF
--- a/data/spelling-exceptions.txt
+++ b/data/spelling-exceptions.txt
@@ -300,3 +300,4 @@ internet
 plugin
 krisselden
 DependenciesGit
+pushState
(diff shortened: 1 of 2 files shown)


ivanoats:
The build failed because `pushState is misspelled` - in a yaml block?

acorncom:
Hey @ivanoats, appreciate the PR.  We're aiming to keep the number of deploy options here to a bare minimum (as we don't want a long laundry list of competitors all on the same page 😀 ), so I'll close this PR for now. But would you mind sending in your instructions to EmberWatch instead?http://emberwatch.com/CONTRIBUTING.html

Our goal is for EmberWatch to end up being our "cookbook" area where we keep our community-maintained recipes and "how tos".  I notice there isn't a "Deploying" page yet, so your work here would be a great way to get that started.  We're likely to end up linking to that page from the guides long-term and having it be our go-to reference for all sorts of ways to deploy Ember apps.

Thanks!

PR #1824, opened 2017-02-26
Update NodeJS Windows/Mac installer URL

unchris:
When I click the link on https://guides.emberjs.com/v2.11.0/getting-started/ to get to the installer, I end up at https://nodejs.org/download/ which gives me a bare folder listing of all NodeJS downloads. 

I think it would be less confusing for a user to end up at https://nodejs.org/en/download/ (note the en/) which brings up a styled webpage with helpful indications of which installer to choose.


locks:
Thanks, they apparently changed the url structure!
@homu r+

PR #1825, opened 2017-02-27
Change var to let for consistency

baweaver:
When reading through some of the docs I'd noticed that people were switching back and forth between `let` and `var` quite a bit. This PR just does a quick replacement of `var -> let` (well, more of `s/\bvar /let /g` really)


locks:
Hi @baweaver, could you change only the source `.md`? Our translation software will take care of updating the various locales afterwards :)

baweaver:
Aha, I'd done all *.md. I'll get to that tonight then. I'll probably open a
few more as I read through things the next week or so to learn.

source/localizable/configuring-ember/embedding-applications.md
```diff
@@ -32,7 +32,7 @@ router's `location`](../specifying-url-type) to
 `none`:
 
 ```config/environment.js
-var ENV = {
```
locks: This is a node.js file, I'm not sure we want to use `let` here? cc @Turbo87

source/localizable/ember-inspector/installation.md
```diff
@@ -74,7 +74,7 @@ bookmarklet option to use the Inspector.
 
 Add the following bookmark:
 
-<a href="javascript: (function() { var s = document.createElement('script'); s.src = '//ember-extension.s3.amazonaws.com/dist_bookmarklet/load_inspector.js'; document.body.appendChild(s); }());">Bookmark Me</a>
+<a href="javascript: (function() { let s = document.createElement('script'); s.src = '//ember-extension.s3.amazonaws.com/dist_bookmarklet/load_inspector.js'; document.body.appendChild(s); }());">Bookmark Me</a>
```
locks: This is bookmarklet code, no `var` here for better browser compatibility.

source/localizable/configuring-ember/embedding-applications.md
rwjblue: Should be fine, now that we only support node@4 and higher...

PR #1829, opened 2017-03-02
[routing] Reusing Route Context

knownasilya:
The section title needs bikeshedding.

Feedback appreciated.


source/localizable/routing/specifying-a-routes-model.md
```diff
@@ -163,3 +163,66 @@ each record in the song model and album model:
   {{/each}}
 </ul>
 ```
+
+## Reusing Route Context
+
+Sometimes you need to fetch a model, but your route doesn't have the params, because it's
```
locks: "params" -> "parameters"

source/localizable/routing/specifying-a-routes-model.md
```diff
@@ -163,3 +163,66 @@ each record in the song model and album model:
...
+});
+```
+
+This is guaranteed to work because the parent route is loaded. But if you tried to
+do `paramsFor` on a sibling route, you wouldn't have the results you expected.
+
+This is a great way to use the parent context to load something that you want.
+Using `paramsFor` will also give you the query params defined on that route's controller.
+
+_Note: This method could also be used to look up the current route's params from an action
```
locks: This isn't a note, it's a continuation of the previous sentence.

locks:
I'm not a big fan of introducing `modelFor` and `paramsFor` like this 🤔

knownasilya:
Willing to revamp, have any specific thoughts on how?

locks:
Gonna bring it up in the meeting today, let's see.

knownasilya:
Notes from Learning Meeting:

- [ ] Relabel this PR to make it related more explicitly to Nested Routes
- [ ] Open PR to move Nested routes to it's own section
- [ ] Link to that issue
- [ ] Revamp the nested routes guide

PR #1840, opened 2017-03-08
Make the observer disclaimer more visible.

martndemus:
Is it possible to put a hamster with a warning sign on the page?


acorncom:
@homu r+  Thanks @martndemus

PR #1845, opened 2017-03-13
[docs] Document metadata reuse

knownasilya:
It's not so obvious that the above is needed, since the model in `setupController` is different then the "model" returned from the promise via `then`.


knownasilya:
Ping

toddjordan:
Thanks @knownasilya, I'm fine with the update.  @locks @acorncom  what do you think?

acorncom:
Yep, looks good to me.  Thanks @knownasilya 

@homu r+

PR #1863, opened 2017-03-21
Fixed grammatically incorrect sentence in ember-data.md

pizza:
In v2.11.0 and earlier, the sentence I changed used to be `When we open the model file, we see:`, which was correct.
It later was changed to the current `When we open the model file reveals a blank class extending DS.Model:`, which doesn't sound right.


toddjordan:
Thank you @pizza ! Yeah, good catch.  

We have a pretty big pr coming in today #1818 

could you check if the same grammar issue is there?  If so I'm inclined to wait until it lands to merge.

pizza:
Thanks! I just checked #1818 (a pretty big PR indeed) and the grammar issue in my PR isn't there.

pizza:
@toddjordan I noticed that this PR was closed in #1818, but this typo is still there, in the last version of `ember-data.md` on master.

locks:
Done!

pizza:
Thanks @locks

Issue #1865, opened 2017-03-21
Tutorial may want to mention the Ember CLI move away from bower


RyanNerd:
As part of best practice for features that are slated or even in part have been deprecated such as `bower` it is helpful to mention this in the tutorial. For example at: https://github.com/emberjs/guides/blob/master/source/localizable/tutorial/ember-cli.md 

```md
**bower.json**: Bower is a dependency management tool.
It can be used to manage front-end plugins and component dependencies
(HTML, CSS, JavaScript, etc).  All Bower components are installed in the
`bower_components` directory. If we add front-end dependencies, such as 
Bootstrap, we will see them listed here, and added to the 
`bower_components` directory.
_Note_: Ember CLI is moving away from bower usage. We mention this now so you
can follow best practices, and to make you aware that this feature may be 
deprecated in the near future. 
```

locks:
Hi @RyanNerd, thank you for your contribution. I appreciate the sentiment of wanting to move off of Bower, however, this characterisation has some potential problems.

We've found that saying something might get deprecated in the future is detrimental (hello controllers :X). Plans change, and we only deprecate when there is a migration strategy delineated, after an RFC process. As of today, using [npm libraries in Ember CLI](https://simplabs.com/blog/2017/02/13/npm-libs-in-ember-cli.html
) still isn't as straightforward and polished and we want it to be.
You can consult our [FAQ on future-proofing](http://emberjs.com/learn/#faq-future-proof) for a more thorough description.

As far as I understand it, there is no plan to remove Bower support in the near future.
What was done in Ember CLI is make Bower no longer a dependency of the project in order to reduce the initial project size, as well as make it "pay go", meaning it is lazily required (pay) only when you actually need it (go). This includes not generating a `bower.json` for a new project in the future, as there are currently no Bower dependencies for a new project.

What was done in Ember itself was to make it into an addon named `ember-source`, like `ember-data`, that is pulled from npm like other addons.
So in that sense, we are consolidating under npm to try to make it consistent across the project dependencies, hence why `bower.json` is now empty.

I'm sorry if the answer seems a bit long, I wanted to try to explain the nuances clearly so I can reference back to post, as many people will probably make similar enquiries.

RyanNerd:
[...]

PR #1866, opened 2017-03-21
Tutorial note that Ember CLI is moving away from bower - Addresses issue #1865

RyanNerd:
Closes Issue #1865


locks:
Hi, I wrote down some feedback at https://github.com/emberjs/guides/issues/1865#issuecomment-288375833. I'll keep it focused on the change here.

I believe it's planned that Ember CLI v2.13.0 will not generate a `bower.json` file anymore, so the change here is to remove references to it, instead of adding an additional note.

Despite the file not being generated, it will still be supported, at least in the Ember CLI v2.x cycle of releases, so we'll probably need a disclaimer to that effect somewhere?

RyanNerd:
@locks Thanks for your comment. Similar issues exist for any language or framework that is evolving. As an example PHP deprecated a number of features (that will be removed in the next major version) and there's no decent tutorial on not using these and what the alternative best practice is. PHP just throws a notification (which are similar to warnings during a compile -- devs almost always ignore the warnings which leads to **not** following best practices).

I'm going to close this PR without a merge. I will create update the issue to: "have the bower section removed or altered in the tutorial/elsewhere(?) **when** 2.13 lands."

PR #1875, opened 2017-03-22
"instances of that class" instead of "instances of that object"

pizza:
I think it's more correct to say "instances of that class" instead of "instances of that object", given that we're defining classes here, and that the Ember object system is designed to mimic a more classical inheritance system.


locks:
👍 
@homu r+

locks:
@homu +

locks:
@homu r+

locks:
@homu retry

PR #1880, opened 2017-03-24
404 links

phocks:
This produces a temporary fix until the next version comes out. Probably need to create a redirect, that redirects it to the latest version. I'm not sure how to do that. Can someone help?


locks:
@phocks thanks for spotting this, my bad! I think what we want here is to link to the current version, since this is the index for a certain version of the guides. You can use the `./` here and it should work, since it's relative to the current path.

locks:
👌

PR #1884, opened 2017-03-26
Fix the path to `/templates/actions/`

snoozer05:
I think it is wrong that the link to the page on the guide is an absolute path. so I fix the link path.


locks:
Appreciated :) @homu r+

PR #1885, opened 2017-03-27
Correct required Node.js version

clekstro:
The guides currently mention Node.js 0.12, which is EOL. Instead, we should push users toward the most recent LTS version of Node.js (hat tip @rwjblue).


locks:
Thank you for the help @clekstro!
@homu r+

Issue #1869, opened 2017-03-22
Add a little more explanation about parameters "Creating a Handlebars Helper" section


toddjordan:
In ["Creating a Handlebars Helper"](https://guides.emberjs.com/v2.12.0/tutorial/hbs-helper/), we explain that we are using destructuring and link to it which is great, but it could use a bit more explanation on where the array of parameters come from.

2nd to last paragraph could go something like this (feel free to make this text clearer/simpler): 
```
Handlebars passes an array of arguments in order from our template to our helper. 
For example, if you create a handlebars helper called `{{my-helper}}`, and declare it in your template as `{{my-helper "foo" "bar"}}`, then when the helper function gets called, it receives an array with two items, "foo" and "bar", in that order.
In the example above, we are using ES2015 destructuring to get the first item in the array and name it to a variable called `type`. 
Then we can check to see if type exists in our communityPropertyTypes array.
```

MarcoUmpierrez:
What about this?

> Handlebars allows to have an indefinite number of arguments in the form of an array. This means that on each execution of the helper function, the number of arguments can change depending on how it's called on the template. For example, if you create a handlebars helper called `{{my-helper}}`, and declare it in your template as `{{my-helper "foo" "bar"}}`, then when the helper function gets called, it receives an array with two items, "foo" and "bar", in that order.
> 
> To simplify the access of these arguments, we make use of the ES2015 destructuring feature, which allows us to give a name to each expected argument in the array - in this case `type`- directly on the helper function header. Then we can check to see if type exists in our communityPropertyTypes array.

theroncross:
That's definitely an improvement. I would make the following changes.

Each argument in the helper will be added to an array and passed to our helper. For example, `{{my-helper "foo" "bar"}}` would result in `myHelper(["foo", "bar"])`. Using [array destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), we can name expected parameters within the array. In the example above, the first argument in the template will be assigned to `type`. This provides a flexible, expressive interface for your helpers, including optional arguments and default values.

mixonic:
Fixed by https://github.com/emberjs/guides/pull/1886

PR #1886, opened 2017-03-27
Update the handlebars helper tutorial with more info on the arguments

mattmarcum:
This addresses [1869](https://github.com/emberjs/guides/issues/1869)


locks:
Hello, this PR and #1887 seem to have some overlap. @mattmarcum @jonpitch do you mind coordinating into a single PR to make it easier to review?

mattmarcum:
@locks I talked to jon and we both decided to leave it up to you decide which one you like better.  We're doing these at the os contrib workshop.

mmun:
I like this! Thank you very much!

Issue #1868, opened 2017-03-22
Tutorial "building a complex component" section improvements


toddjordan:
In the tutorial section called "Building a Complex Component" (complex-component.md), we should add a small blurb at the beginning describing why the component we are about to build is considered "complex".  I'd say the new component is complex because it takes in input, invokes a function that performs the search (closure action), and then allows the consumer of the component to customize the results of the search (via a block component).

Something along the lines of...

```
As they search for a rental, users might also want to narrow their search to a specific city. Let's build a component that will let them filter rentals by city.
While our initial [rental listing component](../simple-component/) simply displayed rental information to the screen, 
this new filter component will additionally allow the user to provide input in the form of filter criteria and allow for customization of how results are shown.
```
---

In the 4th paragraph, we introduce the concept of a "block component" without saying it.  We should mention that what we are showing is referred to as "block component", and link to the guides explanation of it (../../components/wrapping-content-in-a-component/).

---

Finally, After the list-filter.hbs code snippet, while explaining the input helper, we make a couple of references to bound properties.  We need to explain what data binding is here.  Something like

```
The value property of the input will be [bound](../../object-model/bindings/) to the value property in our component, meaning that when ever the value of the property changes, either by the user typing in the input field, or by assigning a new value to it in our program, the new value of the property is present in both the rendered web page and in the code. 
```

mixonic:
Closed by https://github.com/emberjs/guides/pull/1888

PR #1888, opened 2017-03-27
Improves the "building a complex component" tutorial

yogipatel:
Addresses #1868 

This adds an expanation for why this component is complex (vs simple), and
briefly introduces the concepts of bindings and block components.

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -59,7 +59,7 @@ We want the component to simply provide an input field and yield the results lis
 The template contains an [`{{input}}`](../../templates/input-helpers) helper that renders as a text field, in which the user can type a pattern to filter the list of cities used in a search.
 The `value` property of the `input` will be kept in sync with the `value` property in the component. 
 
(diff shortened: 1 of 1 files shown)


yogipatel:
@locks sorry, I forgot about the char limit.

source/localizable/tutorial/autocomplete-component.md
```diff
@@ -54,7 +57,12 @@ We want the component to simply provide an input field and yield the results lis
 ```
 
 The template contains an [`{{input}}`](../../templates/input-helpers) helper that renders as a text field, in which the user can type a pattern to filter the list of cities used in a search.
-The `value` property of the `input` will be bound to the `value` property in our component.
+The `value` property of the `input` will be kept in sync with the `value` property in the component. 
+
+Another way to say this is that the `value` propery of `input` is [**bound**](../../object-model/bindings/) to the `value` property of the component.
```
locks: `propery` is a typo, can you fix it? I think we're ready to merge then

locks:
Now we're ready to roll 😁  Thanks for the help!
@homu r+

PR #1895, opened 2017-03-28
Add link/callout for 2017 survey results

tzellman:
Screenshot


source/_survey-reminder.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="survey-reminder-wrapper">
+  <div class="survey-reminder-container">
+    The results of the
```
locks: Can we add an action here? "See the results of the" for example?

tzellman:
Updated, thanks.

PR #1898, opened 2017-03-30
Updates tutorial

Orbyt:
Updates tutorial to remove references to `application.hbs`, which is no longer generated by `ember new`.


Orbyt:
The related functionality described in the referenced issue was revert in https://github.com/ember-cli/ember-cli/pull/6378

Will close.

PR #1900, opened 2017-04-05
Sample code - closing tag compilation error

kumkanillam:
1. In `index.hbs ` tag closing compilation error fixed
2. `hasBlock`  example code should call only one time with all the required parameters


locks:
Appreciated!

PR #1905, opened 2017-04-10
Replace `className` with `classNames`

gandalfar:


acorncom:
@homu r+ Thanks @gandalfar !

locks:
@homu r+

locks:
@homu retry

PR #1909, opened 2017-04-17
Possible missunderstanding on Installing Addons t…

pgbonino:
downloaded and followed step by step the whole tutorial. I found it really useful and clear.

Nevertheless, only in the section "Installing Addons", I missunderstood something. Particularly, when it says

> This configures Mirage so that whenever Ember Data makes a GET request to /api/rentals, Mirage will return this JavaScript object as JSON. In order for this to work, we need our application to default to making requests to the namespace of /api. Without this change, navigation to /rentals in our application would conflict with Mirage.
> 
> To do this, we want to generate an application adapter: `ember generate adapter application`

it seemed to me like it was going to start getting the data from the `mirage/config.js` file. I even dropped [an issue at stackoverflow ](http://stackoverflow.com/questions/43415729/ember-js-tutorial-is-really-working-mirage-addon) to see what I was doing wrong. And someone kindly made me see that I had to "wait" until [Using Ember Data section](https://guides.emberjs.com/v2.12.0/tutorial/ember-data/).

So, wouldn't it be a good idea to emphasize or remark that in Installing Addons section we are just installing the Mirage add on, and it will take relevance only when de data model is defined and the data fetched?

This is what I did and what I am suggesting in my PR.


source/localizable/tutorial/installing-addons.md
```diff
@@ -112,4 +112,4 @@ export default DS.JSONAPIAdapter.extend({
 
 ```
 
-If you were running `ember serve` in another shell, restart the server to include Mirage in your build.
+If you were running `ember serve` in another shell, restart the server to include Mirage in your build. Note that at this point of the tutorial, the data will be still being fetched from the `app/routes/rentals.js` file. It will be on the [Using Ember Data](https://guides.emberjs.com/v2.12.0/tutorial/ember-data/) section where a data model will be included and the data will be fetched from our new `mirage/config.js` file.
```
toddjordan: 1. Split lines by punctuation.  (new line after each sentence)
2. "the data is still provided by the `app/routes/rentals.js` file" instead of "the data will be still being..."
3. Replace the last sentence with something simpler, like "We will make use of the mirage data we set up here in the upcoming section called [Using Ember Data](...)."

source/localizable/tutorial/installing-addons.es-ES.md
```diff
@@ -100,4 +100,4 @@ export default DS.JSONAPIAdapter.extend({
 
 ```
 
-If you were running `ember serve` in another shell, restart the server to include Mirage in your build.
\ No newline at end of file
+If you were running `ember serve` in another shell, restart the server to include Mirage in your build. Tener en cuenta que en este punto del tutorial, los datos todavía se van a estaro cogiendo de lo que se configuró en el archivo `app/routes/rentals.js`. Cuando lleguemos a la sección [Using Ember Data](https://guides.emberjs.com/v2.12.0/tutorial/ember-data/) veremos cómo utilizar un Data Model y cómo coger los datos desde nuestro nuevo fichero `mirage/config.js`.
```
toddjordan: Not sure if you need to use the Crowdin app for the Spanish translation.  @locks  is providing a translation here ok?

locks: Only the source English files should ever be changed, as instructed in https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md#style-guide.
We should address this soon, as it's confusing people for little benefit.

source/localizable/tutorial/installing-addons.md
pgbonino: Thanks for your comments. Totally agree in points 2 and 3. I  don't understand point 1, though. Why "new line after each sentence"? I can find many examples where several sentences are just separated by a ". "

So what do you exactly mean?

source/localizable/tutorial/installing-addons.es-ES.md
pgbonino: Thank you for your comments. I reset this file from HEAD, so the only file right now to be accepted is the plain English .md file.

toddjordan:
Thank you @pgbonino !  LGTM

Issue #1908, opened 2017-04-14
Tutorial 'Building a Complex Component' has a broken link


wemposd:
Hello,

At chapter 'Building a Complex Component' there is a broken link https://guides.emberjs.com/components/wrapping-content-in-a-component

travismchan:
Are you using the latest guide? If not, which version are you using? Seems to be working on my end.

locks:
I confirm that https://guides.emberjs.com/v2.12.0/tutorial/autocomplete-component/ is working for me too.

travismchan:
Oh, I see what OP is referring to. I've made a pull request for the fix (hopefully it's correct, first contribution). It was in reference to the paragraph with "This is an example of the block form of a component". I was looking at the side links.

locks:
Whoops, I read it as the link to the Building Complex Component guide was wrong.

locks:
PR with the fix has been merged. Thank you!

PR #1910, opened 2017-04-19
Update broken block form wrapping link

travismchan:
Broken wrapping component link via : https://github.com/emberjs/guides/issues/1908#issuecomment-294877900

### DIFF
--- a/source/localizable/tutorial/autocomplete-component.md
+++ b/source/localizable/tutorial/autocomplete-component.md
@@ -17,7 +17,7 @@ As before when we created the [`rental-listing` component](../simple-component),
 In our `app/templates/rentals.hbs` template file, we'll add a reference to our new `list-filter` component.
 
 Notice that below we "wrap" our rentals markup inside the open and closing mentions of `list-filter` on lines 12 and 20.
-This is an example of the [**block form**](/v2.12.0/components/wrapping-content-in-a-component) of a component,
+This is an example of the [**block form**](../components/wrapping-content-in-a-component) of a component,
 which allows a Handlebars template to be rendered _inside_ the component's template wherever the `{{yield}}` expression appears.
 
 In this case we are passing, or "yielding", our filter data to the inner markup as a variable called `rentals` (line 14).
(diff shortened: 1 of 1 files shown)


source/localizable/tutorial/autocomplete-component.md
```diff
@@ -17,7 +17,7 @@ As before when we created the [`rental-listing` component](../simple-component),
 In our `app/templates/rentals.hbs` template file, we'll add a reference to our new `list-filter` component.
 
 Notice that below we "wrap" our rentals markup inside the open and closing mentions of `list-filter` on lines 12 and 20.
-This is an example of the [**block form**](/components/wrapping-content-in-a-component) of a component,
+This is an example of the [**block form**](/v2.12.0/components/wrapping-content-in-a-component) of a component,
```
locks: Hi, this isn't quite right, because we'd need to update the links for every version!
It needs to be relative, so `../components` like in https://github.com/emberjs/guides/pull/1910/files#diff-e0faf55781ad58791e0623fbe42fe37dL2. Can you make that tweak, please?

travismchan: Thanks for the quick response locks, I've made the change. Just wondering, why do some use ../../ like on line 59?

locks: Good question and good catch, it'll have to be `../../`

The original URL is: https://guides.emberjs.com/v2.12.0/tutorial/autocomplete-component/
The destination URL is: https://guides.emberjs.com/v2.12.0/components/wrapping-content-in-a-component/

So you need to go two "up":
`../` -> https://guides.emberjs.com/v2.12.0/tutorial/
`../../` -> https://guides.emberjs.com/v2.12.0/
`../../components/wrapping-content-in-a-component` -> https://guides.emberjs.com/v2.12.0/components/wrapping-content-in-a-component

travismchan: Crystal clear explanation. Done!

locks:
And it's merged 💥

PR #1911, opened 2017-04-19
Update loading-and-error-substates.md

dshuang:


dshuang:
expanded example so the loading page will show

toddjordan:
thanks @dshuang 
@homu r+

PR #1912, opened 2017-04-19
Update method for removing welcome-page

counterbeing:
When going through the guide I noticed that the `application.hbs` template simply didn't exist, and the welcome page itself said to remove the add-on from the `package.json` file.


locks:
Hi @counterbeing, you are using an older version of Ember CLI. Starting with v2.11, I believe, we generate `application.hbs`, and switched to ember-welcome-page v2.

counterbeing:
Whoops. You're right. I had just followed the guide from the beginning and not checked to see if i'd already had a version installed. It appears that a version that I installed months ago was getting priority. Thanks!

locks:
No problem, thanks for sending a PR all the same!

PR #1913, opened 2017-04-19
Replace "A 'didUpdateAttrs' is called..." with "As 'didUpdateAttrs is called..."

richiethomas:
References issue #1914.

The full sentence is:

- "A `didUpdateAttrs` is called prior to rerender, you can use this hook to execute code when specific attributes are changed."

There seems to be a typo here.  The intended meaning seems to be something like the following: 

- "You can use the `didUpdateAttrs` hook to execute code when specific attributes are changed, since it is called prior to rerender."

I inferred that "As" was the intended first word of the sentence, since one of its meanings is analogous to "since" or "because".  As a suggestion, we could also just use either of those two words as well, to potentially make it less ambiguous (especially to non-native English speakers).


richiethomas:
@locks: I just updated 'as' to 'since'.

locks:
Thank you, details matter! 👌

richiethomas:
Glad to hear you say that. 👍

PR #1916, opened 2017-04-19
Update Ember.assert parameters order

hectorh30:
The example on the debbuging page is outdated with [the API](https://emberjs.com/api/#method_assert)


acorncom:
It is indeed, thx much! @homu r+

PR #1917, opened 2017-04-23
add docs for handling missing services

kellyselden:


source/localizable/applications/services.md
```diff
@@ -92,6 +92,20 @@ export default Ember.Component.extend({
 
 This injects the shopping cart service into the component and makes it available as the `cart` property.
 
+Sometimes a service may or may not exist, like when an initializer conditionally registers a service.
+Since normal injection will throw if the service doesn't exist, you must lookup the service instead.
```
toddjordan: May want to be more specific on throwing so its obvious to new users, like "will throw an error", or "will raise an error"

Also, may want to explain "lookup" by linking to the api doc.   like 
```
you must look up the service using Ember's [`getOwner`](https://emberjs.com/api/classes/Ember.html#method_getOwner) API instead. 
```

Issue #1572, opened 2016-07-22
Building a Complex Component - filtered results are not displayed on UI when typing/deleting fast


AndriusZid:
Hi, I have an issue on this tutorial with filtering.

When I filter city, San in results i see only Grand Old Mansion, but when i delete filter field quickly I don't see all results, but in console.log I see that GET request returns correct results with all three cities. It seems that UI missed it and didn't updated as expected.

UI test are passed okay.

![filter-ui](https://cloud.githubusercontent.com/assets/8049177/17056162/3dddbfd2-501a-11e6-85bc-075c6ef41ab3.gif)

rogergl:
I would help, but I have one problem: The code looks absolutely valid to me ?

toddjordan:
The code needs improvement because its asynchronous.  When you call to filter it goes through ember data and ultimately ajax calls, meaning the order that the results come back are not guaranteed.  One way to account for this is when the promise is fulfilled, to compare the query used for the search with the current value of the input field.   I've coded this into the super rentals repo in a [PR](https://github.com/ember-learn/super-rentals/issues/56).

Next step would be to update the guides to explain this.

toddjordan:
added some defensive code to the tutorial.  The ultimate solution may be to use ember-concurrency, but we are waiting until we get a "beginner's tutorial" as part of the docs, so what we have can be an "advanced tutorial"

PR #1918, opened 2017-04-24
Add section on handling async results, and break into subsections

toddjordan:
Addresses #1572 
image

image


source/localizable/tutorial/autocomplete-component.md
```diff
@@ -193,11 +200,78 @@ export default function() {
...
+```app/components/list-filter.js{+9,+10,+11,+19,+20,+21}
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  classNames: ['list-filter'],
+  value: '',
+
+  init() {
+    this._super(...arguments);
+    this.get('filter')('').then((allResults) => {
```
mschinis: This is more for another pull request, but might be useful if we set the search value on the query params and then use it here to retrieve the initial search results

PR #1920, opened 2017-04-24
Added a sentence indicating that the tagName property can be set to an empty string.

RyanNerd:
Added additional sentence indicating that the `tagName` property can be set to an empty string to not have any tag.

See: http://stackoverflow.com/questions/23368862/ember-js-view-without-wrapping-div


locks:
@RyanNerd this has been something we have not wanted to document, you could say it's a sort of "intimate" API. Unfortunately the behaviour can't be deprecated either until we have fragments or a similar way to replace it.

PR #1936, opened 2017-05-04
Extract TOC and pagination related code into plugin for Middleman

antonfefilov:
Initial work. Accordingly issue #10.


.ruby-version
```diff
@@ -0,0 +1 @@
+1.9.3-p551
```
locks: Hey, I'm curious, why 1.9?

locks:
Thank you for the work @antonfefilov! I just did a little tweak, because installing 1.9 might be a headache for some peeps.

PR #1942, opened 2017-05-15
Update ember-data tutorial link version

arthur-c:
Current link refers to an old version


locks:
Hm, this sound be a relative link instead.

acorncom:
Believe that's:

> this **should** be a relative link

locks:
@homu r+

PR #1947, opened 2017-05-21
Add actions to getting-started

rtablada:
This PR adds the `action` helper to the getting started section.

From feedback from developers assessing frameworks, there is often confusion that Ember has local component actions.
Since currently actions are only covered late into the Super Rents tutorial, this is a fair assumption to make.
We can clarify by adding this small `actions` section to the getting started section.
This makes `actions` feel more like a pivotal feature to the Ember ecosystem.

Also with the ability to create functions and actions this eases some of the question of "where is the Javascript" and allows developers to experiment more just with the knowledge from `getting-started` without having to dive deeper into guides.
This also gives us a way to expand knowledge into other areas in later tutorials instead of introducing actions later.

### DIFF
--- a/source/localizable/getting-started/quick-start.md
+++ b/source/localizable/getting-started/quick-start.md
@@ -243,7 +243,7 @@ create a `programmers` route that shows a list of famous programmers. By
 re-using the `people-list` component, you can do it in almost no code at
 all.
 
-## User Events
+## Click Events
 
 So far, your application is listing data, but there is no way for the user
 to interact with the information.
@@ -263,6 +263,8 @@ First add an `action` helper to the `li` in your `people-list` component.
 ```
 
 The `action` helper allows you to add event listeners to elements and call named functions.
+By default, the `action` helper adds a `click` event listener,
+but it can be used to listen for any element event.
 Now, when the `li` element is clicked a `showPerson` function will be called
 from the `actions` object in the `people-list` component.
 Think of this like calling `this.actions.showPerson(person)` from our template.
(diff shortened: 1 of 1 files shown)


source/localizable/getting-started/quick-start.md
```diff
@@ -243,6 +243,50 @@ create a `programmers` route that shows a list of famous programmers. By
 re-using the `people-list` component, you can do it in almost no code at
 all.
 
+## User Events
```
locks: I'm not super sure of the "event" framing here, because of DOM events.

source/localizable/getting-started/quick-start.md
```diff
@@ -243,6 +243,50 @@ create a `programmers` route that shows a list of famous programmers. By
...
+
+So far, your application is listing data, but there is no way for the user
+to interact with the information.
+In web applications, you often want to listen for user events like clicks or hovers.
```
locks: Especially because you have `click()` and `hover()` 👀

source/localizable/getting-started/quick-start.md
locks: I've also been watching a lot of TableTop, so the user doing an action is verbiage that sounds nice to me 👼

source/localizable/getting-started/quick-start.md
rtablada: @locks I see the component tag `click` and `hover` as things to be covered in later parts of the guides?
Also, what is future best practice that we want to show since Glimmer has "tagless" components where the actions are completely declared in the template rather than the component.

source/localizable/getting-started/quick-start.md
rtablada: So either "DOM Events" or "User Actions" but mixing the two wordings is weird?

PR #1949, opened 2017-05-22
Fix 404 page - Issue #1925

### LINKED ISSUE #1925: Fix the search box on our 404 page
If folks hit this a 404 page (try this one as an example: https://guides.emberjs.com/v2.12.0/models/working-with-relationships/) you may want to do a search for what you are after.  Sadly, it looks like our current 404 page (https://github.com/emberjs/guides.emberjs.com/blob/master/snapshots/404.html) may be using our older search system.

We'll need to reference the newer JS files (and put in the appropriate HTML) that is built as part of our guides pages these days ...

brianhays:
This PR fixes #1925 

The previous 404.html page was outside the 'source' directory so it was not being updated/generated via `middleman build` (this is why recent 404 page fixes such as #1919 are not carried over to the live site)

The new 404.html.erb template uses the default layout so it will receive updates to header, footer, etc. along with the rest of the site.


locks:
Appreciated!

Issue #1948, opened 2017-05-21
'Managing dependencies' assumes bower.json is already present in Ember 2.13 projects


myartsev:
As of Ember  2.13, bower.json is not automatically created for new Ember projects.
[This section](https://guides.emberjs.com/v2.13.0/addons-and-dependencies/managing-dependencies/#toc_bower) of the 2.13 guides assumes that it is. This seems like a good place to mention that you have to run `bower init` to create a bower config.

#https://github.com/emberjs/website/issues/2896
#https://github.com/ember-cli/ember-cli/issues/7029

PR #1950, opened 2017-05-22
Remove Bower mention from "Manage Dependencies"

locks:
While we still support Bower, Bower itself is on maintenance mode.
On top of that, you have to do some manual importing, which is
kinda out of scope, maybe?
Fixes https://github.com/emberjs/guides/issues/1948.


rtablada:
Agreed I think killing dead weight especially from the getting started area is helpful.
For people that are just using Ember sans-addons Bower won't come up.

Let's make sure we have a section in Ember-CLI docs but I think that may already be covered?

Issue #1931, opened 2017-04-29
Adding Google maps script to tests/index.html in Services and Utilities section of tutorial


AlexanderEllis:
I noticed some of my rental listing component integration tests failing after completing this section, and it looks like it was because it wasn't accessing the Google maps script correctly (resulting in "TypeError: Cannot read property 'maps' of undefined" when it attempted to access window.google.maps.  I had added the Google maps script to my app/index.html, but it still wasn't working.  

I checked with the completed code repo, and it looks like the [script was added to tests/index.html as well.](https://github.com/ember-learn/super-rentals/blob/master/tests/index.html#L29)  Once I added the script reference in tests/index.html, the tests passed as expected.

Should the tutorial page mention adding the script to the tests/index.html page as well, or is there something I am overlooking?

Thanks!

toddjordan:
no, we should.  good catch.

jakkubu:
Should it though? Shouldn't maps API be stubbed in all tests?

toddjordan:
@jakkubu is right, we shouldn't have a dependency on google maps.  if we do we need to update the tests to effectively stub the dependency.

toddjordan:
latest super rentals code should require no manual updates to the tests index.html.  checking repo tonight to make sure things are now working as expected.

toddjordan:
no longer an issue with the latest version of guides and super-rentals repo

PR #1956, opened 2017-05-27
Added instructions for adding Google Maps script to tests/index.html

AlexanderEllis:
This is a solution to #1931.  I added a quick section about adding the Google Maps script to the tests/index.html page.  Since it didn't fit in any of the other test subheaders, I added one for it, and I tried to keep the language similar to the previous mention of the Google Maps script.  I copied the tests/index.html page from the [super-rentals repo here](https://github.com/ember-learn/super-rentals/blob/master/tests/index.html).


yawboakye:
@AlexanderEllis Thanks for your contribution :+1: :smile:

toddjordan:
We don't want to reference/exercise the google maps api in our super-rentals tests.  If there's a missing callout or error in those tests we should stub the code in our tests to avoid it.

toddjordan:
Thank you for the contribution though.  if you are seeing errors, then lets circle back and see if there's a better way to resolve.

PR #1963, opened 2017-06-01
Improve QP deserialization explanation

Serabe:
Add a phrase about deserializing query params whose datatype is Array.

Fix emberjs/ember.js#15306


acorncom:
@homu r+

PR #1965, opened 2017-06-02
Update installing-addons.md

GuiluisAA:
The example is incorrect. By testing it in the Using Ember Data chapter it does not work. If I test with the final code available in Github, it works normally.


locks:
Thank you for the contribution @GuiluisAA, I went with #1962 as it had a twin PR for another part of the tutorial.

PR #1968, opened 2017-06-04
add missing change for integration test

DanielRuf:
Without the change the integration test would fail by default.


source/localizable/tutorial/hbs-helper.md
```diff
@@ -85,3 +85,29 @@ Each [argument](https://guides.emberjs.com/v2.12.0/templates/writing-helpers/#to
 
 Now in our browser we should see that the first rental property is listed as "Standalone",
 while the other two are listed as "Community".
+
+
+### Integration Test
+
+Update the content of the integration text to the following to fix it:
```
sbatson5: Should this be `integration test` instead of `integration text`?

source/localizable/tutorial/hbs-helper.md
```diff
@@ -85,3 +85,29 @@ Each [argument](https://guides.emberjs.com/v2.12.0/templates/writing-helpers/#to
...
+
+moduleForComponent('rental-property-type', 'helper:rental-property-type', {
+  integration: true
+});
+
+// Replace this with your real tests.
+test('it renders', function(assert) {
+  this.set('inputValue', '1234');
+
+  this.render(hbs`{{rental-property-type inputValue}}`);
```
sbatson5: This should be 
```
this.render(hbs`{{rental-property-type inputValue=inputValue}}`);
```
otherwise it won't render since we aren't assigning `inputValue`

source/localizable/tutorial/hbs-helper.md
DanielRuf: You are right will change it now =)

DanielRuf: Done. Sorry, seems I accidentally slipt onto 'x'.

source/localizable/tutorial/hbs-helper.md
DanielRuf: I don't think so as it is set and it is a property in the current context. Your proposed change does not affect / alter the output.

sbatson5: My bad, I thought this was a component test.

DanielRuf: No problem. But thanks for your feedback and taking a look at my PR.

PR #1970, opened 2017-06-05
fix wording

DanielRuf:
The sentence in the routing guide is a bit weird and it seems there is a word missing.
I've added it and improved the sentence.


locks:
👌

PR #1971, opened 2017-06-06
Update asynchronous-routing.md

Raj-keviv:


source/localizable/routing/asynchronous-routing.md
```diff
@@ -26,7 +26,7 @@ function fulfill(answer) {
 }
 
 function reject(reason) {
-  console.log(`Couldn't get the answer! Reason: ${reason}`);
+  console.log('Couldnot get the answer! Reason: ${reason}');
```
sbatson5: "Could not" should be 2 words. Also, this would need to still be a template string. Single quotes will display "${reason}" instead of the value of `reason`

locks:
Closing has the breaks the logic.

PR #1973, opened 2017-06-14
[Tutorial:Planning Your App] Fix Test Code Diff Display

MDjavaheri:
Previously, the two snippets of test code were being displayed as inline code and not as code blocks. As such, the diff didn't show properly (+'s and -'s). This should fix it.

Example: 
screen shot 2017-06-14 at 4 14 33 pm

Instead of something like this:
screen shot 2017-06-14 at 4 16 37 pm


locks:
👍

PR #1974, opened 2017-06-15
[Tutorial:Simple Component] Fix first integration test

MDjavaheri:
The first integration test on this page is missing an expected value.


locks:
Appreciated!

PR #1976, opened 2017-06-19
improved dev env setup instructions for new contributors

jenweber:
This PR follows up on discussion in the Ember Learning Slack.

I gave my Mac a fresh hard drive and documented the things I had to do to get the Ruby dev environment set up. I think that the details on rbenv are necessary since many devs do not work with Ruby and given the difficulty that a few new contributors and I had.


locks:
👌

amyrlam:
@jenweber this is great, thank you!

PR #1977, opened 2017-06-19
[Tutorial:Ember Data] add +s to first revision of model file

MDjavaheri:
Just adds some aesthetic green highlighting to the initial additions of data attributes to `rental.js` model.


MDjavaheri:
Realized this doesn't fit with the general style of the guide.

PR #1978, opened 2017-06-19
[Tutorial:Building a Complex Component] Add +'s and -'s to modified rentals template

MDjavaheri:
Adds some aesthetic prettiness to the docs to show which lines are being modified.


locks:
Appreciated!

Issue #1554, opened 2016-07-05
Google Maps requires API key - tutorial broken


EvanKnowles:
The tutorial is currently broken, with the message:

testem.js:785 This service requires an API key.  For more information on authentication and Google Maps Javascript API services please see: https://developers.google.com/maps/documentation/javascript/get-api-key

It seems that, in general, the Google Maps API should be loaded from Google and not cached locally in  any case.

toddjordan:
@EvanKnowles The tutorial seems to work for me, but I do get the message you mention above.  I believe this is because Google will give you a certain number of map lookups without a key for your server (and then requires you to pay them).  For the purpose of an intro tutorial, I'm ok with that limitation.  

It will probably be simpler for people downloading the tutorial source to just have a link to google for the library.  We went the vendor route at first to introduce that concept to new users.  It worth holding this open for now to explore alternatives.

ve3:
I confirm this error. Does anyone show an example how to set the Google Maps API key?

toddjordan:
@ve3 I/you do not need the api key to get the tutorial working locally.  You'll get a warning but the maps should show.  If the maps do not show you may need to re-run the curl command.  You shouldn't need a key for the purposes of initial running the tutorial.  The curl approach may not be the best way to incorporate the google maps library, but it was done to demonstrate how one might include a third party vendor library.  We'll research alternatives that are more intuitive for tutorial followers.

ve3:
If no API key, the maps do not show. Re-run curl nothing change.
For me, it needs API key.

toddjordan:
odd.  historically you've been allowed to use the api without a key up to a per day limit.  I just tried it on a local copy of the super-rentals repo and map display works still.  will have to research some more.  An alternative would be to update index.html with maps in one of the script tags and delete the js file from vendor.  You can see this in the super-rentals repo: https://github.com/ember-learn/super-rentals

pmaisonnave:
If think ve3 is right: as announced by Google Geo Developpers, keyless access is no longer supported (but I was said it might still be the case for some "trusted" hosts) : see the following link:

[Building for Scale: Updates to Google Maps APIs Standard Plan](https://googlegeodevelopers.blogspot.fr/2016/06/building-for-scale-updates-to-google.html
)

serproqnx:
Same error, can't complete tutorial :(

mtrenkmann:
[...]

PR #1980, opened 2017-06-21
Replace section on adding maps script with addon

toddjordan:
Addresses #1554 by installing google maps via an addon, and allowing the developer to provide a google maps key via an environment variable.  See https://github.com/ember-learn/ember-simple-google-maps

### DIFF
--- a/source/localizable/tutorial/service.md
+++ b/source/localizable/tutorial/service.md
@@ -12,37 +12,21 @@ There are several ways to include 3rd party libraries in Ember.
 See the guides section on [managing dependencies](../../addons-and-dependencies/managing-dependencies/)
 as a starting point when you need to add one.
 
-The [Google Maps API](https://developers.google.com/maps/documentation/javascript/tutorial) requires us to reference its library from a script tag.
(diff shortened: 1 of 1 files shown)


source/localizable/tutorial/service.md
```diff
@@ -12,37 +12,21 @@ There are several ways to include 3rd party libraries in Ember.
...
-    {{content-for "body-footer"}}
-  </body>
-</html>
+Google Maps requires an API key for deployment.
+You can [Generate an API key](https://developers.google.com/maps/documentation/javascript/get-api-key)
+from Google.
+Add your new API key to the application by stopping the server and restarting it with the environment variable, `GOOGLE_MAPS_API_KEY`.
 
+```shell
+GOOGLE_MAPS_API_KEY=<your key here> ember s
```
acorncom: Should we add the above to the deploy page instead?  Other than that question, I'm 👍 too ...

toddjordan: I think some might not even do the deploy section, so I'll leave it here and then also mention in deploy (will create a separate issue)

Issue #1637, opened 2016-09-08
Flesh out documentation for component lifecycle hooks.


rwjblue:
Add information to the guides (likely in [The Component Lifecycle](https://guides.emberjs.com/v2.7.0/components/the-component-lifecycle/)) that explains the concept that all DOM interaction should be focused in specific hooks and that those hooks are not ran in FastBoot.

This is a guide specific issue for the work discussed in https://github.com/emberjs/ember.js/issues/14235.

toddjordan:
From Fastboot documentation https://ember-fastboot.com/docs/user-guide : 

## Designing Components

### Use didInsertElement for client-side DOM manipulation

In FastBoot, we do not invoke either didInsertElement or willInsertElement hooks. If your components have any direct DOM manipulation you would do it in those hooks.

### Lifecycle Hooks in FastBoot

FastBoot calls the init, didReceiveAttrs, didUpdateAttrs, willRender and willUpdate hooks. Any code in these hooks will be run inside of FastBoot and should be free of references to browser APIs or DOM manipulation.

Issue #1469, opened 2016-06-06
Update 'Application Concerns' sections to discuss fastboot


Gaurav0:
So some of the problems people run into with fastboot are that:

1. Instance initializers are run in parallel for each user connecting within fastboot. You should not share state between them.

2. Despite its name, the `afterRender` run loop queue is executed in fastboot, unless triggered from someplace that isn't run, like `didInsertElement`

3. Services are no longer singletons, and now exist per each fastboot request and destroyed when completed. Same with dependency injections.

acorncom:
We're still working out how / whether to integrate Fastboot docs into guides (seems like we should, but how remains a question).  Will get to this at some point, but not right away (Fastboot hasn't hit v1.0 yet)

locks:
1. Sounds like good advise, regardless.
2. Why wouldn't `afterRender` run?
3.  I need some clarification

Gaurav0:
2. I (and plenty of others) had assumed it wouldn't because of its name, 'afterRender' (hey, the element is rendered to the DOM!) and that most of the time I used it to do jQuery stuff after DOM was rendered. It is also commonly used to get around deprecation warnings and cause double renders. (Not that these uses were good ideas!)

3. Very often when writing Ember, services are used as places to store 'global' data and it is generally assumed that there is only one copy of any service. This allowed for some shortcuts like adding things to the class (prototype) and not the instance. For example, writing

```
Ember.Service.extend({
  arr: []
})
```

instead of

```
Ember.Service.extend({
  init() {
    this.set('arr', []);
  }
});
```

PR #1982, opened 2017-06-21
Add SSR section

locks:
Fixes https://github.com/emberjs/guides/issues/1469.
Fixes https://github.com/emberjs/guides/issues/1637.

The idea is to put most of the documentation on the standalone site in the Guides.
The "Addon Author Guide" I'm not sure where to put, because we don't really have a good place to put resources for addon authors. Maybe a new section on the website altogether?

cc @kratiahuja @danmcclain

### DIFF
--- a/data/pages.yml
+++ b/data/pages.yml
@@ -249,6 +249,14 @@
     - title: "Debugging"
       url: "debugging"
 
+- title: "Server-side Rendering"
+  url: 'ssr'
+  pages:
+    - title: "What is Server-side Rendering?"
+      url: "what-is-ssr"
+    - title: "Setting up FastBoot"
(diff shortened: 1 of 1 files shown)


kratiahuja:
Should we point it to the fastboot website? I worry we have to keep updating both.

locks:
@kratiahuja the idea is to only have one website ;P

kratiahuja:
aah then i guess we also need how to deploy fastboot apps and various options.

locks:
Already opened another PR for a [Deployment](https://github.com/emberjs/guides/pull/1983) section ;P

locks:
* note: `console.log` goes to the terminal

PR #1983, opened 2017-06-21
Add Deploying section

locks:
This should tie in the various deploying things scattered throughout the guides.
It should mention `ember-cli-deploy`.


rwjblue:
Should also mention the difference between debug and prod builds.

PR #1984, opened 2017-06-21
prefer calling super at the end of willDestroyElement()

Dhaulagiri:
I've always been told, and always logically thought, that it's good practice to call super last in teardown hooks, but the docs have always demonstrated the opposite.  This PR adjusts that to match what I _think_ is the right way to do this.


locks:
@Dhaulagiri thank you for the help 👍

PR #1992, opened 2017-06-29
Document build targets

cibernox:
@locks I created a new section as discussed on Slack. I didn't put too much love on the form of the content of the page yet, but I want to check that generally you agree on the place on the guides.


acorncom:
Thanks for the PR @cibernox, but even more, thanks for helping get that feature in 👍

PR #1993, opened 2017-07-02
Remove the un-filtered model

lastcoolnameleft:
I ran through the walkthrough and upon load, it displayed the rentals twice.  I assume the intent is to replace the unfiltered model with the filtered model.


locks:
👍 appreciate the keen eye!

PR #1999, opened 2017-07-11
Fix broken link to qunit api docs

nlhkabu:
https://api.qunitjs.com/category/assert/ is returning a 404.  It appears as though the new URL is https://api.qunitjs.com/assert/


locks:
@nlhkabu appreciated :) I think they recently changed some URLs around, so there might be other instances in our docs.

PR #2004, opened 2017-07-22
QUEST | Convert code samples to RFC #176 imports

locks:
[RFC #176](https://github.com/emberjs/rfcs/blob/master/text/0176-javascript-module-api.md) introduces new modular imports so that you can import only the parts of Ember that you require inside that file.

We need your help converting all of the code samples in the Guides to the new style of imports!

## Contributing

1. Pick a section from the list below. If a section is already taken and you would like to help, please contact the person it was assigned to directly and coordinate with them.
2. Using the correspondence tables in [ember-rfc176-data](https://github.com/ember-cli/ember-rfc176-data):
  2.1. Update `import` statements in code sample to the new imports.
  2.2. Update links to API docs. For example, change the link text from `Ember.Component` to `Component`.
3. Make a PR **against the `new-imports` branch**, which is this one.

#### I can't find the class/method in the correspondence table

Check if the API is **private** or **public**.

If it is **private**, the Guide should be changed to not use it.
You can post here for help, or in your own PR.

If it is **public**, there might have been a lapse.
Comment in https://github.com/ember-cli/ember-rfc176-data/issues/12 and we'll help you sort it out!

#### Should I rebase? Why do I have to target the `new-imports` branch?

We will take care of rebasing and otherwise updating this branch as works continues.
We need to do this until 2.16 is released because we have Continuous Integration on the master branch, so any PRs merged get deployed as soon as CI finishes building.

## Guides section list 

- [x] Getting Started
- [x] Tutorial
- [x] The Object Model
- [x] Routing
- [x] Templates
- [x] Components
- [x] Controllers
- [x] Models
- [x] Application Concerns
- [x] Testing
- [x] Ember Inspector
- [x] Addons and Dependencies
- [x] Configuring Ember.js
- [x] Contributing to Ember.js
- [x] Glossary

### DIFF
--- a/source/object-model/computed-properties.md
+++ b/source/object-model/computed-properties.md
@@ -81,7 +81,7 @@ import EmberObject, { computed } from '@ember/object';
 let obj = EmberObject.extend({
   baz: {foo: 'BLAMMO', bar: 'BLAZORZ'},
 
-  something: Embercomputed('baz.foo', 'baz.bar', function() {
+  something: computed('baz.foo', 'baz.bar', function() {
     return this.get('baz.foo') + ' ' + this.get('baz.bar');
   })
 });
--- a/source/object-model/computed-properties.md
+++ b/source/object-model/computed-properties.md
@@ -12,14 +12,14 @@ We'll start with a simple example.
 We have a `Person` object with `firstName` and `lastName` properties, but we also want a `fullName` property that joins the two names when either of them changes:
 
 ```javascript
-import Ember from 'ember':
+import EmberObject, { computed } from '@ember/object';
 
-Person = Ember.Object.extend({
+Person = EmberObject.extend({
   // these will be supplied by `create`
   firstName: null,
   lastName: null,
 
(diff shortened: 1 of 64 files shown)


acorncom:
Marking testing as completed, thanks for your help @aureliosaraiva !

source/object-model/computed-properties.md
```diff
@@ -70,10 +72,12 @@ You surround the dependent properties with braces (`{}`), and separate with comm
...
 
 ```javascript
-let obj = Ember.Object.extend({
+import EmberObject, { computed } from '@ember/object';
+
+let obj = EmberObject.extend({
   baz: {foo: 'BLAMMO', bar: 'BLAZORZ'},
 
-  something: Ember.computed('baz.foo', 'baz.bar', function() {
+  something: Embercomputed('baz.foo', 'baz.bar', function() {
```
juwara0: One of the `Embercomputed` conversions ^ got missed.

locks: Thanks :)

PR #2019, opened 2017-08-08
Update homu links.

luketheobscure:
http://homu.io/ isn't up anymore.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -19,11 +19,6 @@ Before the review period starts, the previous version should be re-released with
 
 # Pull Requests
 
-You should use [homu](http://homu.io) when accepting pull requests.
-You can read about the available commands in the front page.
-
-The Guides repository has homu [configured to auto-squash commits](http://homu.io/r/emberjs/guides).
-
 Before merging you should check the following:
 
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -19,10 +19,10 @@ Before the review period starts, the previous version should be re-released with
 
 # Pull Requests
 
-You should use [homu](http://homu.io) when accepting pull requests.
+You should use [homu](https://github.com/barosl/homu) when accepting pull requests.
 You can read about the available commands in the front page.
 
-The Guides repository has homu [configured to auto-squash commits](http://homu.io/r/emberjs/guides).
+The Guides repository has homu configured to auto-squash commits.
```
luketheobscure: Is there a different URL this should point to?

locks: We don't actually use any bot at the moment, though we could try setting up bors-ng.

luketheobscure: Then should I just remove references to homu?

locks: I think we should, and we can update if we start using bors.

luketheobscure: All references to homu have been removed.

PR #2021, opened 2017-08-11
Updated 'getOwner' with ES6.

vishusharmagit:
Just updated syntax to use ES6.


locks:
See https://github.com/emberjs/guides/pull/2022#issuecomment-321749192. Thank you!

PR #2022, opened 2017-08-11
Updated Page For ES6.

vishusharmagit:
Just updated whole page after first request.


locks:
Hi @vishusharmagit, thank you for your contribution but we don't want to suggest destructuring like this.
There is also an effort on-going at https://github.com/emberjs/guides/pull/2004 to convert to the new module API that should be released as default with Ember 2.16.

vishusharmagit:
Thanks for updating me about [ember-rfc176-data](https://github.com/ember-cli/ember-rfc176-data).

PR #2027, opened 2017-08-12
Fix typo "contact" -> "content"

PoziWorld:
Before:
Because we only plan to display static contact on our about page

After:
Because we only plan to display static content on our about page


toddjordan:
good catch ❤️

PR #2028, opened 2017-08-12
Fix invalid use of ADDRESS tag inside of P

PoziWorld:
P can only contain phrasing content.
ADDRESS does not belong to phrasing content category.

http://w3c.github.io/html/grouping-content.html#elementdef-p
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

http://w3c.github.io/html/dom.html#phrasing-content-2
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content

http://w3c.github.io/html/grouping-content.html#elementdef-address
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address


toddjordan:
Thanks.  Will have to check this out to make sure page style is not affected.  (browsers include some default styling with the `p` tag).  Since this is changing code, if you could also pr to https://github.com/ember-learn/super-rentals that would be helpful.

PR #2025, opened 2017-08-12
Remove repeated sentence

PoziWorld:
“Show rentals on the home page” and “List the available rentals” are the same thing, aren't they?

PR #2029, opened 2017-08-13
Change "on" -> "as" in "Show rentals on the home page"

PoziWorld:
Line 87 uses "as":
```
test('should show rentals as the home page', function (assert) {
```

Home page redirects to /rentals


source/tutorial/acceptance-test.md
```diff
@@ -6,7 +6,7 @@ Here's a look at what we want to build before we get started.
 
 Let's work through what we want to do on the home page. We want our application to:
 
-* Show rentals on the home page
```
toddjordan: I think what's there is ok.  This list is an abstract list showing what we want the application to do.  Its not referring to the actual "rentals" route yet.

PoziWorld: @toddjordan, what is the difference between this and “List the available rentals” (line 12)?

In my opinion, it is either a repetition (as I assumed in https://github.com/emberjs/guides/pull/2025) or should say “Show rentals as the home page” to avoid repetition.

What do you think?

yawboakye:
@locks Looks like we can safely close this without disturbing the universe

acorncom:
Agreed. Thanks @yawboakye

PR #2034, opened 2017-08-13
Fix "lets" -> "let's" in "Finally, lets add..."

PoziWorld:


toddjordan:
Thanks @PoziWorld 👍 
@homu r+

PR #2036, opened 2017-08-13
Fix "lets" -> "let" in "to instantiate..., and lets"

PoziWorld:


source/tutorial/service.md
```diff
@@ -239,7 +239,7 @@ For our service unit test, we'll want to verify that locations that have been pr
 We will isolate our tests from actually calling Google Maps by stubbing our map utility.
 On line 6 of `maps-test.js` below we create an Ember object to simulate the behavior of the utility, but instead of creating a google map, we return an empty JavaScript object.
 
-Unit tests use the function called `this.subject` to instantiate the object to test, and lets the test pass in initial values as arguments.
+Unit tests use the function called `this.subject` to instantiate the object to test, and let the test pass in initial values as arguments.
```
toddjordan: I think you are on the right track here.  I think it would be even clearer to say "...to instantiate the object ~~to~~ _under_ test, and _to_ let the test pass in additional..."

locks:
Thanks for the contribution @PoziWorld.

PR #2043, opened 2017-08-17
embre URL module : drive -> derive

erreur404:
embre URL module : drive -> derive
I'm not entirely sure which one is the best, so I thought I'd just suggest the change.
In my understanding : to drive the state by URLs is to make the state change with URLs
to derive the state from URLs is to get the app state from the info contained in the URL


locks:
Hi @erreur404, I think the original is correct. The idea is that your app state is driven by the router, as changes to the URL trigger transitions, which makes a route set up the respective controller and template, and optionally loading data.
Do you have a suggestion to rewrite the whole sentence to make it clearer?

locks:
Closing this for inactivity. Feel free to reopen and carry out the conversation! :)

PR #2044, opened 2017-08-17
return properly formatted result object

stdavis:


stdavis:
Superceded by: https://github.com/emberjs/guides/pull/2045

PR #2045, opened 2017-08-17
remove unused parameter and make tests pass

stdavis:
- remove used `val` parameter to allow for linting to pass.
- update the format of the second `filterByCity` mock to the newer `results: [], query: ''` format.

### DIFF
--- a/source/tutorial/autocomplete-component.md
+++ b/source/tutorial/autocomplete-component.md
@@ -326,7 +326,7 @@ moduleForComponent('list-filter', 'Integration | Component | filter listing', {
 test('should initially load all listings', function (assert) {
   // we want our actions to return promises,
   //since they are potentially fetching data asynchronously
-  this.on('filterByCity', (val) => {
+  this.on('filterByCity', () => {
     return RSVP.resolve({ results: ITEMS });
   });
 });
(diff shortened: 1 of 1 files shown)


toddjordan:
Thanks @stdavis can you also check the super rentals code repository and update if necessary? (it may not be since the tests there are currently passing, at last check)

https://github.com/ember-learn/super-rentals

stdavis:
They look a little different:
https://github.com/ember-learn/super-rentals/blob/26189cec932f35ad2885ef2d29f79b4ffbfb09a8/tests/integration/components/list-filter-test.js#L15-L18

https://github.com/ember-learn/super-rentals/blob/26189cec932f35ad2885ef2d29f79b4ffbfb09a8/tests/integration/components/list-filter-test.js#L39-L49

I assume that you would like me to update the PR to match them?

toddjordan:
Interesting... yeah, subtly different.   

for the first block, if we add the assert, we should explain why its there it in the text.   Alternatively we can take the assert out of the code repo (which I'm ok with).

For the second, yeah, just update the tutorial to match.

stdavis:
OK. I removed the unnecessary assert from the code repo in the first test and then updated the guide to match the code repo for the second test.

toddjordan:
Great Job

PR #2047, opened 2017-08-25
Fix "propertyType" attribute in Mirage config

riazjahangir:
JSON data should be dasherized. Earlier pages correctly used "property-type" for the mock data.


locks:
Thanks for the keen eye :)

PR #2048, opened 2017-08-25
corrected a spelling mistake

KenyanGeek:


Serabe:
Thank you for the PR!

In this case, the sentence is right as it is so we are closing it.

Don't hesitate on fixing other issues though!

Thanks!

PR #2050, opened 2017-08-27
Added section for const and let variable declarations. Cleaned up intro. 

jaredgalanis:
As requested by #1751 @locks

### DIFF
--- a/source/getting-started/js-primer.md
+++ b/source/getting-started/js-primer.md
@@ -2,18 +2,26 @@ While the Guides [assume you have a working knowledge of JavaScript](/#toc_assum
 when learning a new framework it can be hard to know what is JavaScript and what is framework,
 especially if you are newer to the ecosystem and less familiar with JavaScript.
 
-On top of that, many new features were introduced to JavaScript with the release of newer specifications like ECMAScript 2015 (also known as ECMAScript 6 or ES6, https://developer.mozilla.org/en/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla). With the current yearly cadence of ECMAScript specifications, developers might have not had time to get familiar with all the new features.
+On top of that, many new features were introduced to JavaScript with the release of newer specifications like ECMAScript 2015 (also known as ECMAScript 6 or ES6, https://developer.mozilla.org/en/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla).
+With the current yearly cadence of ECMAScript specifications,
+developers might have not had time to get familiar with all the new features.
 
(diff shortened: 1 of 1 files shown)


source/getting-started/js-primer.md
```diff
@@ -1,14 +1,141 @@
...
+With the current yearly cadence of ECMAScript specifications,
 developers might have not had time to get familiar with all the new features.
 
 In this guide we will be covering some common JavaScript code patterns that appear in Ember applications,
+and in particular those that explore the new features provided in ES6,
 so you can get a clearer sense of where the language ends and the framework starts.
 
+## `const` and `let` variable declarations
+
+Unlike some programming languages, in JavaScript where your variables are created depends on how you declare them.
```
rtablada: I think "Unlike some programming languages" may be excessive

rtablada: "where your variables are created..."

"In JavaScript, there are a few ways to declare variables that change when and how variables can be used."

source/getting-started/js-primer.md
```diff
@@ -1,14 +1,141 @@
...
+Because of this, classic `var` variable declarations in JavaScript can sometimes be tricky and create confusion over a variable's life cycle.
+
+By introducing the block-level variables (also sometimes called bindings) `const` and `let` in ES6, variable scope has been made easier to more explicitly control.
```
rtablada: These last two sentences can be combined. Move the "tricky" references to the section on `var` instead of as an intro sentence. I also don't particularly like that it is editorializing.

I think there's gotta be a good way to quickly mention `var` is classic, `const` and `let` were introduced in `ES2015` and that's it for the intro.

source/getting-started/js-primer.md
```diff
@@ -1,14 +1,141 @@
...
+
+### `var`
+
+In specifications prior to ES6, `var` variable declarations are *hoisted* to the top of the function they are declared irrespective of where in the function they are actually declared,
```
rtablada: Even in the ES2015 spec, var is still hoisted.

rtablada: You mention "hoisting" without saying what it is.

source/getting-started/js-primer.md
```diff
@@ -1,14 +1,141 @@
...
+    return null;
+  }
+}
+```
+
+### `const` and `let`
+
+#### Block-Level Bindings
+
+On the other hand, `const` and `let` are both block-level bindings.
```
rtablada: Remove "on the other hand"

source/getting-started/js-primer.md
jaredgalanis: Thanks @rtablada, all these comments are super helpful. I've updated in a commit just now. Let me know what you think. Flagging for @locks too.

PR #2051, opened 2017-08-30
Fixes to inconsistent variable names in autocomplete component

cpalmerbright:
* variable name changed between different versions of the file on the same page
* a line of the new code was also changed from earlier version but this wasn't reflected in the change highlighting
* change highlighting is clearer, and original variable name has been used


locks:
Thank you for the contribution @cpalmerbright, it looks much tidier now :)

PR #2059, opened 2017-09-15
WIP updates to Ember Data Overview

jenweber:
This pull request is for the purpose of comments and criticisms of my total rewrite of https://guides.emberjs.com/v2.15.0/models/ . I was going to keep most of it and reorganize, but much of the content is 2-3 years old, back when Ember Data needed to be "sold" to the existing community. We're in a different place now.

This is a first draft which I fully expect to revise based on feedback and to add in the links to other relevant sections of the guides.

Please be candid in your critiques. This content strongly affects new dev UX. 

[Link to markdown of the Models Introduction rewrite](https://github.com/jenweber/guides/blob/1c42cfa2f52e83c9680650033ad1911df44abc01/source/models/index.md)


jenweber:
Feedback log from Slack:

@sivakumar-kailasam suggests an FAQ section on what kinds of objects to return from hooks if Ember Data is not being used

source/models/index.md
```diff
@@ -176,129 +139,13 @@ Frameworks (or apps built on those frameworks) don't need to write
...
-for long, Ember Data will automatically make a request in the
-background each time a cached record is returned from the store. When
-the new data comes in, the record is updated, and if there have been
-changes to the record since the initial render, the template is
-re-rendered with the new information.
-
-## Architecture Overview
+## How are server communications handled? Are they asynchronous?
 
 The first time your application asks the store for a record, the store
```
yawboakye: @jenweber Is it possible to gloss over some details here? For example, throughout the document we haven't revealed the store's underlying abstraction (how it retrieves and store records) until now. We can also replace "persistence layer" with "back end" since that's the term you used throughout. Here's a first attempt at an update:

The first time you try to retrieve a record, the store checks if it has a local copy. If there's no local copy then a request will be made to the back end. This is an actual network call. If a local copy is found then it is returned immediately. Since the record could have changed on back end, between the first time it was requested and now, the store automatically refreshes it in the background. This is also a network call, although it can be prevented.

source/models/index.md
```diff
@@ -1,162 +1,125 @@
...
+switching over to a new server API could be as simple as writing some new
+adapters and serializers. This is one of the reasons why Ember is a good
+choice for long-lived or complex applications.
+
+## Do I need to use Ember Data in my Ember app?
+No. If you prefer to manage requests yourself, you can. Some developers pick
+and choose which Ember Data tools to utilize, such as fetching information
+manually but still loading it into the Store.
+
+## Why is JSON API the default?
```
yawboakye: @jenweber Worth mentioning that JSON API and Ember Data have helped each other grow. Not sure any other JSON API specification provides the surface that Ember Data needs in order for some key abstractions to work (like magic). Key abstractions such as side-loading, loading relationships (`post.get('comments')`)

jenweber:
I'm closing my own PR and will open smaller PRs that can be more easily reviewed and merged.

PR #2061, opened 2017-09-16
Clarify quickstart

adrianbw:
Clarify that the process for creating a new route requires a new terminal window, not the one you're running the application in.


locks:
Appreciated :)

PR #2062, opened 2017-09-17
Update installing-addons.md

adrianbw:
Fix typo.

### DIFF
--- a/source/tutorial/installing-addons.md
+++ b/source/tutorial/installing-addons.md
@@ -146,6 +146,6 @@ export default DS.JSONAPIAdapter.extend({
 
 ```
 
-If you were running `ember serve` in another shell, restart the server to include Mirage in your build.
+If you were running `ember server` in another shell, restart the server to include Mirage in your build.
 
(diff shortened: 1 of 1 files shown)


acorncom:
Thanks @adrianbw but the base command is actually `serve` 😀  The default is `serve` with `server` or `s` being aliases (see https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js#L20) We've had this type of PR sent in before 😉

adrianbw:
How about updating the tutorials to use `ember serve` instead of `ember server`, then? If you've done nothing but the tutorials (which I'd assume is the case here), then this is pretty confusing (in fact, it was so "obviously" a typo that I didn't see a reason to search for it).

acorncom:
Ah, are they mixed? That is indeed confusing. Yes, we'll get that fixed one way or another. Thanks for pointing it out!

adrianbw:
Yeah. I'll take care of it as I go through the tutorial.

acorncom:
@toddjordan @locks Which command do you think we should be defaulting to? I'm inclined toward `ember serve` instead of `ember server`, mainly because it might help it seem less like "Ember has a server I should be running in production". Thoughts?

locks:
We're normalizing everything to `ember serve`, since it's the base command.

PR #2063, opened 2017-09-17
Use `server` instead of `serve`

adrianbw:
Use the base command instead of an alias (see #2062).


yawboakye:
@adrianbw :+1:  :smile: I think the title should be "Use serve instead of server". You swapped "server" and "serve" around. If it doesn't bother you too much you can edit the title.

adrianbw:
Oh, crud, so I did. Sorry!

PR #2065, opened 2017-09-18
Es primer arrow func

jaredgalanis:
As requested by #1751 @locks. Heads up @rtablada.

### DIFF
--- a/source/getting-started/js-primer.md
+++ b/source/getting-started/js-primer.md
@@ -51,11 +51,21 @@ export default Ember.Component.extend({
     ]
   }
 });
-```
 
 ## Arrow Functions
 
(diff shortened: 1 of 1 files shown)


source/getting-started/js-primer.md
```diff
@@ -52,3 +52,68 @@ export default Ember.Component.extend({
...
+}
+```
+Is just like:
+```javascript
+let someArrowFunc = function(value) {
+  return value;
+}
+```
+
+
```
yawboakye: @jaredgalanis Nice! Can you include an example of using an arrow function as a callback function?

jaredgalanis: Yes, no problem @yawboakye, thanks for the feedback!

jaredgalanis:
@yawboakye how does this look to you?

PR #2066, opened 2017-09-18
Minor stuff I came across following the tutorial

adrianbw:
autocomplete-component.md: in one example, the module and consts were switched.
hbs-helper.md: one example did have code addition/deletion that I believe it should have from seeing other examples.
service.md: The first use of camelize() really threw me when I was reading the docs, so I added a note about it.

### DIFF
--- a/source/tutorial/autocomplete-component.md
+++ b/source/tutorial/autocomplete-component.md
@@ -397,12 +397,12 @@ import hbs from 'htmlbars-inline-precompile';
 import wait from 'ember-test-helpers/wait';
 import RSVP from 'rsvp';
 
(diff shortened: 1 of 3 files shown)


locks:
Thank you for the tweaks, very appreciated!
I made a small tweak to `service.md` so it doesn't use the prototype extension. Hope that usage of the API is clear.

adrianbw:
Sure, that works. I'd still like to put a note about those helpers under that first code sample. Thoughts?

locks:
@adrianbw are you on Slack? I'd like to chat on a higher-bandwidth medium.

locks:
We chatted on Slack and it opened the question of whether we even need/want to camelize, and it seems the answer is no. No code is better than no code!

PR #2067, opened 2017-09-18
Added object literal section

jaredgalanis:
As requested by #1751 @locks. Heads up @rtablada.

### DIFF
--- a/source/getting-started/js-primer.md
+++ b/source/getting-started/js-primer.md
@@ -55,9 +55,11 @@ export default Ember.Component.extend({
 
 ## Object Literal Shorthand
 
+In ES6 some functionality was added that allows the use of terser syntax for object literal properties and methods, along with some enhanced functionality for property names.
+
 ### Property intializer syntax
 
-In ES6 properties can be set on object literals without by including the name of the property without the typical colon and value if the object property the same as the local variable you intend to set as the value of that property. This helps reduce duplication. 
+Now properties can be set on object literals without by including the name of the property without the typical colon and value if the object property the same as the local variable you intend to set as the value of that property. This helps reduce duplication. 
 
 For example:
 ```javascript  
@@ -68,7 +70,7 @@ function someFunc(thing1, thing2) {
   };
 }
 ```
-You can now also set duplicate properties on objects without throwing an error when in strict mode.
+However, you can now also set duplicate properties (where properties named the same) on objects without throwing an error when in strict mode.
 
 ### Concise Methods
 
@@ -94,6 +96,6 @@ let some = "thing ";
 let object = {
 	[some + "1"]: "This is the thing."
 };
-console.log(["thing 1"]); // ouputs "This is the thing."
+console.log(["thing 1"]); // outputs "This is the thing."
 ```  
 
--- a/source/getting-started/js-primer.md
+++ b/source/getting-started/js-primer.md
@@ -140,3 +140,5 @@ let person = {
 ```
 
 The one difference with concise methods is that they can use super to access an object's prototype whereas regular methods cannot.
+
+For further reference on property initializer and concise method syntax see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer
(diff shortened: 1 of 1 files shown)


source/getting-started/js-primer.md
```diff
@@ -52,3 +52,50 @@ export default Ember.Component.extend({
...
 });
 ```
+
+## Object Literal Shorthand
+
+In ES6 some functionality was added that allows the use of terser syntax for object literal properties and methods, along with some enhanced functionality for property names.
+
+### Property intializer syntax
+
+Now properties can be set on object literals without by including the name of the property without the typical colon and value if the object property the same as the local variable you intend to set as the value of that property. This helps reduce duplication.
```
yawboakye: Not sure we need `Now` to begin the sentence.

yawboakye: Also, see if you can provide example of how it's done currently:
```
function someFunc(thing1, thing2) {
  return {
    thing1: thing1,
    thing2: thing2
  };
}
```
then move on to describe how the new syntax makes it simpler. Just so the learner doesn't have to conjure things up in their mind. Also remind them that it doesn't apply in cases where object key isn't equal to value. And that both old and new style can be mixed.

source/getting-started/js-primer.md
```diff
@@ -52,3 +52,50 @@ export default Ember.Component.extend({
   }
 });
 ```
+
+## Object Literal Shorthand
+
+In ES6 some functionality was added that allows the use of terser syntax for object literal properties and methods, along with some enhanced functionality for property names.
```
yawboakye: What do you think of "ES6 introduced a new syntax for object literals ..." as the opening line. I'm not too comfortable with `some functionality was added`. If that's just me feel free to ignore this comment.

source/getting-started/js-primer.md
```diff
@@ -55,31 +55,67 @@ export default Ember.Component.extend({
...
+function someFunc(thing1, thing2) {
+  return {
+    thing1: thing1,
+    thing2: thing2
+  };
+}
+```
 
-For example:
+Can be written like this in ES6 using the property initializer syntax:
```
yawboakye: Nice! @jaredgalanis what do you think about saying `can be written, in ES6, as:`? I believe the learner wouldn't have lost the context (the new property initializer syntax) by now and so we don't have to remind them.

source/getting-started/js-primer.md
```diff
@@ -55,31 +55,67 @@ export default Ember.Component.extend({
 
 ## Object Literal Shorthand
 
-In ES6 some functionality was added that allows the use of terser syntax for object literal properties and methods, along with some enhanced functionality for property names.
+ES6 introduced a new more succinct syntax for object literal properties and methods, along with some enhanced functionality for property names.
+
+### Property Initializer Syntax
 
-### Property intializer syntax
+Properties can now be set on object literals by including the name of the property only without the typical colon and value. This prevents some duplication when the object property is the same as the local variable you intend to set as the value of that property.
```
yawboakye: I think we can expand this a little further. ES6's initializer syntax works if and only if the property name is the same as the variable holding the value to be assigned to it. For example, while this works
```javascript
let firstName = 'Ember';
let lastName = 'Tomster';

let tomster = {
  firstName,
  lastName
}
```
this won't
```javascript
let nombre = 'Ember';
let apellido = 'Tomster';

let tomster = {
  firstName,
  lastName
}
```
because the variables holding the values are different than the object's properties.

yawboakye:
@jaredgalanis Another thing but I don't know how hard that will be. Is it possible to replace `someObject`, `thing1`, `thing2`, `someFunc` with alternatives that are almost real? For example,
replace `someObject` with `Document`, and the properties and functions with `author`, `publisher`, `publicatoinDate`, `abstract()`, etc?

jaredgalanis:
Thanks @yawboakye. I tried revising things in the way you suggested. Note, that I also took out the final bit about computed property names. My thinking is that subject is too far down in the weeds for what is meant to be covered here. Especially since it's only new to ES6 in the sense that they can be used with object literals (vs instances).

jaredgalanis:
Note that I also took out the final section about computed property names. My thinking is that subject is too far down in the weeds for what is meant to be covered here. Especially since it's only new to ES6 in the sense that they can be used with object literals (vs instances).

yawboakye:
LGTM. @locks?

PR #2069, opened 2017-09-23
Es primer destructuring

jaredgalanis:
@locks @rtablada

### DIFF
--- a/data/spelling-exceptions.txt
+++ b/data/spelling-exceptions.txt
@@ -91,12 +91,10 @@ eslint
 evented
 facto
 falsy
-familyMembers
 fastboot
 filename
 findget
 firefox
-firstName
 fixesurgent
 foo
 formaction
@@ -151,7 +149,6 @@ jsonapiserializer
 jsonserializer
 keycode
 laravel
-lastName
 li
 lifecycle
 linters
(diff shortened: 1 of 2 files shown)


data/spelling-exceptions.txt
```diff
@@ -89,10 +91,12 @@ eslint
 evented
 facto
 falsy
+familyMembers
```
locks: this shouldn't be an exception

data/spelling-exceptions.txt
```diff
@@ -89,10 +91,12 @@ eslint
...
 fastboot
 filename
 findget
 firefox
+firstName
```
locks: same

data/spelling-exceptions.txt
```diff
@@ -146,6 +151,7 @@ jsonapiserializer
 jsonserializer
 keycode
 laravel
+lastName
```
locks: same

source/getting-started/js-primer.md
```diff
@@ -355,3 +357,120 @@ let person = {
...
+  lastName: "Dale"
+};
+
+let firstName = "Yehuda";
+let lastName = "Katz";
+
+( {firstName, lastName} = person );
+```
+
+Note the parenthesis around the curly braces surrounding the assignment statement and the lack of the variable declaration `let`. This denotes the destructuring is an expression instead of a declaration. Because such destructuring assignments are expressions they can be used any where a value is expected, such as in passing a value to a function.
```
locks: I feel like this example is too confusing, even with the explanation

PR #2070, opened 2017-09-24
Adds modules

jaredgalanis:
@locks @rtablada


source/getting-started/js-primer.md
```diff
@@ -355,3 +355,85 @@ let person = {
...
+
+## Modules
+
+Prior to ES6 JavaScript shared one global scope. ES6 modules allow scope to be broken up and managed. With modules variables do not exist above the top-level scope of the module, and any variable you wish to make available outside a module must be exported. Similarly any variable you wish to include in a module that is external to that module must be imported.
+
+As you become familiar Ember you will notice that many files included in an Ember application use the ES6 `import` and `export` keywords to expose functionality of modules across the application. These import and export statements are automatically generated for you if you use `ember-cli` to create your Ember application or files. Most Ember applications do not need alter these automatically generated import and export statements, but it can be helpful to have a basic understanding of how `import` and `export` work in ES6. Additionally, ES6 importing and exporting is necessary if you want to include additional functionality through the use of Ember classes, such as the `mixin` class.
```
yawboakye: @jaredgalanis Consider cutting down on the Ember knowledge here. What do you think of:

Ember application's code is organized as modules, which means that you can use the `import` and `export` keywords to require and expose functionalities in other parts your application. You usually don't have to do extra work converting your code to modules since Ember CLI does that for you. Files generated with the Ember CLI are already modules. That said, it is still helpful to have a basic understanding of how `import` and `export` work in ES6 so let's take a look at them.

jaredgalanis: @yawboakye sounds good to me. I'll try to work it in.

jaredgalanis:
@yawboakye, I updated to include your shortened intro, but made a few revisions per @rtablada's points of focus at the beginning of https://github.com/emberjs/guides/pull/2090.

yawboakye:
@jaredgalanis :+1: LGTM. Should be good to merge after resolving the conflict cc/ @locks

PR #2072, opened 2017-09-28
Recommend users to use `assert.expect()`

migbar:
It might be a good idea to suggest using `assert.expect()` while testing that external action test doubles.


locks:
Thanks @migbar!

PR #2073, opened 2017-09-29
Improve documentation on loading and error event handling

Serabe:
Improve documentation on how handle and use the default behaviour for
each action.

Addresses emberjs/ember.js#15682


source/routing/loading-and-error-substates.md
```diff
@@ -205,3 +208,6 @@ export default Ember.Route.extend({
 
 Analogous to the `loading` event, you could manage the `error` event
 at the application level to avoid writing the same code for multiple routes.
+
+In case we want to run some custom logic and have the default behaviour of rendering the error template,
+we can handle the `error` event and let it bubble by returning `true`.
```
locks: same as above

source/routing/loading-and-error-substates.md
```diff
@@ -128,6 +128,9 @@ export default Ember.Route.extend({
 });
 ```
 
+In case we want both custom logic and the default behaviour for the loading substate,
+we can implement the `loading` action and let it bubble by returning `true`.
```
locks: add code sample please 🙏

Serabe:
@locks done!

Issue #1838, opened 2017-03-07
Error routes don't need to be added to router.js


jonblack:
This is a follow-up to https://github.com/emberjs/ember.js/issues/5620.

The [error substate documentation](https://guides.emberjs.com/v2.11.0/routing/loading-and-error-substates/#toc_code-error-code-substates) describes how error routes are matched. If you create an error route according to these rules (e.g. `ember g route error`) the error route fails with the message:

    More context objects were passed than there are dynamic segments for the route: error

The solution is to remove the generated line from `app/router.js`:

```
Router.map(function() {
  this.route('error'); // <--- this should be removed
});
```

It would be great if the documentation could add this useful tidbit, and perhaps also explain why it's not needed.

locks:
I was trying to find a related issue, and here it is: https://github.com/emberjs/ember.js/issues/14650#issuecomment-283387314. The error substate is different from having a route named `error`. It wasn't possible due to a bug in the router library, but it has been fixed.
That is why in the route map examples present in that page you see neither a `loading`, nor an `error` route.

jonblack:
What's misleading (particularly for beginners) is the following line:

>Ember will look for an error template **or route** in the following order

I've always considered a route to be something that extends `Ember.Route` *and* has an entry in `router.js`. It's interesting that you point out that the examples on that page don't actually show the error route in the route map, something which I didn't notice before.

How about something like this:

>The error substate differs from having a route named error: it doesn't need an entry in the route map. If you generate your error route using ember-cli (e.g. `ember g route error`), make sure you remove the changes made to `router.js`.

I suppose the same sort of text is required for the loading substate as well?

locks:
As you mentioned, this will apply to both `error` and `loading`, so I think the best approach is to explain this at the top of the document, before we dive into the individual substates.

Are you interested in opening a PR to address that? We can discuss details during the PR review process :)

skitterm:
Yikes, this bit me hard, had no idea this was the reason based off the error message. Should really be explicit in the docs that you should **not** have an error entry in `router.js`.

rwjblue:
Well, the real issue is that the manually added `error` route doesn’t include a dynamic segment. I totally agree that we should handle this situation much better. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #2075, opened 2017-09-30
Add note about states being defined for you

NullVoxPopuli:
based off of this comment: https://github.com/emberjs/guides/issues/1838#issuecomment-287716791

/cc @locks

### DIFF
--- a/source/routing/loading-and-error-substates.md
+++ b/source/routing/loading-and-error-substates.md
@@ -2,7 +2,8 @@ The Ember Router allows you to provide feedback that a route is loading, as well
 as when an error occurs in loading a route.
 
 The routes for `error` and `loading` are implicitly defined for you, so they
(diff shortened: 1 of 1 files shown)


source/routing/loading-and-error-substates.md
```diff
@@ -1,6 +1,9 @@
 The Ember Router allows you to provide feedback that a route is loading, as well
 as when an error occurs in loading a route.
 
+The routes for `error` and `loading` are implicitly defined for you, so they
```
locks: Can you elaborate a bit more? The crux is that they are sub-states of a route, not exactly routes themselves, that is why they won't show in the router map.

NullVoxPopuli: updated!

locks:
:shipit:

PR #2076, opened 2017-09-30
Dockerize

NullVoxPopuli:
Motivation:
Dependency management is not fun. Let docker explicitly define what it takes to get this project running.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -244,4 +244,4 @@ DEPENDENCIES
   wdm (>= 0.1.0)
 
 BUNDLED WITH
-   1.15.1
+   1.15.4
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,11 @@
+version: '2'
+services:
+  guides:
+    build:
+      context: .
+      dockerfile: ./Dockerfile
+    command: bundle exec middleman
+    ports:
+      - "4567:4567"
+    volumes:
+      - .:/guides
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM ruby:2.3
+
+USER root
+
+ENV BUNDLER_VERSION 1.15.1
+
+# en - English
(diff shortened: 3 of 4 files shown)


aureliosaraiva:
I created a Docker in my project, it is fork the guides.

https://github.com/ember-brasil/website

jenweber:
**Some notes for people helping with testing:**

From the guides repo, `git remote add nullvoxpopuli https://github.com/NullVoxPopuli/guides.git`

`git checkout add-docker-for-dependency-less-dev`. 

Install Docker from https://www.docker.com/

`docker-compose build && docker-compose up`

View at http://localhost:4567/

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: '3'
```
aureliosaraiva: I recommend to use version 2 the docker-compose. It improve the compatibility with various version the docker-compose. Version 2 is compatible with 1.9+.

Gemfile.lock
```diff
@@ -244,4 +244,4 @@ DEPENDENCIES
   wdm (>= 0.1.0)
 
 BUNDLED WITH
-   1.15.1
+   1.15.4
```
aureliosaraiva: You can defined the version bundler, thus it not break the current environment of other developers who do not use docker.

`RUN gem install bundler -v 1.15.1`

https://github.com/ember-brasil/website/blob/master/docker-images/ember-guides#L19

NullVoxPopuli: updated!

jenweber:
@NullVoxPopuli Can you squash your commits into one?

I'll do README updates in a separate PR

jenweber:
@locks pending a squash of commits, any objections to merging this? I wanted it to be tested on some other machines, but it's a huge improvement even if it doesn't work for everyone.

locks:
@jenweber seems fine, and we can even squash via GitHub, too.

PR #2080, opened 2017-10-09
Introduce JavaScript Primer with variable declarations section

locks:


source/getting-started/js-primer.md
```diff
@@ -0,0 +1,154 @@
...
+
+```javascript
+for (let i = 0; i < 3; i++) {
+  console.log(i) // => wil print 0, 1, 2, and 3 in succession
+}
+
+console.log(i) // => i is not defined
+```
+
+Using `let` will avoid accidentally leaking and changing the `i` variable from outside of the `for` block.
```
yawboakye: @locks I think we could add a stronger recommendation to prefer `let` and `const` over `var`

locks: We need to edit down the guide to do that :P

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+
+JavaScript initially had two ways to declare variables, globally and `var`.
+With the release of ES2015, `const` and `let` were introduced.
+We will go through the different ways to declare a variable,
+also called bindings because they *bind* a value to a variable name,
+and why modern JavaScript tends to prefer `const` and `let`.
+
+### `var`
+
+`var` declarations exist in the entire body of the function where they are declared.
```
rtablada: > Variables declared using `var` exist in the entire body of the function where they are declared.

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+If you try to access a `var` outside of the function it is declared,
+you will get an error that the `var` is not defined.
```
rtablada: > you will get an error that the variable is not defined

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+This is called function-scoping, the existence of the `var` is scoped to the function.
```
rtablada: Move this sentence after the first sentence.

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+
+For our example, we will declare a `var` named `name`.
+We will try to access it both inside the function and outside,
+and see the results we get:
+
+```javascript
+console.log(name); // => name is not defined
```
rtablada: Remove the `=>` unless we use that universally to show printed values. I think along with arrow functions this could be confused.

Also the print out should be 

```
ReferenceError: name is not defined
```

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+
+function myFunction() {
+  var name = "Tomster";
+
+  console.log(name); // => Tomster
```
rtablada: Remove `=>` and add double quotes around the value to mach console log behavior.

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+especially if you are newer to the ecosystem and less familiar with JavaScript.
+
+On top of that, many new features were introduced to JavaScript with the release of newer specifications like ECMAScript 2015,
+also known as [ECMAScript 6 or ES6](https://developer.mozilla.org/en/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla).
+With the current yearly cadence of ECMAScript specifications,
+developers might have not had time to get familiar with all the new features.
+
+In this guide we will be covering some common JavaScript code patterns that appear in Ember applications,
+and in particular those that explore the new features provided in ES6,
+so you can get a clearer sense of where the language ends and the framework starts.
```
rtablada: I'm not a huge fan of this sentence. It's rambling.

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
+While the Guides [assume you have a working knowledge of JavaScript](/#toc_assumptions),
+when learning a new framework it can be hard to know what is JavaScript and what is framework,
+especially if you are newer to the ecosystem and less familiar with JavaScript.
```
rtablada: > especially if you are newer to the ecosystem and less familiar with JavaScript

I worry that this language could be alienating.

source/getting-started/js-primer.md
```diff
@@ -0,0 +1,161 @@
...
+
+On top of that, many new features were introduced to JavaScript with the release of newer specifications like ECMAScript 2015,
```
rtablada: Remove ~On the top of that~

PR #2081, opened 2017-10-09
Add missing postData property to blog-post component

hybridmuse:
From right above the code example: 
> When {{post.body}} is instantiated, it will have both the editStyle and postData given by its wrapping component [...]


locks:
Appreciated, thanks :)

PR #2086, opened 2017-10-15
Redirect URLs do not contain base path.

axelerate:
Added complete path for defining-components link.


axelerate:
I'm not sure if the current state is the correct one, as following the link will redirect to a 404, given that the full path isn't provided.

PR #2087, opened 2017-10-15
Use Public Website Docker Image

NullVoxPopuli:
This should have no impact on contributors. The command to start up the container is still `docker-compose up`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,9 @@
-FROM ruby:2.3
+FROM dockerizedember/website-dev:latest
 
-USER root
-
-ENV BUNDLER_VERSION 1.15.1
-
-# en - English
-# pt - Portuguese and Brazilian Portuguese
-# es - Spanish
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
:+1:

NullVoxPopuli:
I suppose the C.I. stuff could be changed to use docker, too

acorncom:
Just tested this out, works great, if only we'd had this when I got started on guides ... 😀  Thx @NullVoxPopuli 👍 

One question. If/when we want to update https://hub.docker.com/r/dockerizedember/website-dev/, what is the process for doing so?  Assume (rightly 😉) that you are explaining to a Docker newb ...

NullVoxPopuli:
All we do push a commit to the corresponding github repo and dockerfile will rebuild.

Locally you add the --pull flag to docker compose because we aren't using version tags on this image

PR #2088, opened 2017-10-17
Fix incorrect reference to `Ember` in updated example code

jscn:
Commit 9f5d402 changed a code snippet, but left a reference in the text to the previous code in place. This change updates the text so that it reflects the change in the code.


locks:
Appreciate the keen eye :)

PR #2090, opened 2017-10-19
WIP: JS Primer Var - Edit 1

rtablada:
This is the first pass at editing some of the language and definitions for the JS primer.

Some things I've focused on:

* Passive Voice
* 1st vs 2nd person voice (I've preferred having no `you` or `we` where possible when talking about technical specs)
* Terminology and definitions

### DIFF
--- a/source/getting-started/js-primer.md
+++ b/source/getting-started/js-primer.md
@@ -1,74 +1,99 @@
-Many new features were introduced to JavaScript with the release of newer specifications like ECMAScript 2015,
-also known as [ECMAScript 6 or ES6](https://developer.mozilla.org/en/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla).
+The ECMAScript 2015 specification (also known as ES2015 or ES6) introduced new features to the JavaScript language.
 While the Guides [assume you have a working knowledge of JavaScript](/#toc_assumptions),
 not every feature of the JavaScript language may be familiar to the developer.
 
-In this guide we will be covering some JavaScript features,
-and how they are used in Ember applications.
+This guide will cover some JavaScript features, and how to use them in Ember applications.
+To see a full set of features added in ES2015, check the [Mozilla Developer Network Docs](https://developer.mozilla.org/en/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla).
 
 ## Variable declarations
 
-A variable declaration, also called binding, is when you assign a value to a variable name.
-An example of declaring a variable containing the number 42 is like so:
+A variable declaration is when you assign a variable name to a location in memory.
+This creates a place in memory to store data and allows lookup for that data by name.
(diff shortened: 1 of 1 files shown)


source/getting-started/js-primer.md
```diff
@@ -1,74 +1,99 @@
...
+Variables declared using the `var` keyword only exist within that function.
+This is called *function-scoping*.
+Trying to access a variable declared with `var` outside of its function will throw an error.
+This is because the variable is not declared in the current scope.
 
-For our example, we will declare a `var` named `name`.
-We will try to access it both inside the function and outside,
-and see the results we get:
+The following code declares a variable using  `var` inside of a function.
+Then, the variable is logged both inside and outside of that function:
```
rtablada: I'm a bit 🤔 about this sentence, I think there should be some mention that it "tries" to log the variable since one of the instances will lead to an error

locks: Maybe split into two sentences? It logs inside the function. It tries to log outside but triggers an error.

source/getting-started/js-primer.md
```diff
@@ -1,74 +1,99 @@
...
+ console.log(name); // "Tomster"
 }
+
+console.log(name); // "Tomster"
 ```
 
-This also means that if you have an `if` or a `for` in your code and declare a `var` inside them,
-you can still access the variable outside of those blocks:
+Another odd behavior of the `var` keyword is something called *hoisting*.
+Javascript sets aside memory for variables when it enters a function.
```
locks: JavaScript

source/getting-started/js-primer.md
```diff
@@ -1,74 +1,99 @@
...
+  return name;
 }
 ```
 
-In the previous example, we can see that the first `console.log(name)` prints out `undefined` instead of the value.
-That is because of a feature of JavaScript called *hosting*.
-Any variable declaration is moved by the programming language to the top of the scope it belongs to.
-As we saw at the beginning, `var` is scoped to the function,
-so the previous example is the same as:
+To reduce hoisting confusion it is best practice to declare variables using `var` at the top of functions.
```
locks: "hoisting confusion" sounds strange

locks: p.s. new band name

source/getting-started/js-primer.md
```diff
@@ -1,74 +1,99 @@
...
-you will get an error that the variable is not defined.
+Variables declared using the `var` keyword only exist within that function.
+This is called *function-scoping*.
+Trying to access a variable declared with `var` outside of its function will throw an error.
+This is because the variable is not declared in the current scope.
 
-For our example, we will declare a `var` named `name`.
-We will try to access it both inside the function and outside,
-and see the results we get:
+The following code declares a variable using  `var` inside of a function.
```
knownasilya: Extra space before `var`

rtablada:
Closing to re-edit based on new source.

PR #2094, opened 2017-10-25
Values logged within from within the loop go to 2

rstellar:
The loop increments and the final console.log will be the only one that shows a 3.
I could be misinterpreting what the comment on line 132 was hoping to communicate. 
https://repl.it/NO8W
This is the output for my Repl.it example:


locks:
Thanks!

Issue #2095, opened 2017-10-26
Services and Utilities Guide Compiles with Errors


jciskey:
When working through the Services and Utilities guide in the Ember tutorial, the code provided for app/services/maps.js produces compilation errors:

> /home/ubuntu/super-rentals/app/services/maps.js
>   18:29  error  'Ember' is not defined  no-undef
> 
> ✖ 1 problem (1 error, 0 warnings)

This originates with the line:

`    let camelizedLocation = Ember.String.camelize(location);`

PR #2096, opened 2017-10-26
Add Strings namespace import (#2095)

jciskey:
As-is, the provided sample code compiles with an error because the Ember global object is not defined anywhere.

This update adds an import of the Strings namespace and adjusts the usage from "Ember.Strings.camelize" to just "camelize".

Fixes: #2095


locks:
Appreciated!

PR #2098, opened 2017-10-27
Fix app/routes/rentals.js diff

pepke41:
I was going through the tutorial and on the https://guides.emberjs.com/v2.16.0/tutorial/model-hook/ the app/routes/rentals.js diff made it look like I should delete the following code:
```.js
export default Route.extend({
  model() {
```
That didn't seem right to me so I consulted a friend who knows ember well, and he told me that those two lines should not be deleted. 

My PR removes the lines from the diff that already existed in the code `export default Route.extend({`, and changed the '-' on `model()` to a '+'. That way in the future the changes will be more clear for anyone who is going through the Ember tutorial.


pepke41:
I saw that this was modified, maybe as a try to fix it, but I still think its a little confusing so I'm going to keep this PR open.

locks:
Hi @pepke41, sorry for the delay. Can you update your PR? Seems like there are still some conflicts.

jenweber:
@pepke41 we are going through old PRs and cleaning them up this week. Sorry this never got merged in back in the day, but the help is much appreciated :) There are lots of little things like this that we definitely need help catching.

Congrats on your first merged PR for The Guides! 🎉  Thanks a bunch!

PR #2099, opened 2017-10-27
Update model-hook.md

sakshyamshah:
there seems to be unintended (-)operator in line 3 and 4 of code snippet which instructs to remove portion of code below.
```
export default Route.extend({
  model() {
```

This breaks and throws build error. I have removed (-) from line 3 and 4.


locks:
Appreciated! :)

PR #2105, opened 2017-10-30
Use single quotes to avoid early termination of string

ksolo:
Working through the tutorials and found this.


locks:
Hey @ksolo, what happened here?

ksolo:
@locks It looks like the double quote is fine in the handlebar syntax. My editor flagged it at first but the build worked just fine. Turns out it's not an issue.

PR #2107, opened 2017-10-31

staceyastewart:
This change adds the link to the `transitionTo` function so that someone walking through the tutorial can see an example of how `transitionTo` is used. Since there is a link to `replaceWith` in this paragraph, I think it is useful to link to `transitionTo` as well.


locks:
Thank you, this is helpful!

PR #2109, opened 2017-10-31
Adds output when installing styling

staceyastewart:
Shows the terminal output when `ember install ember-cli-tutorial-style` is successful.


locks:
I wonder if we should sneak in a gif here. Thanks for the contribution @staceyastewart!

PR #2110, opened 2017-10-31
Update hbs-helper.md

staceyastewart:
Adding line 17 as a + is causing the lines from 17 to the end to highlight in green since the line is blank. However, lines 20 and 21 should not be highlighted as they are not additions to the file.


locks:
Thank you!

locks:
Thank you!

PR #2111, opened 2017-10-31
Add lines to model

staceyastewart:
This PR adds line add distinction to the info being added to `app/models/rental.js` to ensure styling continuity.


locks:
Thanks!

PR #2112, opened 2017-11-01
Fix tutorial example changed lines.

mpeteffi:
Fix list-filter's component tutorial example changed lines:
    - show removed code from init method.
    - add line 24 as changed on handleFilterEntry method.


toddjordan:
Thanks! Really appreciate you taking the time to fix.  It seems to have already recently been fixed as part of a separate update.
image

PR #2114, opened 2017-11-05
Fix JS code comment for `for` loop behavior

adamfeldman:
Both loops only count up to `2`; the code comment for the loop using `let` incorrectly described it as behaving differently from `var` and counting up to `3`.


locks:
Thanks!

PR #2115, opened 2017-11-06
htmlsafe imports and links

asizer:
`htmlSafe` is found in the `@ember/string` module, not the `@ember/component` module. Additionally, the link to the htmlSafe method was broken.

### DIFF
--- a/source/templates/writing-helpers.md
+++ b/source/templates/writing-helpers.md
@@ -374,10 +374,11 @@ escape anything that may have come from an untrusted user with the
 ```app/helpers/make-bold.js
 import { helper } from "@ember/component/helper";
 import Handlebars from "handlebars";
+import { htmlSafe } from "@ember/string";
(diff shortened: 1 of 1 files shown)


locks:
`import { htmlSafe } from '@ember/string';` is deprecated as of https://github.com/ember-cli/ember-rfc176-data/pull/38.

asizer:
ugh, ok, but the current `import { htmlSafe } from '@ember/component'` doesn't exist and breaks the build at 2.16.2. Additionally, the `htmlSafe` link is broken in the docs right now, but the class does exist under the `@ember/string` module, so seems helpful to have a link to somewhere.

Took out the deprecated import statements and changed htmlSafe reference to `Ember.String.htmlSafe`, as the current import statement doesn't work.

locks:
Thanks for your contribution :)

PR #2118, opened 2017-11-07
Updates import line

staceyastewart:
This PR updates the import line on the list-filter.js component file and rentals.js controller file.
It also removes the unused Ember variable. from both files.


locks:
Thank you!

PR #2119, opened 2017-11-07
Adds a line that must be removed from the list-filter.js component file

staceyastewart:
This PR adds a line into the list-filter.js file that must be removed and replaced while walking through this part of the guides.
Also adds the add highlighting on a line that is added.


yawboakye:
@staceyastewart Thanks :+1:! Should be good to go after you resolve the conflict

PR #2121, opened 2017-11-07
use an arguably "cleaner" pattern for `reopenClass`

kellyselden:
I didn't know `reopenClass` was "chainable" until I saw https://github.com/ef4/ember-set-body-class/blob/511278f87e9bb6f62b7b030d72a23a85249f013d/addon/components/set-body-class.js#L35-L37.

This pattern doesn't make you come up with a component alias that isn't really used.

cc @ef4


ef4:
I guess it's subjective. To me, having one less name in scope makes it easier to read.

(What we really need is static class properties, which will hopefully make everybody happy.)

rwjblue:
Ya, agreed.

yawboakye:
@rwjblue I also agree that it's easier to read.

@locks @toddjordan looks good to merge.

PR #2123, opened 2017-11-09
Update screenshot for ember-cli-tutorial-style step

williamhaley:
Minor detail, but I just re-ran through the tutorial today and noticed the screenshot taken after applying `ember-cli-tutorial-style` is a little out of date.

Tomster should be a bit bigger, listings should align with other content, and some other minor tweaks seems to have been put in place since the last screenshot was taken.


locks:
Thank you, all the details matter!

PR #2127, opened 2017-11-10
Fixed broken urls to serializer API docs

david-duncan:
I'm guessing there might be more of these that a global find could produce - I just fixed the ones I found on this page


locks:
Appreciated!

PR #2132, opened 2017-11-17
Where to add attribute bindings

pablobm:
Recently, on the community Slack, somebody was asking about this. They mentioned that this page showed the code, but not where it went. I agreed, so I decided to create this small PR.

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -40,46 +40,37 @@ If `isAdministrator` is `false`, Handlebars will produce the following:
 
 ### Adding Other Attributes (Including Data Attributes)
 
-By default, helpers and components only accept a limited number of HTML attributes.
+By default, components only accept a limited number of HTML attributes.
 This means that some uncommon but perfectly valid attributes, such as `lang` or
 custom `data-*` attributes must be specifically enabled. For example, this template:
 
 ```handlebars
 {{#link-to "photos" data-toggle="dropdown" lang="es"}}Fotos{{/link-to}}
-
-{{input type="text" data-toggle="tooltip" data-placement="bottom" title="Name"}}
 ```
 
-renders the following HTML:
+Will render the following HTML:
 
 ```html
 <a id="ember239" class="ember-view" href="#/photos">Fotos</a>
-
-<input id="ember257" class="ember-view ember-text-field" type="text"
-       title="Name">
 ```
 
-To enable support for these attributes an attribute binding must be
-added to the component, e.g.
+To enable support for these attributes, an attribute binding must be
+added for each specific attribute on the component.
+To do this, you can extend the appropriate components
+in your application. For example, for `link-to` you would create your own version
+of this component by extending
 [`Ember.LinkComponent`](https://www.emberjs.com/api/ember/release/classes/LinkComponent)
-or [`Ember.TextField`](https://www.emberjs.com/api/ember/release/classes/TextField)
-for the specific attribute:
 
-```javascript
-Ember.LinkComponent.reopen({
+```app/components/link-to/component.js
+import LinkComponent from '@ember/routing/link-component';
+
+export default LinkComponent.extend({
   attributeBindings: ['data-toggle', 'lang']
-});
-
-Ember.TextField.reopen({
-  attributeBindings: ['data-toggle', 'data-placement']
 });
 ```
 
 Now the same template above renders the following HTML:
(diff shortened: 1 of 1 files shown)


sivakumar-kailasam:
@pablobm Thanks for taking the time to raise the PR. While `reopen` does have its own section in the guides, we would like to suggest avoiding it when we can. We will have to amend the guides to reflect our thoughts about it.

In this use case we can avoid `reopen` and instead add a component named `link-to` in our app & use `Ember.LinkComponent.extend({})` to achieve the desired outcome. Is that something you can modify your PR with?

pablobm:
@sivakumar-kailasam Makes sense. Will update.

pablobm:
@sivakumar-kailasam Updated now (I thought it would take me longer).

I see a new question now: how do people know that the correct component is `-text-field` and not `text-field`? (Apart from reading this, that is). I had to had a look at the source code to find out myself.

sivakumar-kailasam:
It might be best to have the sample only talk about `link-to` to show how this could be done. It would be hard to account for all in-built components/helpers as a part of the official guides. 

//cc @locks thoughts?

yawboakye:
@pablobm @sivakumar-kailasam The statement at the start of the segment:
```By default, helpers and components do not accept *data attributes*```
only holds true for Ember's own components and helpers since they're already defined and require extension to add both attributes and behavior. Would be nice if you could allude to this. I also agree with @sivakumar-kailasam recommendation to stick to one (and the easier) example, `link-to`. It gets slightly murky when you talk about the `input` helper.

**Comment by pablobm** [comment]
@sivakumar-kailasam I agree that mentioning `-text-field` here could complicate things too much. I have removed it. I have also rephrased the paragraph slightly, to something I think may be clearer.

**Comment by pablobm** [comment]
@yawboakye You are right, but that's not the whole story. I have been playing with this a bit this morning and I see this happening with more attributes that just `data-*`. Each helper/component has a set list of attributes they will recognise, and any others will be ignored. For example, if you create a new component yourself and pass it a `lang` attribute, it will be ignored by default even though HTML5 accepts it for any element.

I agree that this paragraph could be rephrased, but I'm thinking of doing it on a different PR, as the concern addressed on this one is separate enough.

File: source/templates/binding-element-attributes.md [hunk 1]
```diff
@@ -44,40 +44,30 @@ By default, helpers and components do not accept *data attributes*. For example
...
+added for the specific attribute on the component.
+To do this, you can extend the appropriate components
+in your application. For example, for `link-to` you would create your own version
+of this component by extending [`LinkComponent`](https://www.emberjs.com/api/ember/2.16/classes/LinkComponent):
 
-```javascript
-Ember.LinkComponent.reopen({
-  attributeBindings: ['data-toggle']
-});
+```app/components/link-to/component.js
```
locks: Should be the classic path, not pods.

pablobm: Fixed.

pablobm:
I rebased this branch to allow for the latest changes to the page (mainly from #2149). This rebase also addresses the comments raised above.

Some context of the conversation may have been lost, so I'll refresh your minds.

Initially, this PR aimed to clarify where the `reopen` calls would live in a project's filesystem. So that's mainly extending the example, including adding a reference to the correct file path.

The following concerns were raised by @sivakumar-kailasam and @yawboakye:

1. `reopen` is no longer the best practice for this, and instead we should be creating new components that extend and override the original ones.
2. Explaining how to extend the `input` helper may complicate this section too much, as it involves extending the undocumented component `-text-field`. Therefore, maybe that part should be left out.

This feedback has been incorporated into this PR.

locks:
`TextField` is documented in the [API docs](https://emberjs.com/api/ember/release/classes/TextField).

pablobm:
@locks Is this the same that `-text-field` used to be then, at least for the purposes of this page?

In other words: should I re-include the text field example, using `text-field` as the component to override?

toddjordan:
@pablobm would you be willing to post a twiddle that validates that the example code works?

pablobm:
@toddjordan Sure: https://ember-twiddle.com/0feff0577066c1414747aef04202aa9d

The problem with that Twiddle is that it doesn't use the imports from the `@ember` namespace, as I can't get them to work. I'm assuming that's a limitation of the tool; please correct me if I'm wrong.

sivakumar-kailasam:

Issue #2092, opened 2017-10-22
Confusing variable connection in the block form guide


BradenLawrence:
I was working through the Super Rentals app and got confused during the section explaining [block components](https://guides.emberjs.com/v2.16.0/tutorial/autocomplete-component/).  I had a hard time seeing the connection between the variables in the following two files:

`// app/templates/rentals.hbs`
`{{#list-filter filter=(action 'filterByCity') as |rentals|}}`
`...`
`{{/list-filter}}`
and
`// app/templates/components/list-filter.hbs`
`...`
`{{yield results}}`

Here, `as |rentals|` refers to the value of `results`, but I initially thought it referred to the ouput of `action 'filterByCity'`. In turn, the `results` variable appears to never be used. I think this could be improved by changing `as |rentals|` to `as |results|`, and clearly calling out the connection between the two.

If this sounds good to you, I can PR this.

acorncom:
Hey @BradenLawrence, sorry to hear about the mixup.  Yes, a PR for that sounds great. If you wouldn't mind sending a corresponding PR to https://github.com/ember-learn/super-rentals we'll get those merged together to keep things in sync ...

BradenLawrence:
Hey @acorncom,
I've opened two PRs to address this:
Guides: https://github.com/emberjs/guides/pull/2134
Super-rentals: https://github.com/ember-learn/super-rentals/pull/82
Let me know if you have any questions!

acorncom:
Closed by #2134

PR #2134, opened 2017-11-19
Explain how yield affects block parameters (#2092)

BradenLawrence:
Updating the code examples for app/templates/rentals.hbs and
app/templates/components/list-filter.hbs to make the connection
between |rentals| and {{yield results}} more explicit, and adding
a short description explaining how the data moves from our action
to our display.

 * Change |rentals| to |filterResults|. This ties it to `results`
   while still being a unique name for the sake of clarity.

 * Add text explaining their connection.

 * Remove the transition text leading into the example code for
   app/components/list-filter.js because the new text already
   does this.

 * Update all later example code where the |rentals| block parameter
   is used.

Fix typo in block parameters introduction

### DIFF
--- a/source/tutorial/autocomplete-component.md
+++ b/source/tutorial/autocomplete-component.md
@@ -22,7 +22,7 @@ Notice that below we "wrap" our rentals markup inside the open and closing menti
 This is an example of the [**block form**](../../components/wrapping-content-in-a-component) of a component,
 which allows a Handlebars template to be rendered _inside_ the component's template wherever the `{{yield}}` expression appears.
 
-In this case we are passing, or "yielding", our filter data to the inner markup as a variable called `rentals` (line 14).
+In this case we are passing, or "yielding", our filter data to the inner markup as a variable called `filteredResults` (line 14).
 
 ```app/templates/rentals.hbs{+12,+13,+14,+15,+16,+17,+18,+19,+20,-21,-22,-23}
 <div class="jumbo">
@@ -38,9 +38,9 @@ In this case we are passing, or "yielding", our filter data to the inner markup 
 
 {{#list-filter
    filter=(action 'filterByCity')
-   as |rentals|}}
+   as |filteredResults|}}
   <ul class="results">
-    {{#each rentals as |rentalUnit|}}
+    {{#each filteredResults as |rentalUnit|}}
       <li>{{rental-listing rental=rentalUnit}}</li>
     {{/each}}
   </ul>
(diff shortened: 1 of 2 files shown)


source/tutorial/autocomplete-component.md
```diff
@@ -38,9 +38,9 @@ In this case we are passing, or "yielding", our filter data to the inner markup
 
 {{#list-filter
    filter=(action 'filterByCity')
-   as |rentals|}}
+   as |filterResults|}}
```
locks: what about `filteredRentals`?

BradenLawrence: Sure, that works. I was thinking of it as the "results of the filter", but I can see how that might be mistaken for an action of some kind.

source/tutorial/autocomplete-component.md
```diff
@@ -71,7 +71,7 @@ the new value of the property is present in both the rendered web page and in th
 
 The `key-up` property will be bound to the `handleFilterEntry` action.
 
-Here is what the component's JavaScript looks like:
+The `handleFilterEntry` action will assign `results` with the list of rentals it returns. The `{{yield}}` helper will then use `results` to fill in the value of our block parameter, `|filteredResults|`. We still need to define how `handleFilterEntry` gets those results though, so let's do that now:
```
yawboakye: @BradenLawrence Nice work! I'm not sure I understand what you wrote here though. I'd be happy if we can work to get this PR merged in as early as possible since the corresponding change on Super Rentals is already live.

BradenLawrence: Hi @yawboakye, my goal here is to explain how block form components and the {{yield}} tag interact. 

The original text does a good job explaining how code we put inside the {{myComponent}} ... {{/myComponent}} block goes to the {{yield}} tag, but it doesn't explain how code we put in {{yield myVariable}} is sent back to the parent. I was confused by this when I first went through the guides because I couldn't figure out where the data in "as |rentals|" was coming from.

I tried to address this by making the variable names match up: "{{yield results}}" feeds into "as |filteredResults|", and walking through it step by step on line 74. Is there a better way I could word this? Let me know if you have any questions!

jenweber: Maybe something like this?
```
The `handleFilterEntry` action will apply the search term filter to the list of rentals, and set a component attribute called `results`. The `results` are passed to the `{{yield}}` helper in the template. In the yielded block component, those same `results` are referred to as `|filteredResults|`. Let's apply the filter to our rentals:
```

jenweber:
@BradenLawrence with a couple more edits, I think this will be ready to merge. Since it's already live in super-rentals, I think we could possibly merge the next iteration in, and do any subsequent edits in a future PR

source/tutorial/autocomplete-component.md
BradenLawrence: Looking back on this, I see I misread this as `filteredResults` instead of `filteredRentals`. Since this was already pulled in to super-rentals like this, would you be alright with keeping it as `filteredResults`? If not, I'll open a new PR to fix it there.

BradenLawrence:
@jenweber those changes make sense to me. I've included them in my most recent push. Let me know if we need any other changes to wrap this up.

PR #2137, opened 2017-11-28
Fixes deep link for Component#event-names

alvincrespo:
As I was trying to figure out an issue with events not triggering in the app I was working on, I looked up the input helper in the docs. After clicking on `Event Names` I get taken to:

Took some time to quickly patch this up.

## Changes

- Updates link to event names in input-helpers.md

Cheers!

### DIFF
--- a/source/templates/input-helpers.md
+++ b/source/templates/input-helpers.md
@@ -51,7 +51,7 @@ To dispatch an action on specific events, such as `enter` or `key-press`, use th
 {{input value=firstName key-press="updateFirstName"}}
 ```
 
-[Event Names](https://www.emberjs.com/api/ember/2.16/classes/Component#toc_event-names) must be dasherized.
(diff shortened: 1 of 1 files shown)


jenweber:
Hi Alvin, thanks for this PR! When I follow the new link, I'm seeing the same result. What do you see? 

I think our routing and proxying for this app may not allow for bookmark urls. @sivakumar-kailasam do you know anything about it?

toddjordan:
doing those internal hash links seems to be broken in the app atm.  it seems our legacy hash conversion code messes with it.  Will probably take some debugging to figure out a fix.

shotgundebugging:
@toddjordan, could you please tell me where the code which should generate these anchored links can be found?

locks:
Let do non-anchor link for now. I changed it from hard-coded to `/release`.

PR #2138, opened 2017-11-29
Fix link to Event Names

shotgundebugging:
`Event Names` section from the `Component` API documentation is not anchored and it was resulting in a 404. Let's link to an article having the same content in the guide, instead of the API docs.


acorncom:
Hmm, in this particular case, the place where people are leaving (`input helpers`) describes dasherized event names when working with the `{{input}}` component. However what we would be linking to indicates using camelCased event names (due to being able to hook those events inside of an event). Those seem to be different usages, no?

yawboakye:
@acorncom That is because they're used in different places. We use the dasherized name in the template (e.g. `{{input key-press="actionName"}}` while we use the camelCased name as a method in the component's JavaScript file. If we can make the distinction then it won't be too confusing @shotgundebugging

yawboakye:
@acorncom @shotgundebugging Looks like this is another PR attempting a fix too: https://github.com/emberjs/guides/pull/2137.

shotgundebugging:
Closing this since, as @toddjordan points out in #2137, this requires a more generic fix:

> doing those internal hash links seems to be broken in the app atm. it seems our legacy hash conversion code messes with it. Will probably take some debugging to figure out a fix.

PR #2141, opened 2017-12-04
Remove reference to ArrayController

simonihmig:
AFAIK it has been removed since 2.0, so no need to reference it in the docs anymore.


acorncom:
Thanks @simonihmig

PR #2145, opened 2017-12-18
Testing guide update for new Ember Qunit API

jessica-jordan:
This PR is an update to the Ember testing guide, updating the code examples according to the new Ember QUnit API proposals from [RFC 232](https://github.com/emberjs/rfcs/blob/master/text/0232-simplify-qunit-testing-api.md) and [RFC 268](https://github.com/emberjs/rfcs/blob/master/text/0268-acceptance-testing-refactor.md).

This is also updates the naming for different types of tests.

Relates to the quest issue emberjs/ember.js#15933


source/testing/acceptance.md
```diff
@@ -8,39 +8,47 @@ ember g acceptance-test login
...
 });
 ```
 
-`moduleForAcceptance` deals with application setup and teardown. The last few lines, within
-the function `test`, contain an example test.
+`module` allows you to scope your tests: Any test setup that is done inside of this scope, will
+apply to all test cases contained in this module. You can read more about module specific setup and
+teardown in the section [Setup and teardown](Acceptance#setup-and-teardown).
+Scoping your tests with `module` also allows you to execute your tests independently from other tests.
+For example, to only run your tests from your `login` module, run `ember t -m 'Acceptance | login'`.
```
rwjblue: It may be better to use the longer form of the various flags/commands in the docs. Its _probably_ much easier to remember `ember test --module` than `ember t -m`...

source/testing/acceptance.md
```diff
@@ -8,39 +8,47 @@ ember g acceptance-test login
...
+import { setupApplicationTest } from 'ember-qunit';
+
+module('Acceptance | posts', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('should add new post', async function(assert) {
+    await visit('/posts/new');
+    await fillIn('input.title', 'My new post');
+    await click('button.submit');
+    assert.equal(this.element.querySelector('ul.posts li:first').textContent, 'My new post');
```
rwjblue: Just **begs** for `qunit-dom` 😝 .  I'll try to tackle that (or convince @Turbo87 to take a stab) in a future RFC...

source/testing/acceptance.md
```diff
@@ -96,6 +104,30 @@ next one starts.
...
+  ```tests/acceptance/new-post-appears-first-test.js
+  test('should add new post', async function(assert) {
+    await visit('/posts/new');
+    await fillIn('input.title', 'My new post');
+    await click('button.submit');
+    assert.equal(this.element.querySelector('ul.posts li:first').textContent, 'My new post');
+  });
+  ```
+
+`Await`ing the execution of the asynchronous helpers this way, will ensure that assertions are always made **after** the
```
rwjblue: Probably shouldn't capitalize here, though I agree that its a tad odd either way.

source/testing/acceptance.md
```diff
@@ -114,21 +146,26 @@ Synchronous helpers are performed immediately when triggered.
...
+import { setupApplicationTest } from 'ember-qunit';
+
+module('Acceptance | posts', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('should add new post', function(assert) {
+    visit('/posts/new');
+    fillIn('input.title', 'My new post');
+    click('button.submit');
+    andThen(() => assert.equal(this.element.querySelector('ul.posts li:first').textContent, 'My new post'));
```
rwjblue: `andThen` is completely gone in tests that are setup with `setupApplicationTest` (it is only available in "legacy" acceptance tests)

source/testing/unit-testing-basics.md
```diff
@@ -33,25 +33,51 @@ should trigger the computed property), and assert that the logic in our
...
+import { run } from '@ember/runloop';
+
+module('Unit | Model | some thing', function(hooks) {
+  setupTest(hooks);
 
-test('should correctly concat foo', function(assert) {
-  const someThing = this.subject();
-  someThing.set('foo', 'baz');
+  test('should correctly concat foo', function(assert) {
+    const someThing = run(() => this.owner.lookup('service:store').createRecord('some-thing'));
```
rwjblue: The required `run` wrapping makes me sad.  Cross linking https://github.com/emberjs/data/issues/5247 with a discussion around whats going on. We have a path forward, just need a champion to tackle fixing it...

source/testing/unit-testing-basics.md
```diff
@@ -33,25 +33,51 @@ should trigger the computed property), and assert that the logic in our
...
 });
 ```
 
-See that we have used `moduleFor`, one of the several unit-test helpers provided by Ember-Qunit.
-Test helpers provide us with some conveniences, such as the `subject` function that handles lookup and instantiation for our object under test.
-Note that in a unit test you can customize the initialization of your object under test by passing to the
-`subject` function an object containing the instance variables you would like to initialize.  For example, to initialize
-the property 'foo' in our object under test, we would call `this.subject({ foo: 'bar' });`
+See that first, we are creating a new testing module using Qunit's `module` function. This will scope
+all of our tests together in one package that can be configured and run independently from other modules
```
rwjblue: s/one package/one group/

?

source/testing/unit-testing-basics.md
```diff
@@ -33,25 +33,51 @@ should trigger the computed property), and assert that the logic in our
...
+defined in our test suite.
+Also, we have used `setupTest`, one of the several unit-test helpers provided by Ember-Qunit.
```
rwjblue: Should the various references to other packages (ember-qunit in this case) be links?

source/testing/acceptance.md
jessica-jordan: Yes, I definitely agree - let's update the examples accordingly as soon as we use `qunit-dom` by default as well

source/testing/acceptance.md
jessica-jordan: Sweet, I wasn't sure about that yet. I can update / remove that section then to only discuss the `async / await` syntax instead

source/testing/unit-testing-basics.md
jessica-jordan: Yes, I like the wording and can update accordingly :thumbsup:

source/testing/testing-components.md
```diff
@@ -364,29 +420,35 @@ const locationStub = Service.extend({
...
-  assert.equal(this.$().text().trim(), 'You currently are located in New York, USA', 'origin location should display');
-  this.set('locationService.city', 'Beijing');
-  this.set('locationService.country', 'China');
-  this.set('locationService.currentLocation', { x: 11111, y: 222222 });
-  assert.equal(this.$().text().trim(), 'You currently are located in Beijing, China', 'location display should change');
+    assert.equal(this.element.textContent.trim(),
+     'You currently are located in Beijing, China', 'location display should change');
+  });
 });
 ```
```
jessica-jordan: I wonder if it makes sense to skip the `Waiting on Asynchronous Behavior` section completely, as I feel we'd like to encourage the use of `async / await` over custom helpers in the Guides. Any thoughts?

rwjblue:
Just checking in here, how are things going?  Everything still on track for 3.0 (roughly 4 weeks from now...)?

jessica-jordan:
@rwjblue My mistake, that I haven't followed up on this one for a while. I'll just be fixing up some of the links and I'll check if I can another review from someone from the learning team as well

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
-also include interactions such as _login_ and _create post_. Let's finally
-assume that you would like to have [automated tests] in place for your application.
+This application would likely include models such as `user` and `post`.
+It would also include interactions such as _login_ and _create post_.
+Let's finally assume that you would like to have [automated tests] in place for your application.
 
-There are three different classifications of tests that you will need:
-**Acceptance**, **Unit**, and **Integration**.
+There are four different types of test setups you can choose from depending on what you will need to test.
+Let's have a look at those test types and when each of them is the most useful.
```
jenweber: I think it's good to list all 4 types in this intro. It may even be worthwhile to mention that "in the past, these were sometimes referred to as Acceptance, Unit, and Integration tests." I could see this change making it hard for people to look back at old materials and figure out what's going on.

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
-scenarios.
-
-It is recommended that components and helpers be tested with integration tests because the component
-interacts with the system in the same way that it will within the context of the application,
+Rendering tests are especially useful for verifying the correct behavior of components and helpers.
+The component or helper interacts with the system in the same way that it will within the context of the application,
 including being rendered from a template and receiving Ember's lifecycle hooks.
 
-Examples of integration tests are:
+Examples of these mid-complex tests are:
```
jenweber: Is mid-complex a specific term known in generic testing/new Ember testing API? I found this a little confusing, like it was a different type of test.

rwjblue: No, it isn’t a term related to the changes in testing API

source/testing/testing-components.md
jenweber: Assuming that async/await covers the same functionality as `wait`, I say we drop the wait explanation. We may still want to cover something in terms of "handling async." It tripped me up as a beginner.

source/testing/index.md
jenweber: Thanks Robert. Let's avoid that label then.

jenweber:
The only blocker I see on the first pass is the `mid-complex` label. I have some small punctuation/wording changes to recommend. Let me know how you want me to submit them (commit, PR to you @jessica-jordan, or comments).

jessica-jordan:
@jenweber I updated the PR to your comments so far - feel free to also comment on any punctuation / wording suggestions or to add the changes as another commit - whatever is easiest for you

source/testing/acceptance.md
Gaurav0: I'm curious as to why we don't use `find`, `findWithAssert` and `findAll` a la `ember-cli-native-helpers` instead of `qunit-dom`. They could have been put right into `@ember/test-helpers`.

source/testing/testing-models.md
```diff
@@ -28,26 +28,25 @@ export default Model.extend({
...
-  // this.subject aliases the createRecord method on the model
-  const player = this.subject({ level: 4 });
+  test('should increment level when told to', function(assert) {
+    // this.subject aliases the createRecord method on the model
+    const player = run(() => this.owner.lookup('service:store').createRecord('player'));
 
-  // wrap asynchronous call in run loop
-  run(() => player.levelUp());
+    // wrap asynchronous call in run loop
+    run(() => player.levelUp());
```
Gaurav0: Is it possible to `await` this?

data/pages.yml
```diff
@@ -180,10 +180,10 @@
...
     - title: "Introduction"
       url: "index"
-    - title: "Acceptance Tests"
-      url: "acceptance"
-    - title: "Unit Testing Basics"
-      url: "unit-testing-basics"
+    - title: "Application Tests"
+      url: "testing-application"
+    - title: "Testing Basics"
+      url: "testing-basics"
```
jenweber: This is important - we need to either a) figure out redirects rather than killing the old links, or b) keep the same urls _for now_ and do the redirect/corrections when guides-app launches. I recommend the 2nd option. URLs are like public API and we should avoid changing them.

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
-This application would likely include models such as `user` and `post`. It would
-also include interactions such as _login_ and _create post_. Let's finally
-assume that you would like to have [automated tests] in place for your application.
+This application would likely include models such as `user` and `post`.
+It would also include interactions such as _login_ and _create post_.
+Let's finally assume that you would like to have [automated tests] in place for your application.
 
-There are three different classifications of tests that you will need:
-**Acceptance**, **Unit**, and **Integration**.
+There are four different types of test setups you can choose from depending on what you will need to test.
```
jenweber: There are four different types of test setups you can choose from:

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+### Application Tests
 
-Acceptance tests are used to test user interaction and application flow.  The tests interact
-with the application in the same ways that a user would, by doing things like filling out
-form fields and clicking buttons.  Acceptance tests ensure that the features within
-a project are basically functional, and are valuable in ensuring the core features of a
-project have not regressed, and that the project's goals are being met.
+Some tests will require you to test user interaction and application flow.
+In these kind of tests you would like to be able to interact with the application in the same ways that a user would,
+by doing things like filling out form fields and clicking buttons.
```
jenweber: In these kind of tests, you interact with the application in the same ways that a user would, such as filling out form fields and clicking buttons.

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+This will ensure that the features within a project are basically functional,
+and ensure that the core features of a project have not regressed, and that the project's goals are being met.
```
jenweber: Application tests ensure that the interactioons within a project are basically functional, the core features of a project have not regressed, and the project's goals are being met.

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+For tests that require a full application setup we can use the `setupApplicationTest` helper.
+You can read more about how to create these kind of tests in the [Application tests] section.
 
-Unit tests are used to test isolated chunks of functionality, or "units".
-They can be written against any isolated application logic.
+### Rendering Tests
 
-Some specific examples of units tests are:
+Some tests serve as a middle ground between those tests, which require a full application setup and those,
+which interact with specific code algorithms on a micro level.
```
jenweber: What about when you need to test the middle ground between a full application setup and unit tests of specific functions? Consider rendering tests.

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+Oftentimes these tests will verify interactions between various parts of the application,
```
jenweber: often

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+* A `fullName` attribute on a controller is computed by combining its `firstName` and `lastName` attributes.
+* The serializer properly converts the blog request payload into a blog post model object.
+* Blog dates are properly formatted through a `time` service.
+
+We can create these isolated container tests with the `setupTest` helper.
+You can read more about these type of tests in the [Testing Routes] and [Testing Controllers] section.
+
+### Simple Unit Tests
+
+If you would like to test an isolated, small-scoped functionality you can create a unit test.
```
jenweber: comma after functionality

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+Unit tests are the go-to testing type, if you neither require user interactions nor an application container setup to create your test case.
```
jenweber: no comma

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+We can create these type of isolated tests which neither require the application's container to be setup
+nor any user interaction with the `setupTest` helper as well.
+
 ### Testing Frameworks
 
 [QUnit] is the default testing framework for this guide, but others are supported through third-party addons.
 
+### Testing Blueprints
+
+By default whenever you are creating a new component, helper, service or any another module in your app,
```
jenweber: comma after service (just an Oxford comma - it's grammatically correct either way)

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+[Ember CLI](https://ember-cli.com/generators-and-blueprints) will automatically create a QUnit-based test file
+- a so-called testing blueprint - for you.
```
jenweber: drop " a so-called testing blueprint"

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+This blueprint will contain all the basic test setup necessary for testing the module you have just created
+to help you get started writing your first test straight away.
```
jenweber: and help you

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+
+  // Replace this with your real tests.
+  test('it exists', function(assert) {
+    let service = this.owner.lookup('service:location');
+    assert.ok(service);
+  });
+});
+```
+
+For this introduction section only note how this test setup follows a nested structure.
```
jenweber: drop "For this introduction section only"

source/testing/index.md
```diff
@@ -1,79 +1,149 @@
...
+Each test case will be wrapped by a [QUnit.module](https://api.qunitjs.com/QUnit/module) function
+which allows us to scope several tests into a group.
+We can now apply the same testing setup to all tests in this test group easily
+or we can run those tests independently from the rest of the test suite.
```
jenweber: comma after easily.

source/testing/testing-application.md
```diff
@@ -8,39 +8,46 @@ ember g acceptance-test login
...
+  test('visiting /login', async function(assert) {
+    await visit('/login');
     assert.equal(currentURL(), '/login');
   });
 });
 ```
 
-`moduleForAcceptance` deals with application setup and teardown. The last few lines, within
-the function `test`, contain an example test.
+`module` allows you to scope your tests: Any test setup that is done inside of this scope, will
```
jenweber: no comma

source/testing/testing-basics.md
```diff
@@ -0,0 +1,189 @@
+Unit tests, as well as container tests are generally used to test a small piece of code
```
jenweber: Unit tests (as well as container tests)...

source/testing/testing-basics.md
```diff
@@ -0,0 +1,189 @@
...
+
+  computedFoo: computed('foo', function() {
+    const foo = this.get('foo');
+
+    return `computed ${foo}`;
+  })
+});
+```
+
+Within the test for this object we'll create an instance, update the `foo` property (which
```
jenweber: comma after object

source/testing/index.md
jessica-jordan: That's a good point and that's just a direct German translation of mine then. There is the word "sogenannt" which would literally translate to "so-called", but in contrast just means "there's this thing I'm gonna mention right now and it is called as follows". I see how this doesn't work in English though.

I'll update all the other places where I used that wording accordingly 👍

source/testing/testing-models.md
jessica-jordan: I think the `run` loop here is one is good to go without `await`ing it

jessica-jordan:
@jenweber Updated and rebased - can you take another look?

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
...
 
-We can test that changing the component's `name` property updates the
-component's `style` attribute and is reflected in the  rendered HTML:
+  test('should change colors', async function(assert) {
+
+
+  });
+});
+```
+Also note, how the callback function passed to the test helper is marked with the keyword `async`. The [ECMAScript 2017 feature async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) allows us to write asynchronous code in an easy-to-read, seemingly synchronous manner. We can better see what this means, once we start writing out our first test case:
```
jenweber: no comma after note

jenweber:
@jessica-jordan I had another peek and it looks good! I can't review code blocks for accuracy but it looks like you hit all the teeny things I found on the first go. There's one thing that should have a comma taken out, commented above. Thanks again for all your work!

jessica-jordan:
@jenweber Thanks for the quick review, I updated the branch according to your latest comment

source/testing/acceptance.md
```diff
@@ -8,39 +8,46 @@ ember g acceptance-test login
...
 ```
 
-`moduleForAcceptance` deals with application setup and teardown. The last few lines, within
-the function `test`, contain an example test.
+`module` allows you to scope your tests: Any test setup that is done inside of this scope will
+apply to all test cases contained in this module.
+Scoping your tests with `module` also allows you to execute your tests independently from other tests.
+For example, to only run your tests from your `login` module, run `ember test --module='Acceptance | login'`.
+`setupApplicationTest` deals with application setup and teardown.
+The last few lines, within the function `test`, contain an example test.
```
acorncom: Both commas on this line (after `lines` and then after `test`) can be taken out

source/testing/acceptance.md
acorncom: I'm interested in the answer to @Gaurav0 question here as well. @jessica-jordan / @rwjblue do you know if the long-term thinking is to shift away from `find`? Or is this just alternate syntax to using `find()`? Will one of them be preferable over another with the long-term changes to components? What we teach here is likely to be what folks start using who are new to testing, so would love to point them in the best direction possible ...

source/testing/acceptance.md
```diff
@@ -96,6 +103,30 @@ next one starts.
...
+  });
+```
+Before making an assertion, wait for the execution of each asynchronous helper to finish with the `await` keyword:
+
+```tests/acceptance/new-post-appears-first-test.js
+  test('should add new post', async function(assert) {
+    await visit('/posts/new');
+    await fillIn('input.title', 'My new post');
+    await click('button.submit');
+    assert.equal(this.element.querySelector('ul.posts li:first').textContent, 'My new post');
```
acorncom: We could also use `this.element.querySelector` here and make it say something akin to:

```js
assert.equal(this.element.querySelector('ul.posts li:first').textContent, 'My new post'); // alternate syntax to find('ul.posts li:first')
```

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
-with the application in the same ways that a user would, by doing things like filling out
-form fields and clicking buttons.  Acceptance tests ensure that the features within
-a project are basically functional, and are valuable in ensuring the core features of a
-project have not regressed, and that the project's goals are being met.
+Some tests will require you to test user interaction and application flow.
+In these kind of tests, you interact with the application in the same ways that a user would, such as filling out form fields and clicking buttons.
+Application tests ensure that the interactions within a project are basically functional, the core features of a project have not regressed, and the project's goals are being met.
 
-In the example scenario above, some acceptance tests one might write are:
+In the example scenario above, some of these tests you might write are:
```
acorncom: some of *the* tests

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+For tests that require a full application setup we can use the `setupApplicationTest` helper.
+You can read more about how to create these kind of tests in the [Application tests] section.
 
-Unit tests are used to test isolated chunks of functionality, or "units".
-They can be written against any isolated application logic.
+### Rendering Tests
 
-Some specific examples of units tests are:
+What about when you need to test the middle ground between a full application setup and unit tests of specific functions?
+Consider rendering tests.
```
acorncom: Consider *using* rendering tests

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+Rendering tests will verify interactions between various parts of the application,
+such as behavior between UI controls.
+Typically they require an UI to be rendered to be able to execute user interactions in the testing scenario.
```
acorncom: Typically they require *the* UI

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+They are valuable in ensuring data and actions are properly passed between different parts of the system,
```
acorncom: Drop the comma at the end of the line

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+and provide confidence that parts of the system will work within the application under multiple scenarios.
```
acorncom: the application *in* multiple scenarios

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
-scenarios.
-
-It is recommended that components and helpers be tested with integration tests because the component
-interacts with the system in the same way that it will within the context of the application,
+Rendering tests are especially useful for verifying the correct behavior of components and helpers.
+The component or helper interacts with the system in the same way that it will within the context of the application,
 including being rendered from a template and receiving Ember's lifecycle hooks.
 
-Examples of integration tests are:
+Examples of these kind of tests are:
```
acorncom: these *kinds* of tests

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
 * Submitting a post without a title displays a red validation state on the field and gives the user text indicating that the title is required.
 * The blog post list scrolls to position a new post at the top of the viewport.
 
+We can create an isolated rendering test with the `setupRenderingTest` helper.
+You can read more about it in the [Testing Components] or the [Testing Helpers] section.
+
+### Container Tests
+
+Container tests are useful if you are testing a small part of your app which doesn't require user interaction to be tested
+but which requires the application's container to be setup.
```
acorncom: Sweet, I had no idea these were coming 😀

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+If you would like to test a functionality regarding an Ember class instance,
+e.g. a controller, service or route, you can create this kind of test to verify its behavior.
```
acorncom: like to test *the* functionality *of* an Ember class instance

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+
+Some specific examples of these type of tests are:
+
+* A `fullName` attribute on a controller is computed by combining its `firstName` and `lastName` attributes.
+* The serializer properly converts the blog request payload into a blog post model object.
```
acorncom: Is a serializer a good example to use here? I've had headaches testing this in the old system, so if it is, hallelujah! 😀 Otherwise, we might not want to specifically mention it if folks are still going to have issues

acorncom: Also, probably should be *A* serializer

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
 
+### Testing Blueprints
+
+By default whenever you are creating a new component, helper, service, or any another module in your app,
+[Ember CLI](https://ember-cli.com/generators-and-blueprints) will automatically create a QUnit-based test file
+- a testing blueprint - for you.
+This blueprint will contain all the basic test setup necessary for testing the module you have just created
+and help you get started writing your first test straight away.
+
+Imagine we'd create a new `location` service by running `ember generate service location`.
```
acorncom: Image **we** create

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+If we have a look at the newly created blueprint under `tests/unit/services/location-test.js` we'd find the following:
```
acorncom: **we'll* find the following

_(seems better to be in present tense than in past)_

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
 Run your tests with `ember test` on the command-line. You can re-run your tests on every file-change with `ember test --server`.
 
-Tests can also be executed when you are running a local development server (started by running `ember server`), at the `/tests` URI which renders the `tests/index.html` template.
+Tests can also be executed when you are running a local development server (started by running `ember server`),
+at the `/tests` URI which renders the `tests/index.html` template.
 A word of caution using this approach:
-Tests run using `ember server` have the environment configuration `development`, whereas tests executed under `ember test --server` are run with the configuration `test`.  This could cause differences in execution, such as which libraries are loaded and available.  Therefore its recommended that you use `ember test --server` for test execution.
+Tests run using `ember server` have the environment configuration `development`,
+whereas tests executed under `ember test --server` are run with the configuration `test`.
+This could cause differences in execution, such as which libraries are loaded and available.
```
acorncom: Should we even mention the `ember server` and `/tests` approach? I think we shouldn't, as it's considered poor practice (like we mention here). Let's go ahead and ✂️ that discussion completely

source/testing/testing-helpers.md
```diff
@@ -36,59 +36,65 @@ are functions, which can be easily tested with `module`.
...
 });
 ```
 
 As seen in the [Writing Helpers] guide. The helper function expects the unnamed
 arguments as an array as the first argument. It expects the named arguments as
 an object as the second argument.
 
-Now we can move on to an integration test. Integration testing helpers is done
-with the `moduleForComponent` helpers, as shown in [Testing Components].
+Now we can move on to a more complex test case, that ensures our helper is rendered correctly as well. This can be done
```
acorncom: test case (and then snip the comma)

source/testing/testing-models.md
acorncom: Should we link to some additional info about the runloop here? We have a page at https://guides.emberjs.com/v2.18.0/applications/run-loop/ that might at least help explain it if this is one of the first times folks have heard about it ...

source/testing/testing-routes.md
```diff
@@ -35,54 +35,57 @@ export default Route.extend({
...
 
 In this route we've [separated our concerns](http://en.wikipedia.org/wiki/Separation_of_concerns):
-The action `displayAlert` contains the code that is called when the action is 
-received, and the private function `_displayAlert` performs the work. While not 
-necessarily obvious here because of the small size of the functions, separating 
-code into smaller chunks (or "concerns"), allows it to be more readily isolated 
+The action `displayAlert` contains the code that is called when the action is
+received, and the private function `_displayAlert` performs the work. While not
+necessarily obvious here because of the small size of the functions, separating
+code into smaller chunks (or "concerns"), allows it to be more readily isolated
```
acorncom: drop the comma after _(or 'concerns')_

source/testing/testing-components.md
```diff
@@ -133,25 +181,26 @@ We recommend using native DOM events wrapped inside the run loop or the [`ember-
 Using jQuery to simulate user click events might lead to unexpected test results as the action can potentially be called twice.
 
 ```tests/integration/components/magic-title-test.js
-import { moduleForComponent, test } from 'ember-qunit';
```
acorncom: Just above this section we have an entire paragraph that seems to have been made obsolete (about native DOM events, jQuery events, etc). Can we drop that while we're here?

source/testing/testing-components.md
acorncom: A major 👍 on dropping the entire wait explanation below, let's do it. It's been a pain 😀

source/testing/acceptance.md
rwjblue: `find` and `findAll` both exist now from `@ember/test-helpers`, but I still believe that we should primarily teach `this.element.querySelector`. The goal of the recent testing refactors is to reduce “magic” by making it very clear what is going on. `find` is yet another possible piece of “magic”, whereas `element.querySelector` is “just” normal DOM API’s...

source/testing/testing-components.md
rwjblue: A few small things RE wait. 

I agree that this has been a somewhat difficult thing for us to explain historically,  but now that things are essentially the same between all test types I’m hoping it becomes a bit easier...

 `wait()` has been renamed to `settled()` in @ember/test-helpers (because `await wait()` hurts my eyes). 

Generally speaking, I do think it’s important to review the concept of “settled” and to explain that you may need to `await settled()` to properly pause execution. Also, FWIW nearly all of the helpers discussed here are internally `return settled()`...

source/testing/testing-models.md
rwjblue: Agreed. 

FWIW I’m hopefully that future work will allow us to remove all manual Ember.run wrapping (in both tests and app code) and rely on the existing systems to handle “waiting” for scheduled tasks (e.g. a rerender in xyz ms).

source/testing/testing-components.md
```diff
@@ -262,11 +312,11 @@ To stub the location service in your test, create a local stub object that exten
 `Ember.Service`, and register the stub as the service your tests need in the
```
jessica-jordan: @Turbo87  Here's also the updated example on how to stub Services - can you take a look if this is good to go?

Gaurav0:
Not sure if it needs to be in the guide, but you can also import a service, extend it to stub a method or two, and register it in place of the existing service.

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
...
+import { setupRenderingTest } from 'ember-qunit';
+
+module('Integration | Component | pretty color', function(hooks) {
+  setupRenderingTest(hooks);
 
-moduleForComponent('pretty-color', 'Integration | Component | pretty color', {
-  integration: true
 });
 ```
+Inside of your `module` and after setting up the test, we can now start to create our first test case. Here we can use the `QUnit.test` helper and we can
```
Turbo87: `test` function

we shouldn't use the term "helper" more than necessary 😉

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
...
 
-We can test that changing the component's `name` property updates the
-component's `style` attribute and is reflected in the  rendered HTML:
+  test('should change colors', async function(assert) {
+
+
+  });
+});
+```
+Also note how the callback function passed to the test helper is marked with the keyword `async`. The [ECMAScript 2017 feature async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) allows us to write asynchronous code in an easy-to-read, seemingly synchronous manner. We can better see what this means, once we start writing out our first test case:
```
Turbo87: I think we should probably introduce async/await **once** in a testing basics section and then refer to that if we have to. explaining it again and again in every document seems redundant to me.

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
...
+import { render } from '@ember/test-helpers';
 import hbs from 'htmlbars-inline-precompile';
 
-moduleForComponent('pretty-color', 'Integration | Component | pretty color', {
-  integration: true
+module('Integration | Component | pretty color', function(hooks) {
+  setupRenderingTest(hooks);
+
+  test('should change colors', async function(assert) {
+    assert.expect(1);
```
Turbo87: we could probably remove this line to make the test simpler.

my rule of thumb for using `assert.expect()` is to use it (only) when assertions are happening in callbacks. async/await is something like callbacks essentially but since it ensures that we always return a Promise I don't think it's needed in this case.

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
...
+  });
 });
 
-test('should change colors', function(assert) {
-  assert.expect(2);
+```
+
+Each test can use the `render()`
+function imported from the `@ember/test-helpers` package to create a new instance of the component by declaring
+the component in template syntax, as we would in our application.
```
Turbo87: it does indeed "create a new instance", but it does that under the hood. what this thing does conceptually is "render" the component to the DOM.

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
...
+    // set the outer context to red
+    this.set('colorValue', 'red');
 
-  this.set('colorValue', 'blue');
+    await render(hbs`{{pretty-color name=colorValue}}`);
 
-  assert.equal(this.$('div').attr('style'), 'color: blue', 'updates to blue');
+    assert.equal(this.element.querySelector('div').getAttribute('style'), 'color: red', 'starts as red');
+
+    this.set('colorValue', 'blue');
```
Turbo87: @rwjblue I think this will be problematic once we do async rendering, right? would it technically need an `await settled()` then? do we need to adjust this or is this fine for now?

source/testing/testing-components.md
```diff
@@ -133,25 +181,26 @@ We recommend using native DOM events wrapped inside the run loop or the [`ember-
...
-  this.render(hbs`{{magic-title}}`);
+    await render(hbs`{{magic-title}}`);
 
-  assert.equal(this.$('h2').text(), 'Hello World', 'initial text is hello world');
+    assert.equal(this.element.querySelector('h2').textContent.trim(), 'Hello World', 'initial text is hello world');
 
-  //Click on the button
-  run(() => document.querySelector('.title-button').click());
+    //Click on the button
+    run(() => this.element.querySelector('.title-button').click());
```
Turbo87: this should use `await click('.title-button')` instead

source/testing/testing-components.md
```diff
@@ -196,38 +245,39 @@ of a test double (dummy function).
...
 
-  this.render(hbs`{{comment-form submitComment=(action externalAction)}}`);
+    await render(hbs`{{comment-form submitComment=(action externalAction)}}`);
 
-  // fill out the form and force an onchange
-  this.$('textarea').val('You are not a wizard!');
-  this.$('textarea').change();
+    // fill out the form and force an onchange
+    this.element.querySelector('textarea').value = 'You are not a wizard!';
+    this.element.querySelector('textarea').dispatchEvent(new Event('change'));
```
Turbo87: this should use `await fillIn('textarea', 'You are not a wizard!');`

source/testing/testing-components.md
```diff
@@ -196,38 +245,39 @@ of a test double (dummy function).
...
 
-  // click the button to submit the form
-   run(() => document.querySelector('.comment-input').click());
+    // click the button to submit the form
+     run(() => this.element.querySelector('.comment-input').click());
```
Turbo87: this should use `await click('.comment-input');`

source/testing/testing-components.md
```diff
@@ -285,23 +335,25 @@ const locationStub = Service.extend({
...
+module('Integration | Component | location indicator', function(hooks) {
+  setupRenderingTest(hooks);
 
-  beforeEach: function () {
-    this.register('service:location-service', locationStub);
+  hooks.beforeEach(function(assert) {
+    this.owner.register('service:location-service', locationStub);
     // Calling inject puts the service instance in the context of the test,
     // making it accessible as "locationService" within each test
     this.inject.service('location-service', { as: 'locationService' });
```
Turbo87: this does not work anymore and is unrelated to the concept of mocking and stubbing anyways

source/testing/testing-components.md
```diff
@@ -322,28 +374,31 @@ const locationStub = Service.extend({
...
```
Turbo87: doesn't work anymore, and is also unnecessary since `this.locationService` is not used anywhere in the test

source/testing/testing-components.md
```diff
@@ -364,29 +419,35 @@ const locationStub = Service.extend({
...
+
+  test('should change displayed location when current location changes', async function (assert) {
+    await render(hbs`{{location-indicator}}`);
+
+    assert.equal(this.element.textContent.trim(),
+     'You currently are located in New York, USA', 'origin location should display');
+
+    this.set('locationService.city', 'Beijing');
+    this.set('locationService.country', 'China');
+    this.set('locationService.currentLocation', { x: 11111, y: 222222 });
```
Turbo87: changing `currentLocation` seems pointless since the property is not actually used anywhere 🤔

source/testing/testing-components.md
Turbo87: The "Testing User Interaction" section above (once updated) will already use `async/await` to wait for async effects of user actions so IMHO this section can be dropped

source/testing/testing-helpers.md
```diff
@@ -36,59 +36,65 @@ are functions, which can be easily tested with `module`.
...
+    assert.equal(this.element.textContent.trim(), '$1.99');
 
-  this.set('sign', '€');
+    this.set('sign', '€');
 
-  assert.equal(this.$().text().trim(), '€1.99', 'Value is formatted with €');
+    assert.equal(this.element.textContent.trim(), '€1.99', 'Value is formatted with €');
+  });
 });
 ```
```
Turbo87: the "Testing Components" link below seems wrong to me

source/testing/testing-helpers.md
```diff
@@ -27,7 +27,7 @@ export default helper(formatCurrency);
...
 
 Let's start by testing the helper by showing a simple unit test and then move on
-to testing with integration tests afterwards.
+to testing with a rendering test afterwards.
 
 We don't have to use the `moduleFor` helper for unit testing helpers. Helpers
```
Turbo87: this paragraph is outdated

source/testing/testing-helpers.md
```diff
@@ -36,59 +36,65 @@ are functions, which can be easily tested with `module`.
 import { formatCurrency } from 'my-app/helpers/format-currency';
 import { module, test } from 'qunit';
 
-module('Unit | Helper | format currency');
+module('Unit | Helper | format currency', function(hooks) {
+  setupTest(hooks);
```
Turbo87: if you explicitly import the helper then you don't need `setupTest()`. looks like the official blueprint is currently wrong too about this (https://github.com/emberjs/ember.js/blob/v3.0.0-beta.6/blueprints/helper-test/qunit-rfc-232-files/tests/__testType__/helpers/__name__-test.js#L17-L22)

/cc @snewcomer @rwjblue

source/testing/testing-controllers.md
```diff
@@ -1,8 +1,9 @@
-_Unit testing methods and computed properties follows previous patterns shown
-in [Unit Testing Basics] because Ember.Controller extends Ember.Object._
+_Container testing methods and computed properties follows previous patterns shown
+in [Testing Basics] because Ember.Controller extends Ember.Object._
 
-Unit testing controllers is very simple using the unit test helper which is part
-of the ember-qunit framework.
+Testing controllers is very simple using the `setupTest` helper which is part
```
Turbo87: describing something as "simple" in documentation can be problematic for people that are learning about this stuff and don't consider it simple yet.

source/testing/testing-controllers.md
```diff
@@ -33,43 +34,47 @@ export default Controller.extend({
...
+module('Unit | Controller | posts', function(hooks) {
+  setupTest(hooks);
+});
 ```
 
-Next we use `this.subject()` to get an instance of the `PostsController` and
-write a test to check the action. `this.subject()` is a helper method from the
-`ember-qunit` library that returns a singleton instance of the module set up
-using `moduleFor`.
+Next we use the owner API to have access to the controller we'd like to test. Using the `this.owner.lookup` method we get an instance of the `PostsController` and can check the action in our test. The `this.owner.lookup` helper returns objects generated by the framework in your applications
```
Turbo87: "have access"  -> "get access" or "gain access"

Turbo87: > "we get an instance" 

we get **the** instance. controllers are singletons, there is only a single instance.

source/testing/testing-controllers.md
```diff
@@ -33,43 +34,47 @@ export default Controller.extend({
...
+import { setupTest } from 'ember-qunit';
 
-moduleFor('controller:posts', {});
+module('Unit | Controller | posts', function(hooks) {
+  setupTest(hooks);
 
-test('should update A and B on setProps action', function(assert) {
-  assert.expect(4);
+  test('should update A and B on setProps action', function(assert) {
+    assert.expect(4);
```
Turbo87: unnecessary IMHO (see above)

source/testing/testing-controllers.md
```diff
@@ -101,55 +106,43 @@ export default Controller.extend({
...
-    // modify the title of the post
-    postCtrl.get('model').set('title', 'bar');
+      // modify the title of the post
+      postCtrl.get('model').set('title', 'bar');
 
-    // assert that the controllers title has changed
-    assert.equal(ctrl.get('title'), 'bar', 'title is updated');
+      // assert that the controllers title has changed
+      assert.equal(controller.get('title'), 'bar', 'title is updated');
+    });
```
Turbo87: I don't see why a runloop would be necessary here 🤔

source/testing/testing-controllers.md
```diff
@@ -101,55 +106,43 @@ export default Controller.extend({
...
-  // grab an instance of `CommentsController` and `PostController`
-  const ctrl = this.subject();
-  const postCtrl = ctrl.get('post');
+    // grab an instance of `CommentsController` and `PostController`
+    let controller = this.owner.lookup('controller:comments');
+    let postCtrl = controller.get('post');
 
-  // wrap the test in the run loop because we are dealing with async functions
-  run(function() {
+    // wrap the test in the run loop because we are dealing with async functions
```
Turbo87: this comment is very misleading since there isn't a single async/await in this test 🤔

source/testing/testing-routes.md
```diff
@@ -1,15 +1,15 @@
-_Unit testing methods and computed properties follows previous patterns shown 
-in [Unit Testing Basics] because Ember.Route extends Ember.Object._
+_Container testing methods and computed properties follows previous patterns shown
+in [Testing Basics] because Ember.Route extends Ember.Object._
 
-Testing routes can be done both via acceptance or unit tests. Acceptance tests 
-will likely provide better coverage for routes because routes are typically used 
-to perform transitions and load data, both of which are tested more easily in 
+Testing routes can be done both via application tests or more container tests. Application tests
```
Turbo87: > both via application tests or more container tests

why "more"? I'd go for "application or container tests"

source/testing/testing-routes.md
```diff
@@ -35,54 +35,57 @@ export default Route.extend({
...
-  const expectedTextBar = 'bar';
-  window.alert = (text) => {
-    assert.equal(text, expectedTextBar, `expected alert to display ${expectedTextBar}`);
-  };
-    
-  // Now use the routes send method to test the actual action
-  route.send('displayAlert', expectedTextBar);
+    // Now use the routes send method to test the actual action
+    route.send('displayAlert', expectedTextBar);
+  });
```
Turbo87: IMHO this should be two separate tests, though I don't really see the necessity here for having two different things anyway 🤔

source/testing/testing-models.md
```diff
@@ -28,26 +28,25 @@ export default Model.extend({
...
+module('Unit | Model | player', function(hooks) {
   // Specify the other units that are required for this test.
-  needs: []
-});
-
-test('should increment level when told to', function(assert) {
-  // this.subject aliases the createRecord method on the model
-  const player = this.subject({ level: 4 });
+  test('should increment level when told to', function(assert) {
+    // this.subject aliases the createRecord method on the model
```
Turbo87: `this.subject` does not exist anymore

source/testing/testing-models.md
```diff
@@ -28,26 +28,25 @@ export default Model.extend({
...
+    const player = run(() => this.owner.lookup('service:store').createRecord('player'));
 
-  // wrap asynchronous call in run loop
-  run(() => player.levelUp());
+    // wrap asynchronous call in run loop
```
Turbo87: there is no async/await here so calling this async in the comment makes it very confusing

source/testing/testing-models.md
```diff
@@ -77,24 +76,25 @@ export default Model.extend({
...
+  setupTest(hooks);
+
+  test('should own a profile', function(assert) {
+    const User = this.owner.lookup('service:store').modelFor('user');
 
-test('should own a profile', function(assert) {
-  const User = this.store().modelFor('user');
-  const relationship = get(User, 'relationshipsByName').get('profile');
+    // lookup the relationship on the user model
+    const relationship = get(User, 'relationshipsByName').get('profile');
```
Turbo87: why is this using `get(User, ...)` instead of `User.get(...)`?

rwjblue: `DS.Model.get` is not a method (`User` here is a "class" not an instance).

source/testing/testing-components.md
rwjblue: Yes, indeed.  The only reason `this.set` works is because it internally "run wraps" things and forces the rendering to happen.

I'm fine with this for now because we will still have to support forced-sync rendering (specifically in order to avoid breaking tests), we can add the `await settled()` (or a new `await rendered()`) once it is possible to do the `set` without run-wrapping.

source/testing/testing-models.md
Turbo87: if there is no `get` then why not just `User.relationshipsByName.get('profile')`?

rwjblue: Because I think `relationshipsByName` is a computed property on the constructor (yaaasss this is super weird [see here](https://github.com/emberjs/data/blob/7c6448d55efe441e9c62165ecc6bce9114ea4bcb/addon/-private/system/model/model.js#L1523))? I dunno, this kinda confuses me anyways, why do we actually need to enumerate all the relationships on this model?

rwjblue: Like, IMHO, testing that this type of wiring works is super in the weeds. Ideally you make a profile instance, then make a user instance, then confirm that `user.get('profile') === profile;` or something.

Turbo87: yeah, agreed

source/testing/testing-components.md
jessica-jordan: > Generally speaking, I do think it’s important to review the concept of “settled” and to explain that you may need to await settled() to properly pause execution. Also, FWIW nearly all of the helpers discussed here are internally return settled()...

I think that's a good point - it might make sense to explain that in this section, too even though the concept of how to use these ready-to-use dom helpers has already been introduced before. I try to update that section accordingly then

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+Unit tests are the go-to testing type if you neither require user interactions nor an application container setup to create your test case.
+
+Some specific examples of these type of tests are:
+
+* The return value of a `getFullName` utility combines its `firstName` and `lastName` parameters correctly.
+* A computed property macro formats a price depending on its `currency` and `cents` dependent keys.
+* A helper computes the current date properly.
+
+We can create these type of isolated tests which neither require the application's container to be setup
+nor any user interaction with the `setupTest` helper as well.
```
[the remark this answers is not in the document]
jessica-jordan: Yes, you're right, good catch 👍

source/testing/index.md
```diff
@@ -1,79 +1,146 @@
...
+
 ### Testing Frameworks
 
 [QUnit] is the default testing framework for this guide, but others are supported through third-party addons.
 
+### Testing Blueprints
+
+By default whenever you are creating a new component, helper, service, or any another module in your app,
+[Ember CLI](https://ember-cli.com/generators-and-blueprints) will automatically create a QUnit-based test file
+- a testing blueprint - for you.
```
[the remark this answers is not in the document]
jessica-jordan: > ... unless you have ember-mocha installed, in which case it will create a Mocha-based test file.

My intention was to give readers a step by step guide based on a clean slate setup in a new Ember app which is based on QUnit only without any additional testing addons installed. This is what I want to convey in the section above with

> [QUnit] is the default testing framework for this guide

I think it's ok to only declare the most basic, happy path for the testing guides for now and add further documentation outside of the scope of this PR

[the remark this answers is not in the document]
jessica-jordan: > also GitHub renders this line a Markdown list which seems wrong

Does it look ok in the rendered guides app though?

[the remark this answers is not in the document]
jessica-jordan: > and technically we don't create "a testing blueprint", we use the "testing blueprint" to create a test file.

:+1:

source/testing/index.md
jessica-jordan: Where is the recommendation that `ember server` is a poor practice? I read it as a good practice in this section actually:

> Therefore its recommended that you use `ember test --server` for test execution.

source/testing/testing-helpers.md
jessica-jordan: Thanks for pointing that one out, I literally just copied that over from the blueprint as well 👍

source/testing/unit-testing-basics.md
```diff
@@ -28,41 +27,46 @@ export default EmberObject.extend({
...
 ```
 
-See that we have used `moduleFor`, one of the several unit-test helpers provided by Ember-Qunit.
-Test helpers provide us with some conveniences, such as the `subject` function that handles lookup and instantiation for our object under test.
-Note that in a unit test you can customize the initialization of your object under test by passing to the
-`subject` function an object containing the instance variables you would like to initialize.  For example, to initialize
-the property 'foo' in our object under test, we would call `this.subject({ foo: 'bar' });`
+See that first, we are creating a new testing module using the [`QUnit.module`](http://api.qunitjs.com/QUnit/module) function.
+This will scope all of our tests together into one group that can be configured
+and run independently from other modules defined in our test suite.
```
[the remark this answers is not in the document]
jessica-jordan: I think it's ok to mention it in a brief sentence to go through each unit of this test step by step

source/tutorial/routes-and-templates.md
```diff
@@ -325,7 +325,7 @@ Some of the helpers we'll use commonly are:
 
 * [`visit`](http://emberjs.com/api/classes/Ember.Test.html#method_visit) - loads a given URL
 * [`click`](http://emberjs.com/api/classes/Ember.Test.html#method_click) - pretends to be a user clicking on a specific part of the screen
-* [`andThen`](../../testing/acceptance/#toc_wait-helpers) - waits for our previous commands to run before executing our function.
+* `andThen` - waits for our previous commands to run before executing our function.
```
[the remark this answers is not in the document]
jessica-jordan: Agree with that one - let's update the rentals tutorial section in its own PR though, as this is another good bunch of work to get up to date with the new QUnit API

source/testing/testing-models.md
jessica-jordan: I added a link to the run loop documentation in a short sentence below that code example

source/testing/testing-components.md
```diff
@@ -25,74 +25,122 @@ export default Component.extend({
 Pretty Color: {{name}}
 ```
 
-The `moduleForComponent` helper will find the component by name (`pretty-color`)
-and its template (if available).  Make sure to set `integration: true` to enable
-integration test capability.
+The `module` from QUnit will scope your tests into groups of tests which can be
+configured and run independently. Make sure to call the `setupRenderingTest` function together with the `hooks` parameter
+first in your new module. This will do all the setup necessary for testing your component for you, including setting up a way to access your component later on in the test and it will clean up after you once your tests in this module are finished.
```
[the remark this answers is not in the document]
jessica-jordan: Would a better description be `a way to access the rendered DOM of your component` ?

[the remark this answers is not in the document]
Turbo87: yep

source/testing/testing-components.md
jessica-jordan: The helpers for DOM interaction are just introduced later on in this section, so I believe it would make sense to leave it at that for now and update that in another PR for restructuring / improving the testing guides in general

source/testing/testing-components.md
jessica-jordan: same as above

jessica-jordan:
@rwjblue @acorncom @Turbo87  I updated the PR according to your comments, updating anything that is essential for getting the Guides up-to-date to the new Ember QUnit testing API - leaving out further suggested improvements that are not directly related to my changes. 

I think the outstanding issues can and should be tackled in another PR though. Thank you for the thorough review!

source/testing/testing-components.md
Turbo87: I disagree. The pattern above is obsolete and should not be used anymore.

rwjblue: I strongly agree with @Turbo87. Invoking `.click()` on an element (and relying on synchronous updating) is fundamentally bad (this was one of the main factors in deciding to ship interaction helpers)...

jessica-jordan: > The helpers for DOM interaction are just introduced later on in this section, so I believe it would make sense to leave it at that for now and update that in another PR for restructuring / improving the testing guides in general

Should the intro to the DOM interaction helpers be moved to the very beginning of the component testing section then in this PR already so we can refer to that in the subsequent code examples?

Turbo87: in the current structure acceptance/application tests are introduced first which also use these helpers, right? shouldn't they be introduced in that chapter then?

locks:
Let's get this going and iterate 😎. Thank you everyone for your hard work!

PR #2146, opened 2017-12-19
Update installing-addons.md

EdgarAllenPoe:
change "property-type" to category so that the code is consistent with that found in the "Using Ember Data" portion of this tutorial


toddjordan:
Good catch, thanks @EdgarAllenPoe!

PR #2149, opened 2017-12-20
It's not just data attributes

pablobm:
While discussing https://github.com/emberjs/guides/pull/2132, I realised that this text could be misleading: it's not just data attributes that can be affected here. Therefore I thought it appropriate to expand a bit. I still kept the words "data attributes" in the section heading because I bet that's something people search for a lot.

Thoughts?

### DIFF
--- a/source/templates/binding-element-attributes.md
+++ b/source/templates/binding-element-attributes.md
@@ -42,7 +42,7 @@ If `isAdministrator` is `false`, Handlebars will produce the following:
 
 By default, helpers and components only accept a limited number of HTML attributes.
 This means that some uncommon but perfectly valid attributes, such as `lang` or
-custom `data-*` attributes, will not work out of the box. For example, this template:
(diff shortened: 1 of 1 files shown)


jenweber:
Hi @pablobm, sorry I'm getting to this a bit late. I agree with you that it could be clearer and I like your explanation. `lang` and Photos/Fotos is a nice touch! One small semantic thing, I'd rather phrase things about what people can do rather than what they can't. `will not work out of the box` could become `must be specifically enabled.` What do you think?

For my own learning, I made a small [Ember Twiddle](https://ember-twiddle.com/33743a9fb7ff5a2f212639bf7c478f00?openFiles=templates.application.hbs%2Ctemplates.components.another-component.hbs) to try this out myself.

pablobm:
@jenweber Makes sense; I just made that change.

I also made another change in a later paragraph, where I noticed it was still talking about data attributes, despite the section now being more generic than that.

PR #2153, opened 2017-12-27
Update Livereload server port

brunocruzcoelho:
Just created an Ember app and I have `Livereload server on http://localhost:7020`


jenweber:
Hi @brunocruzcoelho, I see the same thing in my apps. Thanks for this PR! Let us know if you spot any other issues. Merged.

PR #2158, opened 2017-12-30
Fix typo

dopin:
Store is not injected into components.
I couldn't know the intention to have components here.
But I think just "all routes" is fine.


source/tutorial/ember-data.md
```diff
@@ -56,7 +56,7 @@ We now have a model object that we can use for our Ember Data implementation.
 
 To use our new Ember Data Model object, we need to update the `model` function we [previously defined](../model-hook/) in our route handler.
 Delete the hard-coded JavaScript Array, and replace it with the following call to the [Ember Data Store service](../../models/#toc_the-store-and-a-single-source-of-truth).
-The [store service](https://www.emberjs.com/api/ember-data/2.16/classes/DS.Store) is injected into all routes and components in Ember.
+The [store service](https://www.emberjs.com/api/ember-data/2.16/classes/DS.Store) is injected into all routes in Ember.
```
locks: This was meant to be "routable controllers", although I'm not sure we should say that. Maybe "route-associated controllers"? cc @jenweber

dopin: I thought it had to be controllers. But I personally avoid to use controller as possible.
Let me know if I should update with "and controllers". ;)

jenweber: "injected into all routes and their corresponding controllers" is what I would say

dopin: Updated ;)

jenweber:
@dopin thanks, you are so fast!

@locks merge?

PR #2160, opened 2018-01-01
Update handling-metadata.md

joshuaob:
JSONAPI meta key is not available on a collection retrieved using the findAll method, but it is available on a collection retrieved using the query method and the last example in this guide makes it confusing as it simply does not work.

### DIFF
--- a/source/models/handling-metadata.md
+++ b/source/models/handling-metadata.md
@@ -70,7 +70,7 @@ import Router from '@ember/routing/route';
 
 export default Route.extend({
   model() {
-    return this.store.findAll('user').then((results) => {
+    return this.store.query('user', {}).then((results) => {
       return {
         users: results,
         meta: results.get('meta')


jenweber:
Hi @joshuaob, thanks for noting this. Do you have a minimal implementation that demonstrates the `findAll` with `meta` missing? I'm a little surprised by that and would like to confirm before we make the change. Thanks!

joshuaob:
Hi Jen, no I do not have a minimal implementation. I stumbled across the issue a few days ago while working on an ember app, it led me to this stack overflow post<https://stackoverflow.com/questions/37645345/ember-js-get-meta-informations-from-json/37645483#comment83076037_37645483> where another user confirmed the issue. I also used the solution outlined in that post.

On 8 Jan 2018, at 03:57, Jen Weber <notifications@github.com<mailto:notifications@github.com>> wrote:


Hi @joshuaob<https://github.com/joshuaob>, thanks for noting this. Do you have a minimal implementation that demonstrates the findAll with meta missing? I'm a little surprised by that and would like to confirm before we make the change. Thanks!

jenweber:
Hi @joshuaob, just to let you know, I got a couple people to look this over to make sure it wasn't a bug in Ember Data. Looks like we're good! It's a known, intended behavior. Thanks a bunch for catching this error and fixing it! It's a big help 😄 

Merged.

PR #2161, opened 2018-01-02
Better explanation of RSVP.hash

rtablada:
I was helping a mentee with RSVP.hash and found the docs were a bit confusing.

I think there's still some more wordsmithing that needs to be done to polish this up, but here's a first pass.


source/routing/specifying-a-routes-model.md
```diff
@@ -123,9 +123,8 @@ Routes without dynamic segments will always execute the model hook.
 
 Multiple models can be returned through an
 [RSVP.hash](https://www.emberjs.com/api/ember/2.16/classes/rsvp/methods/hash?anchor=hash).
-The `RSVP.hash` takes
-parameters that return promises, and when all parameter promises resolve, then
-the `RSVP.hash` promise resolves. For example:
+The `RSVP.hash` method takes an object with promises or values as properties as an argument and returns a single promise.
```
locks: ", and returns"

PR #2163, opened 2018-01-04
Fix typo in Query Parameters

dopin:
`article` should be `articles` here. :cop: :mag:


acorncom:
Good catch, thanks @dopin

PR #2171, opened 2018-01-10
There is no 'helpers/resolver.js' unit test now

myartsev:


locks:
Thank you for the contribution!
The acceptance test image seems a bit smaller than the current one, are you able to update it to match?

rwjblue:
Hmm, if we are trying to match this, the next time we release (3.0.0) will have even fewer files (tests/helpers/* are all gone now)...

locks:
Thanks for the update. I have created an issue to track the Ember 3.0 changes.

PR #2176, opened 2018-01-12
Remove Bower as a valid source for community adapters

myartsev:


locks:
Thanks for cleaning up :)

PR #2179, opened 2018-01-16
Fixes #2173: mention that a test server also needs to be restarted after installing an add-on

### LINKED ISSUE #2173: Tutorial: all acceptance tests fail once we use Mirage instead of hardcoded model data
After the switch to [using Mirage](https://guides.emberjs.com/v2.18.0/tutorial/ember-data/) is complete, all the acceptance tests fail.

failed_acceptance

It appears that there is an extra step required to get Mirage to work in a test environment: http://www.ember-cli-mirage.com/docs/v0.3.x/manually-starting-mirage/

myartsev:
Fixes #2173.

### DIFF
--- a/source/tutorial/installing-addons.md
+++ b/source/tutorial/installing-addons.md
@@ -152,6 +152,6 @@ export default DS.JSONAPIAdapter.extend({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


acorncom:
Fixes #2173

PR #2187, opened 2018-01-24
Updated for new modules @ember/object/evented documentation link

morhook:
`Ember.on` is the old-way, now it should be referenced `{on} from @ember/object/evented`


morhook:
Closing in favor of #2187

PR #2189, opened 2018-01-27
Import Route from the new JS modules API

skaterdav85:
This page had mixed usage of where Route came from. I changed it to use `import Route from '@ember/routing/route';` for consistency.


acorncom:
Thanks for the help!

Issue #2195, opened 2018-01-28
Update the navbar to match the main website


jenweber:
See diffs at https://github.com/emberjs/website/pull/3157/files
Can't import all of bootstrap or the styling explodes :(

Current work can be found in `feature/new-header`
![screen shot 2018-01-28 at 6 38 42 pm](https://user-images.githubusercontent.com/16627268/35488804-47aa86ba-045c-11e8-8b92-1a0c9c59b45d.png)

jenweber:
closed with https://github.com/emberjs/guides/pull/2198

PR #2198, opened 2018-01-29
bugfix hamburger menu that is always open (#2195)

jenweber:
I added the collapse js to bootstrap-dropdown and called collapse in a script tag at the bottom of the `_header.erb` file. Opened here in case someone has a better plan.

Previous behavior (present in production) is that the hamburger menu always stays open in mobile 😟  I added some jquery and bootstrap to get it working like a normal hamburger menu - renders closed, click to open. `bootstrap-dropdown` is all just bootstrap basic boilerplate.

Test by running it locally and making the screen tiny. The dropdown flashes for a second but I am not sure how to fix it (and don't think we need to). Seems better than what we have now though.

Check to make sure that the navbar contents aren't still visible when closed:

screen shot 2018-01-29 at 10 39 48 pm


source/stylesheets/base/_bootstrap-navbar.scss
```diff
@@ -816,6 +812,184 @@ select[multiple].input-lg {
...
+     transition-timing-function: ease;
+}
+.caret {
+  display: inline-block;
+  width: 0;
+  height: 0;
+  margin-left: 2px;
+  vertical-align: middle;
+  border-top: 4px dashed;
+  border-top: 4px solid \9;
```
jessica-jordan: The `border-top` property is duplicated and I assume the `\9` could be replaced with a border-style or a border-color - what would work best here?

source/stylesheets/base/_bootstrap-navbar.scss
```diff
@@ -816,6 +812,184 @@ select[multiple].input-lg {
...
+  display: none;
+  float: left;
+  min-width: 160px;
+  padding: 5px 0;
+  margin: 2px 0 0;
+  list-style: none;
+  font-size: 14px;
+  text-align: left;
+  background-color: #ffffff;
+  border: 1px solid #cccccc;
```
jessica-jordan: Is there a variable already available for this color that could be reused?

source/stylesheets/base/_bootstrap-navbar.scss
```diff
@@ -816,6 +812,184 @@ select[multiple].input-lg {
...
+  z-index: 1000;
+  display: none;
+  float: left;
+  min-width: 160px;
+  padding: 5px 0;
+  margin: 2px 0 0;
+  list-style: none;
+  font-size: 14px;
+  text-align: left;
+  background-color: #ffffff;
```
jessica-jordan: Although pure white is straight-forward color, might it make sense to use a variable here (if available) to make sure that we can easily replace it in the future when the style guide changes?

source/stylesheets/base/_bootstrap-navbar.scss
jenweber: This is just bootstrap boilerplate so I'm inclined to leave it as-is unless it causes a problem

jenweber:
@jessica-jordan I don't think we should refactor the bootstrap stylesheets themselves, as this is a stop gap measure until we have Ember-styleguide together. The CSS is just a copy-paste from the bootstap website

source/stylesheets/base/_bootstrap-navbar.scss
jessica-jordan: Yes, sounds good!

PR #2208, opened 2018-01-31
Update links to documentation

rimian:
These links are to an older version of the docs. Updated :)


source/routing/defining-your-routes.md
```diff
@@ -269,5 +269,5 @@ so that when a user navigates to `/a/non-existent/path` they will be shown a mes
 To have your route do something beyond render a template with the same name, you'll
 need to create a route handler. The following guides will explore the different
 features of route handlers. For more information on routes, see the API documentation
-for [the router](https://www.emberjs.com/api/ember/2.16/classes/Router) and for [route
```
locks: Thanks for contributing! Can you instead update to `https://www.emberjs.com/api/ember/release/classes/Router`? It should now be working, and we won't need to continuously update :)

rimian: That's a much better idea! Done :)

PR #2225, opened 2018-02-14
Remove special reference to "needs" in controller testing.

rwjblue:
This is not needed with the newer testing system because the concept of `needs` is gone from the testing harness.


locks:
❤️

PR #2227, opened 2018-02-15
testing: Wrap text after 80 characters

Turbo87:
... because that makes it much more readable in raw text mode


Turbo87:
and then @locks shows me https://github.com/emberjs/guides/blob/master/CONTRIBUTING.md#style-guide and I'm closing the PR again... 😭

PR #2229, opened 2018-02-16
Fix typo

abhishek97:
`Rather than create an arbitrary set of conventions` => `Rather than creating an arbitrary set of conventions`


sivakumar-kailasam:
Thanks for the PR @abhishek97 👍

PR #2231, opened 2018-02-16
Update js-primer.md

raongithub:
Change from >>> You can also not access them before they are declared, or you will get a ReferenceError.
Change to      >>> You also cannot access them before they are declared, or you will get a ReferenceError.


raongithub:
Kindly review.

PR #2240, opened 2018-02-21
Added setupTest(hooks) to tests/unit/models/player-test.js

nizarayari:
I think you forgot to call setupTest(hooks) in the unit-test


jenweber:
Ah yep, we did! Thanks a bunch for taking the time to stop by and fix it!

PR #2243, opened 2018-02-22
[DOC]-update-deprecated-test-helpers

sbatson5:
Ran into a few issues while updating an app to `3.0`.  

* `keyEvent` has been removed in favor of `triggerKeyEvent`
    * Using `keyEvent` will give you the error:
    * `Promise rejected during "visiting /thing": (0 , _testHelpers.keyEvent) is not a function`
* `currentPath` has been removed
* `findAll` has been added

### DIFF
--- a/source/testing/acceptance.md
+++ b/source/testing/acceptance.md
@@ -90,7 +90,7 @@ Some of these handy helpers are:
 * [`fillIn(selector, value)`][2]
   - Fills in the selected input with the given value and returns a promise that
      fulfills when all resulting async behavior is complete. Works with `<select>` elements as well as `<input>` elements. Keep in mind that with `<select>` elements, `value` must be set to the _value_ of the `<option>` tag, rather than its _content_ (for example, `true` rather than `"Yes"`).
-* [`keyEvent(selector, type, keyCode)`][3]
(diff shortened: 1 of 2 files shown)


jenweber:
Hi @sbatson5, thanks for this PR! In addition to the API method name, we also need to update the link that it goes to. Right now, the docs for ember-test-helpers are at https://github.com/emberjs/ember-test-helpers/blob/master/API.md

@jessica-jordan does that seem like the right place to link to for now?

source/testing/acceptance.md
```diff
@@ -130,13 +130,11 @@ previous steps in the test have completed.
 
 Synchronous helpers are performed immediately when triggered.
 
-* [`currentPath()`][6]
```
sbatson5: Not sure if I missed it, but it looks like `currentPath` has been removed:
https://github.com/emberjs/ember-test-helpers/blob/37c4139ee73c19a46670268825881d81eb37d663/tests/unit/setup-application-context-test.js#L6-L14

jenweber:
Thanks a bunch @sbatson5 😁 I appreciate the extra effort to update the link references through the whole thing.

Note to maintainers, we'll have more work to do to clean up this page once we get the test helper docs into the API docs app.

Issue #2215, opened 2018-02-04
stack level too deep on `bundle exec middleman`


jenweber:
```
== The Middleman is loading
/usr/local/var/rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:121: warning: constant ::Fixnum is deprecated
/usr/local/var/rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:121: warning: constant ::Bignum is deprecated
bundler: failed to load command: middleman (/usr/local/var/rbenv/versions/2.4.3/bin/middleman)
SystemStackError: stack level too deep
```

Replicated locally (mac) and on a cloud9 instance. Using `rbenv`.

acorncom:
@jenweber if you shift to Ruby 2.2 do you still hit this problem? It's working fine on my end when doing so ...

jenweber:
@acorncom yes it seems to work for 2.2.2. Did you have issues with 2.4.3 as well or is it just me? If it worked, can you or @locks do bundle install, try to run it, and post your specs if it works? I'm wondering if you have older versions of dependencies hanging around, old-school-npm-style.

I removed all gems, reinstalled, then tried 2.4.3 and still got the same error.

Specs:
- Mac OS 10
- rbenv 1.1.0
- gem 2.4.5
- ruby 2.2.2 works

jenweber:
Also I don't know why I didn't post this earlier, but I get an infinite loop of this error:

> /usr/local/var/rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:131:in `block (2 levels) in <class:Numeric>'

jenweber:
I blindly updated a bunch of dependencies and it seems to work now. Will PR. Disclaimer, I have no idea what I'm doing.

jenweber:
closed with #2249

PR #2249, opened 2018-02-25
update dependencies to solve stack level too deep error on local build

jenweber:
#2215 
Disclaimer, I have no idea what any of these dependencies are or what they do, whatsoever. I did `bundle update` and made some random changes to gemfile until I was able to run the guides app locally. 

`¯\_(ツ)_/¯`

Definitely deploy to staging before merging. I will also test this in a clean environment. ~It will be a miracle if the Travis build passes.~ Somehow it passed. Yay?


locks:
I am relatively sure this will horribly break because our code is not compatible with middleman 4 🤔

jenweber:
I dunno, it runs ok... I will try moving it to 3 and see what happens

jenweber:
I updated only the activerecord version and that seems to fix it.

PR #2251, opened 2018-02-28
Update testing-models.md

denzo:


source/testing/testing-models.md
```diff
@@ -50,7 +50,7 @@ module('Unit | Model | player', function(hooks) {
 ```
 
 Also note, how both creating a record and updating properties on the record through the `levelUp` method requires
-us to wrap these operations into a `run` function. You can read more the Ember run loop [over here](https://guides.emberjs.com/v2.18.0/applications/run-loop/).
+us to wrap these operations into a `run` function. You can read more the Ember run loop [over here](https://guides.emberjs.com/v3.0.0/applications/run-loop/).
```
sivakumar-kailasam: This could probably be https://guides.emberjs.com/current/applications/run-loop/

sivakumar-kailasam:
Thanks for the PR @denzo, if you could change the URL to current we could merge this

denzo:
@sivakumar-kailasam done 👌

jenweber:
Hey @denzo sorry to ask for you to edit this again, but links within the guides should be relative, so that when someone follows them in the future, they don't end up looking at docs for a newer version than what they were viewing. I think this was originally hardcoded at a specific version by mistake.

@sivakumar-kailasam we should only really be using `release` and `current` urls for linking to the API docs.

Here's an example of a relative link:
```
[some example](../ember-data#toc_updating-the-model-hook)
```

denzo:
@jenweber great idea. Done 👌

jenweber:
Merged, thanks @denzo for the quick turnaround and the help!

PR #2252, opened 2018-03-01
Add 3.0 test concepts to the tutorial

toddjordan:
Goal is the convert the tutorial over to use the new test format and apis.


source/tutorial/acceptance-test.md
```diff
@@ -42,18 +42,18 @@ Since we haven't added any functionality to our application yet, we'll use this
...
-test('visiting /list-rentals', function(assert) {
-test('visiting /', function(assert) {
-  visit('/list-rentals');
-  visit('/');
-
-  andThen(function() {
+```/tests/acceptance/list-rentals-test.js{-8,+9,-10,+11,-12,+13}
+import { module, test } from 'qunit';
+import { visit, currentURL } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
```
Turbo87: I'd suggest putting this line right below the qunit import to group them a little better

source/tutorial/autocomplete-component.md
```diff
@@ -510,21 +510,34 @@ Open our existing acceptance test, `tests/acceptance/list-rentals-test.js`, and
...
-  keyEvent('.list-filter input', 'keyup', 69);
-  andThen(function() {
-    assert.equal(find('.listing').length, 1, 'should show 1 listing');
-    assert.equal(find('.listing .location:contains("Seattle")').length, 1, 'should contain 1 listing with location Seattle');
-  });
+test('should filter the list of rentals by city', async function(assert) {
+  await visit('/');
+  await fillIn('.list-filter input', 'seattle');
+  await triggerKeyEvent('.list-filter input', 'keyup', 69);
+  assert.equal(findAll('.results .listing').length, 1, 'should display 1 listing');
```
Turbo87: assert.ok(find(...)) should be sufficient here if you only want to check for the existance of something

source/tutorial/autocomplete-component.md
```diff
@@ -510,21 +510,34 @@ Open our existing acceptance test, `tests/acceptance/list-rentals-test.js`, and
...
+  assert.ok(find('.listing .location').innerHTML.includes('Seattle'), 'should contain 1 listing with location Seattle');
```
Turbo87: Use textContent instead of innerHTML

source/tutorial/autocomplete-component.md
Turbo87: also find/findAll mainly exist to ease the migration from the old testing APIs. new code, according to @rwjblue, should use this.element.querySelector() (or qunit-dom one the RFC is merged)

source/tutorial/routes-and-templates.md
```diff
@@ -323,30 +329,34 @@ such as visiting routes, filling in fields, clicking on links/buttons, and waiti
...
-test('should link to information about the company.', function (assert) {
-  visit('/');
-  click('a:contains("About")');
-  andThen(function() {
-    assert.equal(currentURL(), '/about', 'should navigate to about');
-  });
+```/tests/acceptance/list-rentals-test.js{+2,+3,+4,+8,+9,+10}
+test('should link to about page', async function(assert) {
+  await visit('/');
+  await click("a[href='/about']");
```
Turbo87: I guess you are aware that this is not exactly best practice :wink:

source/tutorial/service.md
```diff
@@ -366,27 +366,40 @@ To stub these services we simply have to register a stub service that implements
...
-  beforeEach() {
-    this.application.register('service:maps', StubMapsService);
-  }
+module('Acceptance | list-rentals');
+module('Acceptance | list rentals', function(hooks) {
+  setupApplicationTest(hooks);
+
+  hooks.beforeEach(function() {
+    let { owner } = getContext();
+    owner.register('service:maps', StubMapsService);
```
Turbo87: That getContext() is unnecessary because it's the same as `this`. You can just use this.owner.register()

source/tutorial/service.md
```diff
@@ -366,27 +366,40 @@ To stub these services we simply have to register a stub service that implements
...
+import {
+  click,
+  currentURL,
+  find,
+  findAll,
+  visit,
+  fillIn,
+  getContext,
+  triggerKeyEvent
+} from '@ember/test-helpers'
```
Turbo87: Most of these seem unused right now 🤔

source/tutorial/simple-component.md
```diff
@@ -202,106 +202,112 @@ Our component integration test will test two different behaviors:
...
+  setupRenderingTest(hooks);
+
+  let rental = EmberObject.create({
+    image: 'fake.png',
+    title: 'test-title',
+    owner: 'test-owner',
+    category: 'test-type',
+    city: 'test-city',
+    bedrooms: 3
+  });
```
Turbo87: Since we can't guarantee that nothing mutates that object we should create a new one for each test in a beforeEach block

source/tutorial/simple-component.md
```diff
@@ -312,63 +318,66 @@ In the first test, we just want to verify the output of the component, so we jus
 ```tests/integration/components/rental-listing-test.js{+4,+5}
 test('should display rental details', function(assert) {
   this.set('rentalObj', rental);
-  this.render(hbs`{{rental-listing rental=rentalObj}}`);
+  await render(hbs`{{rental-listing rental=rentalObj}}`);
   assert.equal(this.$('.listing h3').text(), 'test-title', 'Title: test-title');
```
Turbo87: This probably should not use jquery anymore

source/tutorial/subroutes.md
```diff
@@ -379,15 +379,13 @@ Regardless, we hope this has helped you get started with creating your own ambit
...
-    assert.equal(currentURL(), '/rentals/grand-old-mansion', 'should navigate to show route');
-    assert.equal(find('.show-listing h2').text(), "Grand Old Mansion", 'should list rental title');
-    assert.equal(find('.description').length, 1, 'should list a description of the property');
-  });
+```/tests/acceptance/list-rentals-test.js{+2,+3,+4,+5,+6}
+test('should show details for a specific rental', async function(assert) {
+  await visit('/rentals');
+  await click("a[href='/rentals/grand-old-mansion']");
+  assert.equal(currentURL(), '/rentals/grand-old-mansion', "should navigate to show route");
+  assert.ok(find('.show-listing h2').innerHTML.includes("Grand Old Mansion"), 'should list rental title');
```
Turbo87: as above, textContent over innerHTML

source/tutorial/acceptance-test.md
```diff
@@ -78,35 +78,37 @@ Our launched Chrome web browser now shows 9 successful tests. If you toggle the
...
 
-moduleForAcceptance('Acceptance | list-rentals');
+module('Acceptance | list-rentals', function(hooks) {
+  setupApplicationTest(hooks);
 
-test('should show rentals as the home page', function (assert) {
-});
+  test('should show rentals as the home page', function (assert) {
```
sbatson5: The blueprints will generate these as `async` functions. Should you start with them as async here?

source/tutorial/simple-component.md
toddjordan: yes.  in the process of killing all the jqueries 👍

source/tutorial/subroutes.md
toddjordan: thx for the tip.  will update ❤️

source/tutorial/acceptance-test.md
```diff
@@ -42,18 +42,18 @@ Since we haven't added any functionality to our application yet, we'll use this
...
+import { visit, currentURL } from '@ember/test-helpers';
+import { setupApplicationTest } from 'ember-qunit';
+
+module('Acceptance | my acceptance test', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('visiting /list-rentals', async function(assert) {
+  test('visiting /', async function(assert) {
+    await visit('/list-rentals');
+    await visit('/');
```
alexander-alvarez: I'm not following this test case, so maybe I'm missing something but shouldn't `assert.equal(currentURL(), '/list-rentals');` be between the `await visit('/');`

Feel free to disregard if I'm completely missing the context

source/tutorial/simple-component.md
```diff
@@ -312,63 +318,66 @@ In the first test, we just want to verify the output of the component, so we jus
...
-  this.render(hbs`{{rental-listing rental=rentalObj}}`);
-  assert.equal(this.$('.image.wide').length, 0, 'initially rendered small');
-  run(() => document.querySelector('.image').click());
-  assert.equal(this.$('.image.wide').length, 1, 'rendered wide after click');
-  run(() => document.querySelector('.image').click());
-  assert.equal(this.$('.image.wide').length, 0, 'rendered small after second click');
+module('Integration | Component | rental listing', function (hooks) {
+  setupRenderingTest(hooks);
+
+  let rental = EmberObject.create({
```
alexander-alvarez: ditto above with `beforeEach`

source/tutorial/acceptance-test.md
toddjordan: I think this is a merge error.  good catch

toddjordan: The  +, - numbers at the top of the code block indicate added/removed lines.

source/tutorial/routes-and-templates.md
toddjordan: struggling with the best practice here... I used to do `click('a:contains("About")'` which no longer works.   Anyways open to better practices.

source/tutorial/service.md
toddjordan: typo.

source/tutorial/routes-and-templates.md
Turbo87: best practice is using something like https://github.com/simplabs/ember-test-selectors 😉

source/tutorial/autocomplete-component.md
toddjordan: won't qunit-dom back find when that happens?  According to https://github.com/emberjs/ember-test-helpers/blob/master/API.md#find that might be happening already

Turbo87: I don't understand the question. qunit-dom is not using `find()` under the hood, but `document.querySelector()` instead.

toddjordan: sorry.  wasn't clear.  ember-test-helpers find is using querySelector

toddjordan: I see what you are saying.  querySelecter is done at root with find

source/tutorial/routes-and-templates.md
toddjordan: yeah, agree.  Just hesitating on bringing another addon into the tutorial....

source/tutorial/service.md
toddjordan: cool. didn't realize owner was available and public these days

source/tutorial/autocomplete-component.md
toddjordan: Is this.element documented somewhere?

toddjordan: does this.element return the root in an acceptance test anyways (sorry for the multiple comments, my thinking is spaced out while I work lol)

source/tutorial/acceptance-test.md
```diff
@@ -19,7 +19,7 @@ At this point, you can continue to the [next page](../routes-and-templates/) or
 
 ### Testing Our Application As We Go
 
-We can represent the goals above as [Ember acceptance tests](../../testing/acceptance/). Acceptance tests interact with our app like an actual person would, but are automated, helping ensure that our app doesn't break in the future.
+We can represent the goals above as [Ember acceptance tests](../../testing/acceptance/). Application tests interact with our app like an actual person would, but are automated, helping ensure that our app doesn't break as you make changes.
```
sivakumar-kailasam: So `application` tests?  Is this a new term we're introducing? If so how do we announce this to the community?

source/tutorial/autocomplete-component.md
```diff
@@ -303,57 +303,63 @@ similar to [how we tested our rental listing component earlier](../simple-compon
...
 
-test('should initially load all listings', function (assert) {
-  // we want our actions to return promises,
-  //since they are potentially fetching data asynchronously
-  this.on('filterByCity', () => {
-    return RSVP.resolve({ results: ITEMS });
+  test('should initially load all listings', async function (assert) {
+    // we want our actions to return promises,
+    //since they are potentially fetching data asynchronously
+    this.on('filterByCity', () => {
```
sivakumar-kailasam: this could just be `this.on('filterByCity', resolve({results: ITEMS});`

source/tutorial/autocomplete-component.md
```diff
@@ -368,90 +374,95 @@ Since our component is expecting the filter process to be asynchronous, we retur
...
+            {{item.city}}
+          </li>
+        {{/each}}
+        </ul>
+      {{/list-filter}}
+    `);
+
+    return settled().then(() => {
+      assert.equal(this.element.querySelectorAll('.city').length, 3);
+      assert.equal(this.element.querySelector('.city').textContent().trim(), 'San Francisco');
```
sivakumar-kailasam: `qunit-dom` just made it into ember-cli's default blueprints, so should we start mentioning it in the guides too? Because anyone who reads these changes going forward will see it in their newly generated app. We can always add a note mentioning that it comes recommended & that people should start using it even in their old apps after using the [codemod](https://github.com/simplabs/qunit-dom-codemod)

source/tutorial/routes-and-templates.md
sivakumar-kailasam: @toddjordan may be add a note at the end of this page of the guides mentioning the [testing section](https://emberobserver.com/categories/testing) of ember observer as a way to further improve the testing experience?

Turbo87: > Just hesitating on bringing another addon into the tutorial....

I do agree on that.

We could still use the `data-test-foobar` attributes though since they do work without the addon. The main thing that the addon does is stripping those attributes out from the production build to save bytes.

But in the end it would be one additional thing to teach, so maybe not ideal in this context.

source/tutorial/acceptance-test.md
Turbo87: yes, with the new testing APIs "acceptance" tests are replaced by "application" tests.

source/tutorial/autocomplete-component.md
Turbo87: I don't think this will work either since `this.on()` is no longer a thing. 

I think the following should work:
```js
this.set('filterByCity', () => resolve(...));
```

```hbs
{{some-component onFoo=(action filterByCity)}}
```

source/tutorial/autocomplete-component.md
Turbo87: yes, we should update the guides to use it by default. it has the nice side-effect of making the assertions much less noisy.

source/tutorial/autocomplete-component.md
toddjordan: sorry the codemod in super-rentals repo did this, which works:
```javascript

  test('should initially load all listings', async function(assert) {
    assert.expect(2);
    this.actions.filterByCity = () => {
      return resolve({ results: ITEMS });
    };

    await render(hbs`
      {{#list-filter filter=(action 'filterByCity') as |rentals|}}
        <ul>
        {{#each rentals as |item|}}
          <li class="city">
            {{item.city}}
          </li>
        {{/each}}
        </ul>
      {{/list-filter}}
    `);

    return settled().then(() => {
      assert.equal(this.element.querySelectorAll('.city').length, 3);
      assert.equal(this.element.querySelector('.city').textContent.trim(), 'San Francisco');
    });
  });
```

toddjordan: will update here...

alexander-alvarez: to my knowledge the codemod did the `this.actions.<actionName>` because it was safer than potentially blowing away people's properties, not because it is encouraged to write `this.actions`  @rwjblue  to confirm

given that glimmer actions are just functions and will be passed in like `(action filterByCity)` and we will eventually get there, I prefer @Turbo87 's recommendation.

In my tests I use the below to great sucess
```js
this.filterByCity = ( ) => {
 // stuff
}
await render(hbs`{{my-component click=(action filtlerByCity)}`);
```

but I'm weird because I do that also in my components too instead of using the `actions`  hash most of the time

toddjordan: ok, will update 👍

rwjblue: Yes, confirm RE: the codemod results. Whenever possible, we should use and suggest closure actions.

source/tutorial/autocomplete-component.md
toddjordan: super-rentals is on 3.0, and the blueprints aren't qunit dom, is it in beta?

toddjordan:
@sivakumar-kailasam removed EmberObject usage in tests with regular `{}`s

source/tutorial/autocomplete-component.md
sivakumar-kailasam: It's on master so guessing it'll be out with 3.1

source/tutorial/acceptance-test.md
```diff
@@ -62,56 +62,58 @@ test('visiting /', function(assert) {
 
 A few of things to note in this simple test:
 
-* Acceptance tests are setup by calling the function `moduleForAcceptance`. This function ensures that your Ember application is started and shut down between each test.
+* Application tests are setup by calling the function `setupApplicationTest`. This function ensures that your Ember application is started and shut down between each test.
 * QUnit passes in an object called an [`assert`](https://api.qunitjs.com/assert/) to each test function. An `assert` has functions, such as `equal()`, that allow your test to check for conditions within the test environment. A test must have one passing assert to be successful.
-* Ember acceptance tests use a set of test helper functions, such as the `visit`, `andThen`, and `currentURL` functions used above. We'll discuss those functions in more detail later in the tutorial.
+* Ember acceptance tests use a set of test helper functions, such as the `visit` and `currentURL` functions used above. We'll discuss those functions in more detail later in the tutorial.
```
sivakumar-kailasam: This line still uses `acceptance` instead of `application`.

source/tutorial/autocomplete-component.md
toddjordan: we'll circle back and update then

PR #2253, opened 2018-03-01
Fix "Testing Models" test example

patocallaghan:
I noticed the [Testing Models](https://guides.emberjs.com/v3.0.0/testing/testing-models/) example was missing the `setupTest` call.


jenweber:
Good catch, thank you @patocallaghan for the code and @rwjblue for the review.

PR #2268, opened 2018-03-10
Can not use keyword 'await' outside an async function

sajidmatrix:
helper tests 'rental-property-type-test' are failing since async keyword is not allowed outside an async function.


toddjordan:
Thanks for the fix!!!

PR #2269, opened 2018-03-11
Update model-hook.md

algo74:
Test fails if     
`assert.equal(this.element.querySelectorAll('.results .listing').length, 3, 'should display 3 listings');`
because listings are not in an element with class "results"


toddjordan:
according to the example set up in the [super rentals repo](https://github.com/ember-learn/super-rentals) listing is in results:
image

algo74:
`<ul class="results">` is added in [https://guides.emberjs.com/v3.0.0/tutorial/autocomplete-component/](https://guides.emberjs.com/v3.0.0/tutorial/autocomplete-component/)
image
the code in [https://guides.emberjs.com/v3.0.0/tutorial/model-hook/](https://guides.emberjs.com/v3.0.0/tutorial/model-hook/) doesn't have it yet, but the test is already there.

toddjordan:
yep you're right.  good catch.

PR #2271, opened 2018-03-11
Fix second test in autocomplete-component

algo74:
At least in my case, the test was falling because `triggerKeyEvent(this.element.querySelector('.list-filter input'), "keyup", 83);` apparently did not modify the value of the input field.


toddjordan:
Works in the super rentals repo: https://github.com/ember-learn/super-rentals/blob/master/tests/integration/components/list-filter-test.js#L65

Maybe try that out and try to find the difference in what you are seeing.

algo74:
In the repo, the line 68 is
`assert.ok(this.element.querySelector('.city'), 'one result found');`
if replaced with correct 
`assert.equal(this.element.querySelectorAll('.city').length, 1, 'One result returned');`
the test fails.

My solution is probably not very good though. For instance, `fillIn` is introduced in the next test on this page.

toddjordan:
it seems like the assert.ok would be ok, since I believe it gives back undefined if it can't find the selector.  Will have to investigate though.

algo74:
The test must assert that the component shows exactly one item. A query with empty string returns three items (it is tested in the first test).

toddjordan:
You're right.  Thanks @algo74 ❤️

PR #2277, opened 2018-03-13
Make docs for CP laziness more explicit.

### LINKED ISSUE #2263: Computed properties need a clearer warning about when they get updated
https://guides.emberjs.com/v3.0.0/object-model/computed-properties/

Computed properties need a clearer warning that they only get updated when you call them or when they are being used in the markdown of an active template. This has probably bitten every Ember developer. Check this section over and see if it needs something stronger/clearer.

_Note, this issue is set aside for the EmberConf contributors workshop on 3/12. If it's not claimed by the end of that day, feel free to grab it!_

Bestra:
Fixes #2263.

This first pass adds the section about computed property laziness right after the very first introduction.  It might be better further down with the 'Dynamic Updating' section though.

### DIFF
--- a/source/object-model/computed-properties.md
+++ b/source/object-model/computed-properties.md
@@ -40,6 +40,48 @@ The first time you access the `fullName` property, the function will be called a
 Subsequent access of `fullName` will read from the cache without calling the function.
 Changing any of the dependent properties causes the cache to invalidate, so that the computed function runs again on the next access.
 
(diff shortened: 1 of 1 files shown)


jenweber:
Thank you @Bestra! This explanation would have saved me a lot of grief back in the day :) Let me know if you have any questions/opinions about my feedback.

acorncom:
Looks good to me, thanks Chris

Issue #2273, opened 2018-03-12
Explain `.[]` before `.@each` in "Computed Properties and Aggregate Data"


locks:
Guide: https://guides.emberjs.com/v3.0.0/object-model/computed-properties-and-aggregate-data/

This guide explains `@each`, and then compares it to `[]` but there is no explanation of what `[]` is, making the property confusing.
There are also many cases where `[]` is the appropriate API to use, but people only know about `@each`.

Bestra:
I'll plan on picking this up some time this week, but if anyone else is burning to take it from me let me know!

acorncom:
Closed by #2282

PR #2282, opened 2018-03-13
Add documentation for '[]'

Bestra:
Addresses #2273. 
- Document `[]` before talking about `[]`
- Shuffles the `@each` heading to be consistent with the new `[]
- reorders the circumstances for busting the CP cache from least specific
  to most specific

### DIFF
--- a/source/object-model/computed-properties-and-aggregate-data.md
+++ b/source/object-model/computed-properties-and-aggregate-data.md
@@ -5,9 +5,9 @@ array properties to track changes on them, `[]` and `@each`.
 
 ## `[]`
 
-Sometimes a computed property needs to update when items are added or removed from an array.
+Sometimes a computed property needs to update when items are added to, removed from, or replaced in an array.
 In those cases we can use the `[]` array key to tell the property to update at the right time.
-We'll use the obligatory todo list for our examples:
+We'll use the familiar todo list for our examples:
 
 ```app/components/todo-list.js
 import EmberObject, { computed } from '@ember/object';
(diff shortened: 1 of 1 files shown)


jenweber:
@locks can you review this one? Thanks!

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -1,9 +1,52 @@
+When a computed property depends on the contents of an array, there are a few
+extra methods you'll need to use in order to correctly recognize when the
+contents of the array change. Arrays have two special keys you can append to
+array properties to track changes on them, `[]` and `@each`.
+
+## `[]`
+
+Sometimes a computed property needs to update when items are added or removed from an array.
```
locks: It should also trigger if you replace one of the items (like, when you put a value directly in an index).

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -1,9 +1,52 @@
...
+In those cases we can use the `[]` array key to tell the property to update at the right time.
+We'll use the obligatory todo list for our examples:
```
locks: Could you replace obligatory with familiar?

locks: Can you replace obligatory with familiar?

source/object-model/computed-properties-and-aggregate-data.md
Bestra: You're right!

```
Sometimes you need a computed property to update when you add, remove, or replace an item in an array.
```
^^ Do you think that's too concise?

I think it'd also be worth calling the replacement case out explicitly in the two lists for `[]` and `@each` as well. 

I'll make a new commit with a crack at the changes and squash once we settle on the final wording

Bestra: Alternate wording
```
Sometimes you need a computed property to update when the membership of an array changes.
```

PR #2283, opened 2018-03-13
Improve test on "Building a Complex Component"

strout:
Check that exactly one result is rendered, instead of checking for at least one.


jenweber:
HI @strout, thanks for this work. In order to merge it, we need the same change to be made to the super rentals example app. Could you also make a PR [here](https://github.com/ember-learn/super-rentals/blob/5feb9ea13dfff554a3564fb267fb8e5d3d7cb405/tests/acceptance/list-rentals-test.js)?

jenweber:
Thank you @strout for The Guides PR (your first, yay!) and @rmminusrfslash for the Super Rentals PR that makes this mergeable.

Merged 🎉

Issue #2280, opened 2018-03-13
[Tutorial] Tests Fail With 404 on /api/rentals after Ember Data step


acorncom:
Migrated from https://github.com/emberjs/ember.js/issues/16368

I've been following the official Ember tutorial and have been writing the recommended tests along with the app code. After completing the Ember Data step I noticed that all of my acceptance tests were failing with a 404 on the GET request to /api/rentals. I am not seeing this issue with the app code however. The rentals display correctly on the rentals page.

Can anyone with some Ember experience shed some light on what's wrong and/or how to fix it? Thanks.

> Promise rejected during "should show rentals as the home page": Ember Data Request GET /api/rentals returned a 404
Payload (Empty Content-Type)
Not found: /api/rentals@ 381 ms
Source:
Error: Ember Data Request GET /api/rentals returned a 404
Payload (Empty Content-Type)
Not found: /api/rentals
at ErrorClass.EmberError (http://localhost:7357/assets/vendor.js:24125:25)
at ErrorClass.AdapterError (http://localhost:7357/assets/vendor.js:157167:15)
at new ErrorClass (http://localhost:7357/assets/vendor.js:157185:22)
at Class.handleResponse (http://localhost:7357/assets/vendor.js:169227:18)
at ajaxError (http://localhost:7357/assets/vendor.js:169720:25)
at Class.hash.error (http://localhost:7357/assets/vendor.js:169308:23)
at fire (http://localhost:7357/assets/vendor.js:3607:31)
at Object.fireWith [as rejectWith] (http://localhost:7357/assets/vendor.js:3737:7)
at done (http://localhost:7357/assets/vendor.js:9646:14)
at XMLHttpRequest. (http://localhost:7357/assets/vendor.js:9887:9)

/cc @rayascott

acorncom:
@rayascott This sounds like you may not have made the changes to `mirage/config.js` mentioned in this section? https://guides.emberjs.com/v3.0.0/tutorial/installing-addons/#toc_ember-cli-mirage

That setup should indeed work ...

If you find you're still hitting snags, we'd love to get you unstuck. We find that it's sometimes more satisfying to get these snags worked out in realtime. If that'd be helpful for you as well, feel free to join us in the Ember community slack at https://ember-community-slackin.herokuapp.com. There's both a `#-help` channel and a `#-team-learning` channel (where we'd love to discuss snags that exist in the tutorial).

Let me know if that helps!

rayascott:
[...]

Issue #2272, opened 2018-03-12
In ember tutorials: acceptance tests now fail after Mirage is added


algo74:
Mirage does not start by default during tests.

I had to add `setupMirage(hooks);` to make it work again

some helpful info: https://stackoverflow.com/questions/49215315/mirage-responds-404-on-acceptances-tests

toddjordan:
Good catch... The tutorial was made with 0.4.1, which worked as is, but it looks like 0.4.2 requires the extra call.

acorncom:
Fixed by #2290

PR #2290, opened 2018-03-16
Add mirage setup code for new test structure

toddjordan:
Addresses #2272 , #2280
See associated pr: https://github.com/ember-learn/super-rentals/pull/97

@samselikoff validating that this is the way to setup application tests now with mirage.

image


acorncom:
Yep, that syntax looks accurate. Thanks Todd

PR #2292, opened 2018-03-17
[DOC] - consistently use template strings

sbatson5:
## What's in this PR?

I noticed a couple of examples in the computed properties guide used string concatenation but all the others used template strings. It was slightly confusing that it wasn't consistent.


acorncom:
Thanks Scott

Issue #2291, opened 2018-03-17
Brace expansion on non-nested properties?


sbatson5:
I was looking at the guides for computed properties and saw the section talking about brace expansion for dependent keys.  The example shown is:

```
  fullName: computed('{firstName,lastName}', function() {
    let firstName = this.get('firstName');
    let lastName = this.get('lastName');

    return `${firstName} ${lastName}`;
  })
```

I had always thought it was bad practice to do brace expansion on `this`.  I.e.
`person.{firstName,lastName}` would be fine but `{firstName,lastName}` isn't.  I feel like otherwise, you would never _not_ use brace expansion.  You could do things like:
`{someService.foo.bar,person.firstName,income,etc}`

acorncom:
@sbatson5 I would tend to concur. Would you mind sending in a PR and we’ll fix it?

sbatson5:
PR here: https://github.com/emberjs/guides/pull/2294 @acorncom

PR #2294, opened 2018-03-17
update brace expansion example

sbatson5:
## What's in this PR?

Fixes https://github.com/emberjs/guides/issues/2291

This updates the example on `brace expansion` on dependent keys of computed properties.  The existing example used keys on the same object as the CP but in my experience, this has been considered bad practice.  This update also shortens this section (which never hurts :P )


locks:
Thanks for the tweak :)

acorncom:
Nice work @sbatson5, you just passed Tom 😀

PR #2295, opened 2018-03-17
Remove references to sync queue

jonnii:
The `sync` queue is deprecated in Ember 3.0.

### DIFF
--- a/source/applications/run-loop.md
+++ b/source/applications/run-loop.md
@@ -124,14 +124,13 @@ What are the queues, and what is their priority order?
 
 ```javascript
 Ember.run.queues
-// => ["sync", "actions", "routerTransitions", "render", "afterRender", "destroy"]
+// => ["actions", "routerTransitions", "render", "afterRender", "destroy"]
 ```
 
-Because the priority is first to last, the "sync" queue has higher priority than the "render" or "destroy" queue.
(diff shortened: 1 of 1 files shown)


jonnii:
@rwjblue there's an animation on that page too, I'm not sure where the code lives, I think it was written by machty a long time ago. Any ideas? I can update that too to remove sync from there if needed.

acorncom:
@jonnii It looks like it's an iframe that loads code up from here (https://s3.amazonaws.com/emberjs.com/run-loop-guide/index.html). Good catch there.

Looks like that's an Ember AppKit app that just does that one thing. If we change the referenced app though we'll change the app for all previous versions of the guides ...

We may want to temporarily copy that into the guides here and then revisit this after we've launched our new guides viewer ... /cc @mansona

jonnii:
@machty am I right in thinking that this is one of your creations?

machty:
@jonnii yes, but it was forked and prettified; I believe this is the latest version https://github.com/olivia/ember-run-loop-visual/tree/ui-prettification (4-5 years ago!?!). Who's up for modifying and rebuilding an old ass ember app :)

jonnii:
@machty I'd be up for the challenge. let me take a look.

jenweber:
Merged. Thanks @jonnii for reporting this and fixing it!

Deprecation is here https://github.com/emberjs/ember.js/pull/16132

PR #2299, opened 2018-03-21
Add "async" before test function

khall:
The "async" keyword isn't included in the initial adding of the new test function, which causes a "await is a reserved word" error later when the tests are run, since "await" is being used without "async".

### DIFF
--- a/source/tutorial/autocomplete-component.md
+++ b/source/tutorial/autocomplete-component.md
@@ -312,7 +312,7 @@ import hbs from 'htmlbars-inline-precompile';
 module('Integration | Component | rental-listing', function(hooks) {
   setupRenderingTest(hooks);
 
-  test('should initially load all listings', async function (assert) {
+  test('should initially load all listings', function (assert) {
   });
 
   test('it renders', async function(assert) {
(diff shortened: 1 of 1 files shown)


locks:
It seems slightly funky that an empty test is already marked as `async`.
Hm, can we not add it when we add the `await`? We could even mention it in the prose.

khall:
I can buy that, updated.

toddjordan:
If we are gonna do this we need to remove the `async` from where the test is mentioned earlier on line 354.

source/tutorial/autocomplete-component.md
```diff
@@ -369,9 +369,9 @@ We will test the full logic of filtering in application tests, described in the
 
 Since our component is expecting the filter process to be asynchronous, we return promises from our filter, using [Ember's RSVP library](https://www.emberjs.com/api/ember/release/modules/rsvp).
 
-Next, we'll add the call to render the component to show the cities we've provided above.
+Next, we'll add the call to render the component to show the cities we've provided above. Since we use `await` when rendering, we must also mark the test's function as `async`.
```
toddjordan: make the async and await words be hyperlinked:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await

toddjordan: "Since we use [the `await` function](...) to wait for rendering, we must..."

sivakumar-kailasam:

PR #2300, opened 2018-03-21
Add importing of "click"

khall:
Final test has importing of click in it, but there's no step to import it.


jenweber:
@khall thanks for catching this and fixing it! Congrats on your first PR to The Guides :D

PR #2301, opened 2018-03-22
Add missing comma

TALlama:


acorncom:
Thanks a bunch @TALlama, that is indeed a problem 😀

Issue #2303, opened 2018-03-22
Update the guides to remove usage of `.get`


locks:
In Ember v3.1.0, you can now omit `.get` to retrieve computed properties.

## Tasks
- [ ] Update code samples to remove `.get` where that is possible
- [ ] Explain the situations where `.get` is still required in "The Object Model"
- [ ] Run the codemod in https://github.com/ember-learn/super-rentals and make sure code snippets in the Guides match

PR #2305, opened 2018-03-22
Use es5 getters

kennethlarsen:
Issue reference: #2303 

This PR changes use of `get()` to es5 getters in places where it makes sense according to the 3.1 release.

Sadly, since I am not a human codemod there might be errors here or areas I have overseen so a sanity check would be nice.

### DIFF
--- a/source/applications/dependency-injection.md
+++ b/source/applications/dependency-injection.md
@@ -250,8 +250,7 @@ export default Component.extend({
   }),
 
   click() {
-    let player = this.get('audioService');
-    player.play(this.get('song.file'));
+    this.audioService.play(this.get('song.file'));
   }
 });
 ```
--- a/source/models/relationships.md
+++ b/source/models/relationships.md
@@ -174,7 +174,7 @@ import PaymentMethod from './payment-method';
 
 export default PaymentMethod.extend({
   last4: DS.attr(),
-  
+
   obfuscatedIdentifier: computed('last4', function () {
     return `**** **** **** ${this.get('last4')}`;
   })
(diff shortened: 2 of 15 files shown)


source/applications/dependency-injection.md
```diff
@@ -250,7 +250,7 @@ export default Component.extend({
   }),
 
   click() {
-    let player = this.get('audioService');
+    let player = this.audioService;
     player.play(this.get('song.file'));
```
Turbo87: this could probably be simplified to `this.audioService.play(...)` now :)

source/models/relationships.md
```diff
@@ -174,9 +174,9 @@ import PaymentMethod from './payment-method';
 
 export default PaymentMethod.extend({
   last4: DS.attr(),
-  
+
   obfuscatedIdentifier: computed('last4', function () {
-    return `**** **** **** ${this.get('last4')}`;
+    return `**** **** **** ${this.last4}`;
```
Turbo87: I'm unsure about this one as it involves Ember Data models. I think we might have to keep using `get()` here for now. 🤔

source/object-model/computed-properties-and-aggregate-data.md
```diff
@@ -25,7 +25,7 @@ export default Component.extend({
   },
 
   titles: computed('todos.[]', function() {
-    let todos = this.get('todos');
+    let todos = this.todos;
     return todos.mapBy('title');
```
Turbo87: like above, these can now be simplified to e.g. `return this.todos.mapBy('title')`

source/models/relationships.md
locks: I believe as long as `last4` isn't a relationship we should be fine, because that's where the proxies are.

sivakumar-kailasam:

PR #2311, opened 2018-03-25
Update covers the new format of config/targets.js

jrjohnson:
The default format of this file has changed a bit and I thought maybe the branch needed a bit of explanation.


acorncom:
Thanks @jrjohnson !

Issue #2307, opened 2018-03-23
creating records with relationships needs more detail


jenweber:
The explanation for creating belongsTo/HasMany records is [a bit sparse](https://guides.emberjs.com/v3.0.0/models/relationships/#toc_creating-records). Could use more detail about the order the records need to be created in and how they're being associated.

karan-pathak:
@jenweber I restructured the documentation and added some things in https://github.com/emberjs/guides/pull/2318, dont know for sure if this is what you wanted.

jenweber:
Closed with #2318 😄

PR #2318, opened 2018-04-04
update create record documentation for ember relationships

karan-pathak:
#2307

### DIFF
--- a/source/models/relationships.md
+++ b/source/models/relationships.md
@@ -277,12 +277,12 @@ export default DS.Model.extend({
 });
 ```
 
-Now, suppose we want to add comments to an existing blogPost. We can do this in two ways, but for both of them, we will first need a reference to our existing blogPost as given below:
+Now, suppose we want to add comments to an existing blogPost. We can do this in two ways, but for both of them, we first need to look up a blog post that is already loaded in the store, using its id:
 
 ```javascript
 let myBlogPost = this.get('store').peekRecord('blog-post', 1);
 ```
-Now we can either set the `belongsTo` relationship in our new comment, or, update the blogPost's `hasMany` relationship. As you might observe, we don't need to set both `hasMany` and `belongsTo` for a record, ember data will do that for us.
+Now we can either set the `belongsTo` relationship in our new comment, or, update the blogPost's `hasMany` relationship. As you might observe, we don't need to set both `hasMany` and `belongsTo` for a record. Ember Data will do that for us.
 
 First, let's look at setting the `belongsTo` relationship in our new comment:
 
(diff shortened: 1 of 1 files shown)


source/models/relationships.md
```diff
@@ -277,31 +277,37 @@ export default DS.Model.extend({
 });
 ```
 
-When a user comments on a blogPost, we need to create a relationship between the two records. We can simply set the `belongsTo` relationship in our new comment:
+Now, suppose we want to add comments to an existing blogPost. We can do this in two ways, but for both of them, we will first need a reference to our existing blogPost as given below:
```
jenweber: How about this?
"We can do this in two ways, but for both of them, we first need to look up a blog post that is already loaded in the store, using its id:"

source/models/relationships.md
```diff
@@ -277,31 +277,37 @@ export default DS.Model.extend({
...
+
+```javascript
+let myBlogPost = this.get('store').peekRecord('blog-post', 1);
+```
+Now we can either set the `belongsTo` relationship in our new comment, or, update the blogPost's `hasMany` relationship. As you might observe, we don't need to set both `hasMany` and `belongsTo` for a record, ember data will do that for us.
```
jenweber: Period after "record". Ember Data should be capitalized.

source/models/relationships.md
```diff
@@ -277,31 +277,37 @@ export default DS.Model.extend({
...
+
+First, let's look at setting the `belongsTo` relationship in our new comment:
 
 ```javascript
-let blogPost = this.get('store').peekRecord('blog-post', 1);
 let comment = this.get('store').createRecord('comment', {
-  blogPost: blogPost
+  blogPost: myBlogPost
```
jenweber: I really like this distinction of calling it `myBlogPost`

source/models/relationships.md
```diff
@@ -277,31 +277,37 @@ export default DS.Model.extend({
...
 });
 comment.save();
 ```
 
-This will create a new `comment` record and save it to the server. Ember Data will also update the blogPost to include our newly created comment in its `comments` relationship.
+In the above snippet, we have referenced `myBlogPost` while creating the record. This will let ember know that the newly created comment belongs to `myBlogPost`.
```
jenweber: capitalize Ember

jenweber:
Thanks @karan-pathak, I think your edits make it much clearer what the purpose of these snippets are and how to work with the relationships. I like how you reference a few times that there are multiple ways to establish the relationships, and you tie it back into hasMany/belongsTo. I have just a few minor changes to request before we can merge. See my code comments. Thanks!

source/models/relationships.md
karan-pathak: It's better. I made this change.

source/models/relationships.md
karan-pathak: Done.

source/models/relationships.md
karan-pathak: Thanks! :)

source/models/relationships.md
karan-pathak: Done.

karan-pathak:
Hi @jenweber, I've made the changes that you requested :)

jenweber:
Thank you @karan-pathak! This helps address confusion that has come up multiple times in Ember Slack, particularly for new users.

Merged, closes #2307

PR #2322, opened 2018-04-13
Set headers in init() so people following the guides don't get eslint errors

sdhull:
As [suggested](https://embercommunity.slack.com/archives/C04KG57CF/p1523580736000122) on the `#-team-learning` slack channel, this example has been updated so it won't produce eslint errors for people who copy it directly out of the guide.

I added a note in parentheses explaining why, which I was kind of on the fence about including. Happy to update this as the wise learning team sees fit, lmk.


jenweber:
@sdhull Can you change this to us `this.set()` instead of just this.headers? After that, it looks good to go.

jenweber:
Super, thank you! So fast 😁 Congrats on your first Guides PR

PR #2328, opened 2018-04-27
update 'final code' to be same as previous code

ardenzhan:
test doesn't pass if using this.$ instead of this.element.querySelector


mansona:
losing in favour of https://github.com/ember-learn/guides-source/pull/28 👍

PR #2330, opened 2018-05-05
updated defining-models.md | ES6 syntax

rahulsingh-js:
I believe, we all should start using ES6 syntax in upcoming documents

### DIFF
--- a/source/models/defining-models.md
+++ b/source/models/defining-models.md
@@ -56,7 +56,7 @@ export default DS.Model.extend({
   firstName: DS.attr(),
   lastName: DS.attr(),
 
-  fullName: computed('firstName', 'lastName', function() {
+  fullName: computed('firstName', 'lastName', () => {
     return `${this.get('firstName')} ${this.get('lastName')}`;
   })
 });


jenweber:
Hi @rahhulsingh-js, arrow functions should not be used for computed properties. This code will not work.

That said, your PRs raised three important th

jenweber:
Accidentally hit submit & close before I was done typing. See my edited comment above.

Thanks!

rahulsingh-js:
Thank for letting me know. But it seems to me a real concern that why an
anonymous function wouldn't work using arrow syntax operation!

locks:
@rahulsingh-js arrow functions are not merely a short-hand, they also maintain `this` bound to outer scope of the function.

When you do:

 ```app/models/person.js
 import DS from 'ember-data';
 import { computed } from '@ember/object';
 
   fullName: computed('firstName', 'lastName', () => {
     return `${this.get('firstName')} ${this.get('lastName')}`;
   })
 });
 ```

You are saying that the `this` inside the callback is the same as the surrounding scope.
In JavaScript, `this` is bound to the function definition, and since `DS.Model.extend({})` is a function call to which an object is passed, the surrounding scope is the scope of the module.
By specification, `this` inside a module is undefined, meaning that the above code will turn into something like:

 ```app/models/person.js
 import DS from 'ember-data';
 import { computed } from '@ember/object';

// `this` scope: module.
// `this` value: undefined.
 
    // `this` scope: module.
    // `this` value: undefined.

   fullName: computed('firstName', 'lastName', () => {
     // `this` scope: module.
     // `this` value: undefined.

     return `${this.get('firstName')} ${this.get('lastName')}`;
   })
 });
 ```

rahulsingh-js:
Thanks

rahulsingh-js:
I would be happy to help

PR #2332, opened 2018-05-05
updated creating-updating-and-deleting-records.md | ES6 syntax

rahulsingh-js:
Modified docs according to ES6+ standards.


sivakumar-kailasam:
Closing for the same reason as #2333

PR #2336, opened 2018-05-05
updated defining-your-routes.md | ES6 syntax

rahulsingh-js:
Modified docs according to ES6+ standards.


jenweber:
Closed, see comments on https://github.com/emberjs/guides/pull/2330. Most code examples in the Guides that use `this` will likely break with arrow functions.

PR #2337, opened 2018-05-10
Fixing lint errors

mehrdadrafiee:
Replaces double quotes with single quotes and removes unnecessary comma.

PR #2338, opened 2018-05-10
Fix lint error

mehrdadrafiee:
Replaces double quotes with single quotes.


jenweber:
See comments on https://github.com/emberjs/guides/pull/2337#issuecomment-388243067

